PR #67, opened 2017-05-16
Inheritance should use :

### LINKED ISSUE #66: Inheritance uses "extends" or "implements" keyword, should use :
Right now, kotlinpoet emits the following for inheritance 

```
class MyClassImpl implements MyClass
```

should be 

```
class MyClassImpl: MyClass
```

burntcookie90:
Currently using `extends`/`implements` when we should be using `:`

Fixes #66

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -118,24 +118,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
           implementsTypes = superinterfaces
         }
 
-        if (!extendsTypes.isEmpty()) {
-          codeWriter.emit(" extends")
-          var firstType = true
-          for (type in extendsTypes) {
-            if (!firstType) codeWriter.emit(",")
-            codeWriter.emit(" %T", type)
-            firstType = false
-          }
-        }
-
-        if (!implementsTypes.isEmpty()) {
-          codeWriter.emit(" implements")
-          var firstType = true
-          for (type in implementsTypes) {
-            if (!firstType) codeWriter.emit(",")
-            codeWriter.emit(" %T", type)
-            firstType = false
-          }
+        (extendsTypes + implementsTypes).forEachIndexed { i, typeName ->
+          codeWriter.emit(if (i == 0)  " :" else ",")
+          codeWriter.emit(" %T", typeName)
         }
 
         codeWriter.emit(" {\n")
--- a/src/test/java/com/squareup/kotlinpoet/KotlinFileTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/KotlinFileTest.kt
@@ -471,8 +471,8 @@ class KotlinFileTest {
         |
         |import com.squareup.wire.Message
         |
-        |class Taco extends Message {
-        |  class Builder extends Message.Builder {
+        |class Taco : Message {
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -118,24 +118,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
-        if (!extendsTypes.isEmpty()) {
-          codeWriter.emit(" extends")
-          var firstType = true
-          for (type in extendsTypes) {
-            if (!firstType) codeWriter.emit(",")
-            codeWriter.emit(" %T", type)
-            firstType = false
-          }
+        if(!extendsTypes.isEmpty() || !implementsTypes.isEmpty()) {
```
JakeWharton: nit: space after `if`

JakeWharton: Use `isNotEmpty()`?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -118,24 +118,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
-          codeWriter.emit(" implements")
-          var firstType = true
-          for (type in implementsTypes) {
-            if (!firstType) codeWriter.emit(",")
-            codeWriter.emit(" %T", type)
-            firstType = false
-          }
+        var firstType = true
+        (extendsTypes + implementsTypes).forEach {
+          if (!firstType) codeWriter.emit(",")
```
JakeWharton: can use `forEachIndexed` and `i > 0` for this to save the boolean

swankjesse: maybe do

```
codeWriter.emit(firstType ? ":" : ",")
```

swankjesse: Also – I think we want a space before the ":" to match examples on kotlinlang.org.

burntcookie90:
Sounds good to me, I'll update and push in a bit

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
burntcookie90: @swankjesse can't ternary so I'll do an inline `if`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -118,24 +118,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
-        if (!implementsTypes.isEmpty()) {
-          codeWriter.emit(" implements")
-          var firstType = true
-          for (type in implementsTypes) {
-            if (!firstType) codeWriter.emit(",")
-            codeWriter.emit(" %T", type)
-            firstType = false
-          }
+        (extendsTypes + implementsTypes).forEachIndexed { i, typeName ->
+          codeWriter.emit(if (i == 0)  " :" else ",")
```
burntcookie90: this could be combined into `codeWriter.emit("${if (i == 0)  " :" else ","} %T", typeName)` but that feels a bit complex

swankjesse:
nice

PR #70, opened 2017-05-18
[WIP] Adds nullable wrapper for TypeName

rharter:
**Don't Merge**: For discussion and feedback.

This also includes a convenience extension function to turn any TypeName into a Nullable version.

TODO:

* Add convenience methods to *Spec builders to create nullable varieties.
* Add more tests.
* Make `NullableTypeName` Comaprable


rharter:
Wanted to share an early version of this (I'm working half days for this part of paternity leave 😉).

I worked though adding `nullable` to TypeName, but that touched a ton of code.  Tried this again, and it turned out to be much simpler than I first thought.  

The thought was to add convenience functions to the *Spec companion objects so that people can super easily define nullable versions, and also fix up the `ClassName.bestGuess()` parsing to make the ClassName nullable if the string ends with a `?`.

ronshapiro:
Could this be the job of the `CodeWriter`? Whenever it dispatches to `TypeName.emit()` it can check the nullable bit? This approach seems clean but is the opposite of what we did for annotations on a `TypeName` (i.e. we don't have `AnnotatedTypeName`). There are times that accepting a `ClassName` or `ParameterizdTypeName` is important for an API, and this would make those methods awkward.

rharter:
I was actually trying out something similar to the annotation approach in
another branch. I'll continue with that tomorrow.

The challenge there ends up being that all factory methods for various
TypeNames then need a nullable param added, which makes it very high
touch.  I'll get a poc of that tomorrow so we can discuss.

swankjesse:
@rharter what I was thinking is that all of the public factory methods for all of the various TypeName implementations would return a non-nullable instance, and there’d be an abstract `nullable()` method on `TypeName` that they’d all override. Kind of like we do with `annotated(...)` and `withoutAnnotations()`.

rharter:
Sorry for all the commits, but here you go.  

This update adds the `asNullable()` and `asNonNullable()` methods and a `nullable` property to `TypeName`.  The CodeWriter, in turn, is responsible for writing the `?` if the TypeName is nullable.

swankjesse:
I think this is great! One small request: wanna add a test for something like `Map<String?, List<Int?>>` ? Because we can!

Once that’s in I’ll merge!

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -296,4 +295,10 @@ class CodeBlockTest {
       assertThat(expected).hasMessage("statement exit %] has no matching statement enter %[")
     }
   }
+
+  @Test fun nullableType() {
+    val type = TypeName.get(String::class).asNullable()
+    val typeBlock = CodeBlock.builder().add("%T", type).build()
```
ronshapiro: CodeBlock.of()?

rharter:
@swankjesse Good call.  Those tests identified some issues that have now been fixed.

rharter:
...or have they?

So the issue here is who owns the emission of the `?`.  If the CodeWriter is responsible for it, then it works properly for `%T`, but the `toString` won't match, since it doesn't emit there.

The TypeName itself can't do it, though, since it needs to show up in different places for different types.  For instance, `ClassName` should result in "Foo?", but `ParameterizedTypeName` should be "Foo<...>?".  If ClassName is responsible for emitting it's own `?`, then PTypeName ends up emitting "Foo?<...>".  

The current implementation makes CodeWriter responsible for the final `?`, and compound types like ParameterizedTypeName responsible for the inner `?`s, but that results in the toString being different, as noted in the test failure.

Any suggestions for how to handle that?

JakeWharton:
I think each should be responsible in toString, and PTN should handle
writing the owning name and question mark instead of delegating to the
normal CN.toString.

rharter:
of course

rharter:
@swankjesse this is ready for merge from my end.  Let me know if you want any changes.

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -296,4 +295,28 @@ class CodeBlockTest {
...
   }
+
+  @Test fun nullableType() {
+    val type = TypeName.get(String::class).asNullable()
+    val typeBlock = CodeBlock.of("%T", type)
+    assertThat(typeBlock.toString()).isEqualTo("java.lang.String?")
+
+    val list = ParameterizedTypeName.get(ClassName.get(List::class).asNullable(), TypeName.get(Int::class).asNullable()).asNullable()
+    val listBlock = CodeBlock.of("%T", list)
+    assertThat(listBlock.toString()).isEqualTo("java.util.List<kotlin.Int?>?")
```
swankjesse: cool!

swankjesse:
I squashed it, linewrapped it, & merged it!

swankjesse:
Thanks Ryan!

JakeWharton:
Great work. Thanks!

PR #73, opened 2017-05-19
Support delegated property.

kafji:


swankjesse:
Thanks. I rebased and merged manually. I also made a small tweak to your API: I hope you like it.

```
    val prop = PropertySpec.builder(String::class, "foo")
        .delegate("Delegates.notNull()")
        .build()
```

kafji:
Thank you.

src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -19,6 +19,15 @@ import com.google.common.truth.Truth.assertThat
 import org.junit.Test
 
 class PropertySpecTest {
+
+  @Test fun delegated() {
+    val prop = PropertySpec.builder(String::class, "foo")
+        .delegated(true)
+        .initializer("Delegates.notNull()")
```
JakeWharton: Should probably use a %T and reference the class

Issue #60, opened 2017-05-16
Flip API signatures to be (name, type) instead of (type, name) ?


swankjesse:
For consistency with the language. Plus we could potentially make the type part elided.

PR #74, opened 2017-05-20
#60 Flip API signatures to be (name, type) instead of (type, name)

geralt-encore:
https://github.com/square/kotlinpoet/issues/60


swankjesse:
Rebased & pushed upstream. Thanks!

Issue #49, opened 2017-05-15
object support


JakeWharton:
Basically the same as `class` except no constructors allowed

consp1racy:
No companion object either

burntcookie90:
should be able to be closed now

PR #75, opened 2017-05-21
#49 object support

burntcookie90:
Initial implementation of #49

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -101,6 +101,8 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitModifiers(modifiers, setOf(PUBLIC))
         if (kind == Kind.ANNOTATION) {
           codeWriter.emit("annotation class %L", name)
+        } else if (kind == Kind.OBJECT) {
+          codeWriter.emit("object class %L", name)
         } else {
           codeWriter.emit("%L %L", kind.name.toLowerCase(Locale.US), name)
         }
@@ -231,6 +233,10 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         setOf(KModifier.PUBLIC),
         setOf(KModifier.PUBLIC)),
 
+    OBJECT(
+        setOf(KModifier.PUBLIC),
+        setOf(KModifier.PUBLIC)),
+
     INTERFACE(
         setOf(KModifier.PUBLIC),
         setOf(KModifier.PUBLIC, KModifier.ABSTRACT)),
@@ -323,7 +329,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun superclass(superclass: TypeName): Builder {
-      check(kind == Kind.CLASS) { "only classes have super classes, not $kind" }
+      check(kind == Kind.CLASS || kind == Kind.OBJECT) { "only classes have super classes, not $kind" }
       check(this.superclass === ANY) { "superclass already set to ${this.superclass}" }
       this.superclass = superclass
       return this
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -190,6 +195,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         firstMember = false
       }
 
+      // Companion object
```
JakeWharton: Delete, not useful

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -101,6 +102,10 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitModifiers(modifiers, setOf(PUBLIC))
         if (kind == Kind.ANNOTATION) {
           codeWriter.emit("annotation class %L", name)
+        } else if (kind == Kind.OBJECT) {
+          codeWriter.emit("object class %L", name)
+        } else if (kind == Kind.COMPANION) {
+          codeWriter.emit("companion object")
```
JakeWharton: These can have custom names. Do we care?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -311,6 +330,13 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       return this
     }
 
+    fun companionObject(companionObject: TypeSpec): Builder {
+      check(kind == Kind.CLASS || kind == Kind.INTERFACE) { "$kind can't have a companion object" }
+      require(companionObject.kind == Kind.COMPANION) { "expect a companion object class but was $kind "}
```
JakeWharton: Expected

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -323,7 +349,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun superclass(superclass: TypeName): Builder {
-      check(kind == Kind.CLASS) { "only classes have super classes, not $kind" }
+      check(kind == Kind.CLASS || kind == Kind.OBJECT || kind == Kind.COMPANION) {
+        "only classes have super classes, not $kind"
```
JakeWharton: Can have

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2143,6 +2143,106 @@ class TypeSpecTest {
...
+        .addInitializerBlock(CodeBlock.builder().build())
+        .addFun(FunSpec.builder("test")
+            .addModifiers(KModifier.PUBLIC)
+            .build())
+        .build()
+
+    assertThat(toString(type)).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |object class MyObject {
```
JakeWharton: Shouldn't emit `class` here

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
burntcookie90: good question, it's not a feature i've ever used, but it wouldn't be hard to include

JakeWharton: Skip for now. Would be a good followup

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -323,7 +348,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun superclass(superclass: TypeName): Builder {
-      check(kind == Kind.CLASS) { "only classes have super classes, not $kind" }
+      check(kind == Kind.CLASS || kind == Kind.OBJECT || kind == Kind.COMPANION) {
```
swankjesse: nice

Issue #76, opened 2017-05-21
Support LambdaTypeName 


burntcookie90:
Should have a `LambdaTypeName` in order to support function parameters and the such.

JakeWharton:
Merged

PR #78, opened 2017-05-21
lambda type name

burntcookie90:
This adds a `LambdaTypeName` as specified in #76 in order to have lambda parameters for functions.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
@@ -1,33 +1,52 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.squareup.kotlinpoet
 
 class LambdaTypeName internal constructor(
+    private val receiver: TypeName? = null,
     private val parameters: List<TypeName> = emptyList(),
     private val returnType: TypeName = TypeName.get(Unit::class),
     nullable: Boolean = false,
     annotations: List<AnnotationSpec> = emptyList()
 ) : TypeName(nullable, annotations) {
-  override fun asNullable() = LambdaTypeName(parameters, returnType, true, annotations)
+  override fun asNullable() = LambdaTypeName(receiver, parameters, returnType, true, annotations)
 
-  override fun asNonNullable() = LambdaTypeName(parameters, returnType, false, annotations)
+  override fun asNonNullable() = LambdaTypeName(receiver, parameters, returnType, false, annotations)
 
   override fun annotated(annotations: List<AnnotationSpec>): TypeName
-      = LambdaTypeName(parameters, returnType, nullable, annotations)
+      = LambdaTypeName(receiver, parameters, returnType, nullable, annotations)
 
-  override fun withoutAnnotations(): TypeName = LambdaTypeName(parameters, returnType, nullable)
+  override fun withoutAnnotations(): TypeName = LambdaTypeName(receiver, parameters, returnType, nullable)
 
   override fun abstractEmit(out: CodeWriter): CodeWriter {
     annotations.forEach { it.emit(out, true) }
     if (nullable) {
       out.emit("(")
     }
+
+    receiver?.let {
+      out.emit("%T.", it)
+    }
+
     out.emit("(")
-    if (!parameters.isEmpty()) {
-      parameters.forEachIndexed { i, it ->
-        if (i != 0) {
-          out.emit(", ")
-        }
-        out.emit("%T", it)
+    parameters.forEachIndexed { i, it ->
+      if (i != 0) {
+        out.emit(", ")
       }
+      out.emit("%T", it)
     }
     out.emit(") -> %T", returnType)
     if (nullable) {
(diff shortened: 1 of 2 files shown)


burntcookie90:
This still has a good bit of work left, especially in testing, pushing it up for an initial design review.

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,48 @@
+package com.squareup.kotlinpoet
```
JakeWharton: copyright header

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class LambdaTypeName internal constructor(
+    private val parameters: List<TypeName> = emptyList(),
+    private val returnType: TypeName = TypeName.get(Unit::class),
```
JakeWharton: We have this as a constant. Should just be able to be `UNIT`

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,48 @@
...
+
+  override fun asNonNullable() = LambdaTypeName(parameters, returnType, false, annotations)
+
+  override fun annotated(annotations: List<AnnotationSpec>): TypeName
+      = LambdaTypeName(parameters, returnType, nullable, annotations)
+
+  override fun withoutAnnotations(): TypeName = LambdaTypeName(parameters, returnType, nullable)
+
+  override fun abstractEmit(out: CodeWriter): CodeWriter {
+    annotations.forEach { it.emit(out, true) }
```
JakeWharton: not sure these can even be annotated... i'll check

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,48 @@
...
+    out.emit(") -> %T", returnType)
+    if (nullable) {
+      out.emit(")")
+    }
+    return out
+  }
+
+  companion object {
+    /** Returns a lambda type with `returnType` and parameters of listed in `parameters` **/
+    @JvmStatic fun get(returnType: TypeName, parameters: List<TypeName>)
```
JakeWharton: It's weird these are out-of-order. When you consider that we'll also need the ability to specify a receiver type, it may be better to do:
```
fun get(receiver: TypeName? = null, parameters: List<TypeName>, returnType: TypeName)
fun get(receiver: TypeName? = null, vararg parameters: TypeName, returnType: TypeName)
```

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,74 @@
...
+    if (nullable) {
+      out.emit("(")
+    }
+
+    receiver?.let {
+      out.emit("%T.", it)
+    }
+
+    out.emit("(")
+    if (!parameters.isEmpty()) {
```
JakeWharton: nit: isNotEmpty

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
JakeWharton: nit: 2017

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -19,6 +19,7 @@ import com.google.common.collect.ImmutableSet
 import com.google.common.collect.Iterables.getOnlyElement
 import com.google.common.truth.Truth.assertThat
 import com.google.testing.compile.CompilationRule
+import com.sun.tools.corba.se.idl.constExpr.BooleanNot
```
JakeWharton: wut

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -175,6 +176,97 @@ class FunSpecTest {
...
+      |fun foo(f: ((kotlin.Boolean, java.lang.String) -> kotlin.Unit)?): java.lang.String {
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun functionParamMultipleNullableLambdaParam() {
+    val unitType = ClassName.get(Unit::class)
+    val booleanType = ClassName.get(Boolean::class)
+    val stringType = ClassName.get(String::class).asNullable()
+    val lambdaTypeName = LambdaTypeName.get(parameters = *arrayOf(booleanType, stringType), returnType = unitType).asNullable()
```
JakeWharton: can you do `get(booleanType, stringType, returnType = unitType)`?

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
burntcookie90: yanked from another file, good catch

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
burntcookie90: wat

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
burntcookie90: nah, `booleanType` registers as the `receiver` parameter

burntcookie90:
Ok, I've also added initializer support which, I think, rounds out most of the features. Is there anything else to be added (or tests for coverage)?

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,81 @@
...
+    if (nullable) {
+      out.emit("(")
+    }
+
+    receiver?.let {
+      out.emit("%T.", it)
+    }
+
+    out.emit("(")
+    if (parameters.isNotEmpty()) {
```
JakeWharton: nit: this `if` block has no real value. drop it. (sorry should have noticed earlier)

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -175,6 +175,116 @@ class FunSpecTest {
...
+      |""".trimMargin())
+  }
+
+  @Test fun functionParamNoLambdaParamWithInitializer() {
+    val unitType = UNIT
+    val lambdaInitializer = CodeBlock.builder()
+        .add("{")
+        .add("}")
+        .build()
+    val lambdaType = LambdaTypeName.get(returnType = unitType, initializer = lambdaInitializer)
```
JakeWharton: This seems like it should be part of the param spec and not on the type.

burntcookie90: You're totally right, hadn't even thought about that. I'll revert.

swankjesse:
Merged manually!

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -0,0 +1,72 @@
...
+
+  override fun asNonNullable() = LambdaTypeName(receiver, parameters, returnType, false, annotations)
+
+  override fun annotated(annotations: List<AnnotationSpec>): TypeName
+      = LambdaTypeName(receiver, parameters, returnType, nullable, annotations)
+
+  override fun withoutAnnotations(): TypeName = LambdaTypeName(receiver, parameters, returnType, nullable)
+
+  override fun abstractEmit(out: CodeWriter): CodeWriter {
+    annotations.forEach { it.emit(out, true) }
```
swankjesse: as merged I used `emitAnnotations(out)` which puts a space delimiter between each

burntcookie90:
thanks!

PR #81, opened 2017-05-22
Beginnings of custom accessors.

swankjesse:
Still more to do here, but this is a good start.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -108,7 +108,7 @@ internal class CodeWriter @JvmOverloads constructor(
     trailingNewline = true // Force the '//' prefix for the comment.
     comment = true
     try {
-      emit(codeBlock)
+      emitCode(codeBlock)
       emit("\n")
     } finally {
       comment = false
@@ -122,7 +122,7 @@ internal class CodeWriter @JvmOverloads constructor(
     emit("/**\n")
     kdoc = true
     try {
-      emit(kdocCodeBlock)
+      emitCode(kdocCodeBlock)
     } finally {
       kdoc = false
     }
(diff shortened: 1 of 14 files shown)


src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -195,13 +195,13 @@ internal class CodeWriter @JvmOverloads constructor(
   }
 
   @Throws(IOException::class)
-  fun emit(s: String): CodeWriter = emitAndIndent(s)
+  fun emitCode(s: String) = emitCode(CodeBlock.of(s))
```
swankjesse: I renamed this because we had two methods, `emit(String)` and `emit(String, vararg)` that had different behavior and that was really weird for the calling code. The vararg one interpreted its input as code; the non-vararg didn’t.

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -281,4 +281,36 @@ class KotlinPoetTest {
...
+                .build())
+            .build())
+        .build()
+    assertThat(source.toString()).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |import kotlin.Int
+        |
+        |var propertyWithCustomAccessors: Int = 1
+        |  get() {
```
swankjesse: I think we start with full verbose syntax, then later tweak it to elide stuff we can elide

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -468,3 +488,9 @@ class FunSpec private constructor(builder: FunSpec.Builder) {
     }
   }
 }
+
+internal val String.isConstructor: Boolean
+  get() = this == CONSTRUCTOR
+
+internal val String.isAccessor: Boolean
```
vanniktech: Return type can be interfered here

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -468,3 +488,9 @@ class FunSpec private constructor(builder: FunSpec.Builder) {
     }
   }
 }
+
+internal val String.isConstructor: Boolean
```
vanniktech: Return type can be interfered here

swankjesse: Will fix in follow up.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
swankjesse: Will fix in follow up.

PR #82, opened 2017-05-22
Docs update. bestGuess method input not ASCII only

andreilisun:
I suppose you just forgot to update a comment in https://github.com/square/javapoet/commit/dfdc30b636ffbd053cfd26cac1c659db7a01a48e and copied it here.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -145,7 +145,7 @@ class ClassName private constructor(
 
     /**
      * Returns a new [ClassName] instance for the given fully-qualified class name string. This
-     * method assumes that the input is ASCII and follows typical Java style (lowercase package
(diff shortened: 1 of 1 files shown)


swankjesse:
I don’t think this method is going to do anything reasonable if the name isn’t ASCII.

andreilisun:
Sorry, was confused with [test](https://github.com/square/kotlinpoet/blob/8accecbf71153a63533ba5b952ce5d3b1cf5222c/src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt#L33-L38) and [codePointAt(...)](https://github.com/square/kotlinpoet/blob/8accecbf71153a63533ba5b952ce5d3b1cf5222c/src/main/java/com/squareup/kotlinpoet/ClassName.kt#L159). Of course by convention name is always written in ASCII letters.

swankjesse:
Yeah. I prefer to write codepoint-aware code everywhere, even in cases where it's not necessary.

PR #84, opened 2017-05-22
#77: Draft of typealias support

Egorand:


src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -0,0 +1,97 @@
...
+/** A generated typealias declaration */
+class TypeAliasSpec private constructor(builder: TypeAliasSpec.Builder) {
+  val name: String
+  val type: TypeName
+  val modifiers: Set<KModifier>
+
+  init {
+    this.name = builder.name
+    this.type = builder.type
+    this.modifiers = builder.modifiers
```
JakeWharton: These can be initialized inline on the property

JakeWharton: Also this is missing `toImmutableSet`

src/test/java/com/squareup/kotlinpoet/TypeAliasSpecTest.kt
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
JakeWharton: 2017

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
JakeWharton: 2017

src/test/java/com/squareup/kotlinpoet/TypeAliasSpecTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    val typeAliasSpec = TypeAliasSpec
+        .builder("Word", String::class)
+        .visibility(KModifier.PRIVATE)
+        .build()
+    assertThat(typeAliasSpec.toString()).isEqualTo("""
+        |private typealias Word = java.lang.String
+        |""".trimMargin())
+  }
+
+  @Test fun forbiddenModifiers() {
```
swankjesse: I think this test is probably overkill. We want our tests to exercise the interesting logic in our code, but not to re-implement that logic

swankjesse:
Supercool!

src/test/java/com/squareup/kotlinpoet/TypeAliasSpecTest.kt
Egorand: That makes sense, removed

Egorand:
Oops, looks like `CodeWriter` APIs have changed, will rebase and fix later.

swankjesse:
Thanks!

PR #85, opened 2017-05-22
Adds support for collapsing TypeSpec properties into constructor params. 

### LINKED ISSUE #83: Collapse properties into constructor params when present
When a TypeSpec has constructor params that match it's properties, the properties should be collapsed into the constructor parameters.

This

```
class Foo(val a: String) {
}
```

Instead of this:

```
class Foo(a: String) {
  val a: String
}
```

rharter:
Fixes #83

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
+++ b/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
@@ -148,10 +148,7 @@ class KotlinFile private constructor(builder: KotlinFile.Builder) {
       when (member) {
         is TypeSpec -> member.emit(codeWriter, null)
         is FunSpec -> member.emit(codeWriter, null, setOf(KModifier.PUBLIC))
-        is PropertySpec -> {
-          member.emit(codeWriter, setOf(KModifier.PUBLIC))
-          codeWriter.emit("\n")
-        }
+        is PropertySpec -> member.emit(codeWriter, setOf(KModifier.PUBLIC))
         else -> throw AssertionError()
       }
     }
--- a/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
+++ b/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
@@ -148,7 +148,10 @@ class KotlinFile private constructor(builder: KotlinFile.Builder) {
       when (member) {
         is TypeSpec -> member.emit(codeWriter, null)
         is FunSpec -> member.emit(codeWriter, null, setOf(KModifier.PUBLIC))
-        is PropertySpec -> member.emit(codeWriter, setOf(KModifier.PUBLIC))
+        is PropertySpec -> {
+          member.emit(codeWriter, setOf(KModifier.PUBLIC))
+          codeWriter.emit("\n")
+        }
         else -> throw AssertionError()
       }
     }
--- a/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -37,13 +37,15 @@ class PropertySpec private constructor(builder: Builder) {
   val setter: FunSpec? = builder.setter
 
   @Throws(IOException::class)
-  internal fun emit(codeWriter: CodeWriter, implicitModifiers: Set<KModifier>) {
+  internal fun emit(codeWriter: CodeWriter, implicitModifiers: Set<KModifier>,
+                    withInitializer: Boolean = true,
+                    inline: Boolean = false) {
     codeWriter.emitKdoc(kdoc)
     codeWriter.emitAnnotations(annotations, false)
     codeWriter.emitModifiers(modifiers, implicitModifiers)
     codeWriter.emit(if (mutable) "var " else "val ")
     codeWriter.emitCode("%L: %T", name, type)
-    if (initializer != null) {
+    if (withInitializer && initializer != null) {
       if (delegated) {
         codeWriter.emit(" by ")
       } else {
@@ -51,7 +53,7 @@ class PropertySpec private constructor(builder: Builder) {
       }
       codeWriter.emitCode("%[%L%]", initializer)
     }
-    codeWriter.emit("\n")
+    if (!inline) codeWriter.emit("\n")
     if (getter != null) {
       codeWriter.emitCode("%>")
       getter.emit(codeWriter, null, implicitModifiers)
@@ -78,7 +80,6 @@ class PropertySpec private constructor(builder: Builder) {
     try {
       val codeWriter = CodeWriter(out)
       emit(codeWriter, emptySet<KModifier>())
-      codeWriter.emit("\n")
       return out.toString()
     } catch (e: IOException) {
       throw AssertionError()
--- a/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -49,7 +49,6 @@ class PropertySpec private constructor(
       }
       codeWriter.emit("%[%L%]", initializer)
     }
-    codeWriter.emit("\n")
   }
 
   override fun equals(other: Any?): Boolean {
@@ -66,6 +65,7 @@ class PropertySpec private constructor(
     try {
       val codeWriter = CodeWriter(out)
       emit(codeWriter, emptySet<KModifier>())
+      codeWriter.emit("\n")
       return out.toString()
     } catch (e: IOException) {
       throw AssertionError()
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -153,7 +153,6 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
           codeWriter.emit(if (i == 0)  " :" else ",")
           codeWriter.emitCode(" %T", typeName)
         }
-
         codeWriter.emit(" {\n")
       }
 
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -76,6 +76,19 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     val previousStatementLine = codeWriter.statementLine
     codeWriter.statementLine = -1
 
+    val ctorProperties: Map<String, PropertySpec> =
+        if (primaryConstructor == null || primaryConstructor.parameters.isEmpty()) {
+          emptyMap<String, PropertySpec>()
+        } else {
+          propertySpecs
+              .filter { it.initializer != null }
+              .filter { prop ->
+                primaryConstructor.parameters.find { it.name == prop.name && it.type == prop.type && CodeBlock.of("%N", it) == prop.initializer } != null
+              }
+              .map { Pair(it.name, it) }
+              .toMap()
+        }
+
     try {
       if (enumName != null) {
         codeWriter.emitKdoc(kdoc)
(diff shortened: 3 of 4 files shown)


src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
```diff
@@ -148,7 +148,10 @@ class KotlinFile private constructor(builder: KotlinFile.Builder) {
       when (member) {
         is TypeSpec -> member.emit(codeWriter, null)
         is FunSpec -> member.emit(codeWriter, null, setOf(KModifier.PUBLIC))
-        is PropertySpec -> member.emit(codeWriter, setOf(KModifier.PUBLIC))
+        is PropertySpec -> {
```
swankjesse: This seems off to me . . . the other members have enough context when emitting to include a delimiter or not. Might be worth actually passing that context into the emit method. Like how TypeSpec’s emit() method supplies an enclosing enum name

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -110,7 +110,21 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
-        primaryConstructor?.emitParameterList(codeWriter)
+        primaryConstructor?.let {
+          codeWriter.emit("(")
+          var firstParameter = true
+          it.parameters.forEachIndexed { index, param ->
+            if (!firstParameter) codeWriter.emit(",").emitWrappingSpace()
+            val property = propertySpecs.find { it.name == param.name && it.type == param.type }
+            if (property != null) {
+              property.emit(codeWriter, setOf(PUBLIC))
```
swankjesse: ie. probably put a boolean in here like `inlineInConstructor` or something

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -151,8 +165,12 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
       // Non-static properties.
       for (propertySpec in propertySpecs) {
+        if (primaryConstructor?.parameters?.find { it.name == propertySpec.name && it.type == propertySpec.type } != null) {
```
swankjesse: I wonder if would simplify to index inlined parameters? Like build a `Map<String, PropertySpec>` ?

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -91,6 +91,36 @@ class KotlinPoetTest {
         |""".trimMargin())
   }
 
+  @Test fun primaryConstructorProperties() {
```
swankjesse: We should probably only inline parameters where the initializer is the name of the parameter. Like if it was `val cheese: Cheese = cheese`. That way we’re not risking changing semantics.

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -91,6 +91,36 @@ class KotlinPoetTest {
...
+    val source = KotlinFile.get(tacosPackage, TypeSpec.classBuilder("Taco")
+        .primaryConstructor(FunSpec.constructorBuilder()
+            .addParameter("cheese", String::class)
+            .addParameter("cilantro", String::class)
+            .beginControlFlow("require (!cheese.isEmpty())")
```
swankjesse: nit: `require` is a method, not a keyword so no space after it

rharter: Hahaha, this but was copy pasta from the method above ☺️

rharter: I can fix both, though.

src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
rharter: I didn't like it either, hadn't considered an inline flag. Will do.

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
rharter: Ahh, good thinking, I'll get it done.

rharter:
There's one thing that doesn't sit well with this approach for me. Want to bring it up for discussion and am happy to overlook it if you feel it no big deal.

The *poet apis are all about modeling your code, but this approach kind of doubles up the work you have to do, i.e to create a class that has properties defined in the constructor you have to declare a model that has both the constructor and properties, which doesn't model the actual code you're trying to generate. 

Would it make more sense, and make the generation code more representative of the actual resulting code, to simply add a `KModifier` like `VAL` and `VAR` that apply to parameters, but only to constructor parameters? That way you write a spec that more closely resembles the desired code, instead of us inferring meaning only after the user writes twice as much code as is needed.

We could do this as an optimization, but it seems clumsy as the only way to accomplish the goal, since it's not how you think of the code.

swankjesse:
I think I want KotlinPoet to model the de-sugared `.kt` file. I’ll try to explain why, partly to explain it to you and mostly to explain it to myself. 

Suppose we want to emit this class:

```
class Name(first: String, last: String) {
  val first = first.trim()
  val last = last.trim()

  override fun toString() = "$first $last"
}
```

This code _almost_ gets you there.

```
    val name = TypeSpec.classBuilder("Name")
            .addParameter("first", String::class)
            .addParameter("last", String::class)
            .build())
        .addProperty(PropertySpec.builder("first", String::class)
            .initializer("first.trim()")
            .build())
        .addProperty(PropertySpec.builder("last", String::class)
            .initializer("last.trim()")
            .build())
        .addFun(FunSpec.builder("toString")
            .addModifiers(KModifier.OVERRIDE)
            .returns(String::class)
            .addStatement("return %S", "\$first \$last")
            .build())
        .build()
```

But it doesn’t quite. KotlinPoet isn’t yet smart enough to elide the return type, or to use an expression body when that’s possible. Instead it omits something more verbose:

```
class Name(first: String, last: String) {
  val first: String = first.trim()

  val last: String = last.trim()

  override fun toString(): String {
    return "$first $last"
  }
}
```

Here’s the interesting bit: I don’t _want_ user code to specify whether `toString()` is eligible for an expression body or not! Instead I want to either figure that out automatically, or not  do it at all. I want this because I think it makes it easier to make certain changes to the generating code. For example, suppose we wanted to tweak the logic:

```
        .addFun(FunSpec.builder("toString")
            .addModifiers(KModifier.OVERRIDE)
            .returns(String::class)
            .addStatement("if (first.isEmpty()) return last")
            .addStatement("if (last.isEmpty()) return first")
            .addStatement("return %S", "\$first \$last")
            .build())
```

We shouldn’t have to do the extra work of going from an expression body to a not-expression body. Generating code is hard enough as it is, it’s the tool’s job to format the declaration.

Next, suppose we decide we don’t need the `.trim()` calls. I think in this case the changes to the generated code should be identical to the previous code, just without the `.trim()` calls.

```
    val name = TypeSpec.classBuilder("Name")
            .addParameter("first", String::class)
            .addParameter("last", String::class)
            .build())
        .addProperty(PropertySpec.builder("first", String::class)
            .initializer("first")
            .build())
        .addProperty(PropertySpec.builder("last", String::class)
            .initializer("last")
            .build())
        .addFun(FunSpec.builder("toString")
            .addModifiers(KModifier.OVERRIDE)
            .returns(String::class)
            .addStatement("return %S", "\$first \$last")
            .build())
        .build()
```

Removing the `trim()` calls also makes the code eligible for the inline property declaration sugar. So if KotlinPoet is super smart it’ll do that automatically. Or if we decide we don’t want `.trim()` but we do want `.toTitleCase()`, then that should be a very small change.

If KotlinPoet models the sugared syntax, then the user needs to be aware of the sugar and when they are eligible for it. If KotlinPoet only models the desugared syntax, then that responsibility can be handed off to the framework.

I know that it’s more work to write the code that’ll get sugared away. But I think the purpose of writing it is to force you to define the complete model. Otherwise small behavior changes (adding or removing `.trim()`) lead to large model changes and I don’t like that.

I think this fits into the _seductive code_ thing I was complaining about recently. It sucks when you have to choose between concise, compact code and explicit modeling. In the case of the code generator I think it’s better to force you to put more stuff into the model, if only because then when you want to edit that model it will exist!

swankjesse:
... and one more thing – we can offer APIs to make it really easy to define pairs of parameters and properties. I think KotlinPoet should have a simple, fast API. But I think the resulting TypeSpec model should more closely represent the AST than the text you’d type to produce it.

rharter:
Good explanation. I think that makes a lot of sense. This should now be good to go.

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -37,21 +37,23 @@ class PropertySpec private constructor(builder: Builder) {
   val setter: FunSpec? = builder.setter
 
   @Throws(IOException::class)
-  internal fun emit(codeWriter: CodeWriter, implicitModifiers: Set<KModifier>) {
+  internal fun emit(codeWriter: CodeWriter, implicitModifiers: Set<KModifier>,
+                    withInitializer: Boolean = true,
+                    inline: Boolean = false) {
```
swankjesse: `inline` is also a keyword. Dangerous?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -76,6 +76,19 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     val previousStatementLine = codeWriter.statementLine
     codeWriter.statementLine = -1
 
+    val ctorProperties: Map<String, PropertySpec> =
```
swankjesse: supernit: we always say `constructor` not `ctor`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -76,6 +76,19 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+        if (primaryConstructor == null || primaryConstructor.parameters.isEmpty()) {
+          emptyMap<String, PropertySpec>()
+        } else {
+          propertySpecs
```
swankjesse: Might be just me, but I don’t find this particularly readable. Helper method?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -110,7 +123,21 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
-        primaryConstructor?.emitParameterList(codeWriter)
+        primaryConstructor?.let {
+          codeWriter.emit("(")
+          var firstParameter = true
+          it.parameters.forEachIndexed { index, param ->
+            if (!firstParameter) codeWriter.emit(",").emitWrappingSpace()
+            val property = ctorProperties[param.name]
+            if (property != null) {
+              property.emit(codeWriter, setOf(PUBLIC), withInitializer = false, inline = true)
```
swankjesse: nice

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -83,7 +83,41 @@ class KotlinPoetTest {
...
+            .addParameter("cheese", String::class)
+            .addParameter("cilantro", String::class)
+            .addParameter("lettuce", String::class)
+            .beginControlFlow("require(!cheese.isEmpty())")
+            .addStatement("%S", "cheese cannot be empty")
+            .endControlFlow()
+            .build())
+        .addProperty(PropertySpec.builder("cheese", String::class).initializer("cheese").build())
+        .addProperty(PropertySpec.varBuilder("cilantro", String::class).initializer("cilantro").build())
+        .addProperty(PropertySpec.builder("lettuce", String::class).initializer("lettuce.trim()").build())
```
swankjesse: nice

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -110,7 +123,21 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
         codeWriter.emitTypeVariables(typeVariables)
 
-        primaryConstructor?.emitParameterList(codeWriter)
+        primaryConstructor?.let {
+          codeWriter.emit("(")
+          var firstParameter = true
+          it.parameters.forEachIndexed { index, param ->
+            if (!firstParameter) codeWriter.emit(",").emitWrappingSpace()
```
JakeWharton: `index > 0`, drop the local

swankjesse:
I rebased & merged this.

swankjesse:
Thanks!

PR #87, opened 2017-05-24
Fix syntax for enum classes and overridden members.

swankjesse:


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -233,31 +227,42 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
     ANNOTATION(
+        "annotation class",
         emptySet(),
-        setOf(KModifier.PUBLIC, KModifier.ABSTRACT))
+        setOf(KModifier.PUBLIC, KModifier.ABSTRACT));
+
+    override fun toString(): String {
+      return super.toString()
+    }
```
JakeWharton: Delete

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -437,7 +437,7 @@ class FunSpec private constructor(builder: Builder) {
       val methodName = method.simpleName.toString()
       val funBuilder = FunSpec.builder(methodName)
 
-      funBuilder.addAnnotation(Override::class)
+      funBuilder.addModifiers(KModifier.OVERRIDE)
```
ronshapiro: Can you fix the javadoc above?
```diff
-An [Override] annotation will be added.
+An `override` modifier will be added.
```

swankjesse: Yeah, good catch. Fixed.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
swankjesse: Fixed!

PR #88, opened 2017-05-24
Invent a good-enough order for modifiers and test it.

swankjesse:


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -313,4 +313,72 @@ class KotlinPoetTest {
...
+            .addStatement("return %S", "a")
+            .build())
+        .build())
+    assertThat(source.toString()).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |import java.lang.String
+        |
+        |open class A {
+        |  protected operator infix open external fun get(v: String): String
```
JakeWharton: This hurts my brain

Issue #24, opened 2017-05-12
Replace java.util.regex with kotlin.text


JakeWharton:

Egorand:
Looks like this would require using `kotlin-stdlib-jre8`: http://stackoverflow.com/questions/37088679/kotlin-regex-named-groups-support

JakeWharton:
We don't need to use named groups. Indexed ones work fine when I tried to
convert previously.

PR #89, opened 2017-05-24
Replace java.util.regex with kotlin.text

Egorand:
Fix for #24

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -18,8 +18,6 @@ package com.squareup.kotlinpoet
 import java.io.IOException
 import java.io.StringWriter
 import java.lang.reflect.Type
-import java.util.regex.Matcher
-import java.util.regex.Pattern
 import javax.lang.model.element.Element
 import javax.lang.model.type.TypeMirror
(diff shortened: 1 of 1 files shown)


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -330,10 +328,10 @@ class CodeBlock private constructor(builder: CodeBlock.Builder) {
...
 
   companion object {
-    @JvmField internal val NAMED_ARGUMENT
-        = Pattern.compile("%(?<argumentName>[\\w_]+):(?<typeChar>[\\w]).*")
-    @JvmField internal val LOWERCASE
-        = Pattern.compile("[a-z]+[\\w_]*")
+    @JvmField internal val NAMED_ARGUMENT = Regex("%([\\w_]+):([\\w]).*")
+    @JvmField internal val LOWERCASE = Regex("[a-z]+[\\w_]*")
+    @JvmField internal val ARG_NAME = 1
+    @JvmField internal val TYPE_NAME = 2
```
JakeWharton: nit: these two can be marked `const`

Egorand: Makes sense

JakeWharton:
Thanks!

PR #93, opened 2017-05-25
Remove Filer writing and originating elements concept.

JakeWharton:
These stem from javac annotation processors inside which Kotlin cannot be generated.


max4t:
Hi, if you don't mind me asking : why removing the `writeTo(Filer)`?
I've created an annotation processor for android which generates kotlin files to `Filer.createResource(SOURCE_OUTPUT, ...)` and it works like a charm.
I was hoping to use `writeTo(Filer)` when switching to kotlinPoet

JakeWharton:
javac can't compile kotlin code so that method is useless.

max4t:
javac can't for sure but i'm using android studio 3 (canary) and the default kotlin buildchain compile the generated kotlin files in `Filer.createResource(SOURCE_OUTPUT, ...)`

JakeWharton:
Please provide a sample which demonstrates this behavior

max4t:
Here is a sample :
- In `app/build/generated/source/kapt/debug/com/example/extension/ViewMakerExtension.kt` (generated):
```kotlin
...
fun Maker.getDummy() = "Test String"
...
```
- In `app/src/main/java/com/example/MainActivity.kt`:
```kotlin
class MainActivity : AppCompatActivity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)
        (findViewById(R.id.txt_view) as TextView).text = Maker().getDummy()
    }
}
```

JakeWharton:
I mean an executable sample

max4t:
Here is the [https://github.com/max4t/KotlinGeneratingProcessor](sample)

JakeWharton:
Wow interesting! I've never used kapt. I'm surprised it does this. We
potentially might want to revert this then.

max4t:
Thanks that would be great.

It works because `kapt.generateStubs = true` in `app/build.gradle` otherwise you get an missing reference.
I don't remember exactly how it works but if i recall correctly kapt generated the references (the stubs) without content so that javac doesn't complain and then all is linked at the end or something like that

JakeWharton:
Hmm that's kapt1 though which is deprecated. Does it work with kapt3?

max4t:
It seems like by default kapt1 is used. kapt3 is available by adding `apply plugin: 'kotlin-kapt'`

hotchemi:
Hi @JakeWharton @max4t any update here?
I just found official Kotlin code generation example and it seems works well with the latest Kotlin and kapt ver. https://github.com/JetBrains/kotlin-examples/tree/master/gradle/kotlin-code-generation

PR #98, opened 2017-05-28
Rename TypeName.abstractEmit() to TypeName.emit().

swankjesse:


src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -34,7 +34,7 @@ internal fun <K, V> Map<K, List<V>>.toImmutableMultimap(): Map<K, List<V>> {
 internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V>
     = Collections.unmodifiableMap(LinkedHashMap(this))
 
-internal fun <T> Collection<T>.toImmutableList()
+internal fun <T> Collection<T>.toImmutableList() : List<T>
```
swankjesse: Prevents a warning on callsites because Kotlin can’t infer nullability otherwise.

PR #99, opened 2017-05-28
Implement support for basic expression bodies.

swankjesse:


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -50,13 +50,65 @@ import kotlin.reflect.KClass
...
+          return null // Argument doesn't match.
+        }
+        prefixArgCount++
+      }
+    }
+
+    // We found a prefix. Prepare the suffix as a result.
+    val resultFormatParts = ArrayList<String>()
+    if (firstFormatPart != null) {
+      resultFormatParts.add(firstFormatPart as String)
```
swankjesse: the `as String` here is something IntelliJ wanted, but I don’t understand why. Sees like the value should be a `String` and not a `String?` here.

JakeWharton: It's because it can't guarantee the second read will be non-null.

Try
```
firstFormatPart?.let {
  resultFormatPart.add(it)
}
```
to only read the var once.

JakeWharton:
Can't review on mobile. Will find a computer soon.

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -304,9 +304,7 @@ class KotlinPoetTest {
         |
         |import java.lang.String
         |
-        |fun String.shrink(): String {
-        |  return substring(0, length - 1)
-        |}
+        |fun String.shrink(): String = substring(0, length - 1)
```
vanniktech: The return type can be inferred, will this also be supported later?

swankjesse: It's difficult to do well. We’d need to either infer type information (probably impossible) or have users supply hints (awkward).

We could also make it something you configure manually. I don’t love that option either.

geralt-encore: What is the point of inferring types in generated code? It doesn't really add any value. Moreover,  I think that it is good in general for generated code to be as explicit as possible.

JakeWharton: Couldn't agree more about inferred types. There's no point in generated code.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -501,3 +509,5 @@ internal val String.isConstructor: Boolean
 
 internal val String.isAccessor: Boolean
   get() = this == GETTER || this == SETTER
+
+val EXPRESSION_BODY_PREFIX = CodeBlock.of("%[return ")
```
JakeWharton: `private`

PR #101, opened 2017-05-29
Generate valid annotation classes

geralt-encore:
According to https://kotlinlang.org/docs/reference/annotations.html:

- No `default` keyword

- Can have primary constructor

- Can have properties (only collapsed to primary constructor parameters tho. So at the moment it is possible to add a property to an annotation which won't be collapsed and will generate invalid Kotlin code. Not sure how to prevent it.)

- Can't have body
 
The only problem is that I haven't been able to combine property collapsing with default value yet. Seems like it doesn't play nicely at the moment. That's why I ignored `innerAnnotationInAnnotationDeclaration` test since it doesn't make sense without supplying a default value to annotation.


JakeWharton:
Can you rebase?

geralt-encore:
Sure. Rebased.

swankjesse:
Merged manually! Thanks!

Issue #86, opened 2017-05-23
KClass use is converter into java.* in code.


JakeWharton:
See tests for a lot of examples of this

JakeWharton:
Some interesting bits here.

`KClass` provides a simple name and a qualified name but no good way to convert it to the current internal representation unless we parse the qualified name like `bestGuess`.

We need to handle things like `Array<String?>::class` properly.

JakeWharton:
Might be blocked by https://youtrack.jetbrains.com/issue/KT-17912 and https://youtrack.jetbrains.com/issue/KT-18104

JakeWharton:
For now let's go with a `bestGuess`-style approach of using the `qualifiedName` and `simpleName` to try and infer the package and any outerclasses. While this may be wrong some of the time, it will be correct _most_ of the time which is better than the current behavior of being semantically wrong _all_ of the time.

JakeWharton:
Further complicated by the fact that I can write `Array<String>::class` but I can't get `String` as the parameter only `Array` yet I can't write `Array::class` to create a `ParameterizedTypeName`. I can write `List::class` though, but not `List<String>::class`.

Super weird...

PR #103, opened 2017-05-31
Extract Kotlin name from KClass instead of Java name.

JakeWharton:
Refs #86. Not the whole way there, but a first step.

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt
@@ -67,7 +67,7 @@ class ClassNameTest {
     assertBestGuessThrows("java.util.")
     assertBestGuessThrows("java..util.Map.Entry")
     assertBestGuessThrows("java.util..Map.Entry")
-    assertBestGuessThrows("java.util.Map..Entry")
+    assertBestGuessThrows("kotlin.collections.Map..Entry")
     assertBestGuessThrows("com.test.$")
     assertBestGuessThrows("com.test.LooksLikeAClass.pkg")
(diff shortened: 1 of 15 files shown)


JakeWharton:
Rebased

src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt
```diff
@@ -137,12 +137,12 @@ class ClassNameTest {
     } catch (expected: IllegalArgumentException) {
     }
 
-    try {
-      ClassName.get(Array<Any>::class)
-      fail()
-    } catch (expected: IllegalArgumentException) {
-    }
-
+    // TODO
```
swankjesse: ?

src/test/java/com/squareup/kotlinpoet/TypesEclipseTest.kt
```diff
@@ -36,6 +37,7 @@ import javax.lang.model.util.Types
 import javax.tools.DiagnosticCollector
 import javax.tools.JavaFileObject
 
+@Ignore("Not clear this test is useful to retain in the Kotlin world")
```
swankjesse: Yeah.

src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt
JakeWharton: See https://github.com/square/kotlinpoet/issues/86#issuecomment-305071978. This is the language leaking implementation details of the JVM 😢 .

I'm not sure what to do here yet...

PR #104, opened 2017-06-01
Use apply {} in Builder methods

Egorand:
Eliminates the need to do `return this` in Builder methods.


swankjesse:
Finally looking at this now. I do like that it’s less code. But I also feel outside my comfort zone trying to write idiomatic Kotlin. I’m going to merge it because it satisfies Shipley’s law, but I reserve the right to change my mind if somebody who writes more Kotlin tells me this is weird!

Egorand:
Haha, fair enough! In the end, what `apply` does is just "do whatever's in the brackets and return `this`". If there's a function doing that - I don't see a reason not to use it.

Issue #62, opened 2017-05-16
DSL for Kotlin generation


Egorand:
In addition to the APIs, it would be great to have a fluent DSL for defining Kotlin files.

burntcookie90:
I had written a DSL for javapoet, it was nice until it wasn't. I'm not too much a fan of it.

JakeWharton:
I think we can do a few niceties like #39 and then maybe `.apply` and `.let` will get us 90% of what a DSL would without the overhead trying to project the API into it.

EddieRingle:
> a fluent DSL for defining Kotlin files

I chuckled a little at this since I imagine most would just end up re-inventing Kotlin.

max4t:
> 90% of what a DSL would

Minus the readability


By overloading the methods in the builder, the example from the Readme could look like this :
```kotlin
val greeterClass = ClassName.get("", "Greeter")
val kotlinFile = KotlinFile.builder("", "HelloWorld") {
    addType("Greeter") {
        primaryConstructor() {
            addParameter("name", String::class)
        }
        addProperty("name", String::class) {
            initializer("name")
        }
        addFun("greet") {
            addStatement("println(%S)", "Hello, \$name")
        }
    }
    addFun("main") {
        addParameter("args", TypeName.get(Array<String>::class))
        addStatement("%T(args[0]).greet()", greeterClass)
    }
}

kotlinFile.writeTo(System.out)
```
which, i think, is much less "dense"

Of course, it would require to overloading the methods to match the builder of the parameter with something like this:
```kotlin
fun addType(name: String, block: TypeSpec.Builder.() -> Unit) = addType(
        TypeSpec.classBuilder(name).apply(block).build())
```

tieskedh:
Maybe you can add the function 

infix fun String.of(clazz: KClazz) 
= pairOf(this, clazz) 

and make  addFun accept a vararg of these pairs, then you can call the method like:

AddFun("main", "args" of Array<String>::class){
     //code
} 

 Almost like I did here: https://github.com/tieskedh/KotlinPoetDSL

But I made it more advanced, too also include

AddFun("main", "args" varArg String::class){
     //code
}

wakingrufus:
I have started implementing a dsl wrapper for poet in my project. If I get it working, I will open a PR to contribute it upstream here.

jonnyzzz:
Yet another comment from #435, answering also to @breandan request ;) 
[...]

PR #106, opened 2017-06-03
Dsl builder as extensions

max4t:
"Proof of concept" #62 

Mostly syntactic sugar but now, this works:
```kotlin
val greeterClass = ClassName.get("", "Greeter")
val kotlinFile = KotlinFile.builder("", "HelloWorld").apply {
    addClassType("Greeter") {
        primaryConstructor() {
            addParameter("name", String::class)
        }
        addProperty("name", String::class) {
            initializer("name")
        }
        addFun("greet") {
            addStatement("println(%S)", "Hello, \$name")
        }
    }
    addFun("main") {
        addParameter("args", TypeName.get(Array<String>::class))
        addStatement("%T(args[0]).greet()", greeterClass)
    }
}.build()

kotlinFile.writeTo(System.out)
```

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/DslUtil.kt
+++ b/src/main/java/com/squareup/kotlinpoet/DslUtil.kt
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+import java.lang.reflect.Type
+import kotlin.reflect.KClass
+
+// AnnotationSpec.Builder.addMember
+fun AnnotationSpec.Builder.addMember(name: String, block: CodeBlock.Builder.() -> Unit) =
+        addMember(name, CodeBlock.Builder().apply(block).build())
+
+// FunSpec.Builder.addKdoc
+fun FunSpec.Builder.addKdoc(block: CodeBlock.Builder.() -> Unit) = addKdoc(
+        CodeBlock.Builder().apply(block).build())
+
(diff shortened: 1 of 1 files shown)


Egorand:
Looks great! 

How challenging would it be to maintain both the builders API and the DSL? Also, I'm not sure about the usage requirements for kotlinpoet: if it's primarily designed to be used from Java - API is the better way to go, and DSL doesn't make much sense; if Kotlin - I'd vote for going DSL-first. @JakeWharton, @swankjesse, what are your thoughts?

swankjesse:
This is a nice DSL. I’m not completely sure that we want any DSL for this library.

trevjonez:
you might consider doing a separate module for the purposes of a dsl consumption.

something like this has already been done for javapoet. https://github.com/agrosner/KPoet

JakeWharton:
That DSL is practically unreadable though, and is an example of what we're
not trying to do.

max4t:
These DSL extensions weren't meant to change or replace the current builder API (that would be too much work to maintain and for users to change their existing code).
The idea was to provide a different syntax for the same method not to add new method:
```kotlin
addType(TypeSpec.classBuilder("MyClass")
        ....
        .build()
)
```
or
```kotlin
addClassType("MyClass") {
        ....
}
```
Too much DSL would be harder to maintain and lose in flexibility and readability. KPoet may be good for personal use, i'm not sure it is as a library.
The DSL should remain as minimal as possible to be usable for kotlin users

max4t:
Basically, all i did was map the `addXXX(YYYSpec)` to the `YYYSpec` builders, nothing more. So maintainability should be too much of a hassle.

JakeWharton:
Yes. That wasn't a comment on your PR implementation itself! I said this on
one of the issues, but we've been focusing on correctness of the output
before dealing with the niceties of the public API.

max4t:
The DSL extensions i provided were not meant to replace the current API or change it, that would require too much work from the maintainers

Egorand:
@max4t I guess kotlinpoet is also expected to be used from Java, where all lambdas with receivers will translate into either parameterized lambdas on Java 8, or anonymous classes on Java 7. The advantage of builders here is that they're used almost in the same way from both Java and Kotlin.

max4t:
@JakeWharton I know and agree. If implemented, the DSL should be minimal and not hinder the core of the library.
@Egorand which is why the current API shouldn't be replaced. However, this library is meant to generate kotlin source file which is most likely done by kotlin users (just like java users most likely generate java source) so providing a syntax that uses the feature of this language seems useful to me.

JakeWharton:
Going to close this for now. It's linked to the open issue so we can recover it if need be. But as I said before, we're focused on correctness before any syntactic niceties.

PR #107, opened 2017-06-04
Use extension and invoke functions for creating TypeName and subclasses.

### LINKED ISSUE #36: Hide static factories which accept MutableMap from public API
e.g., `TypeName.get(TypeMirror, MutableMap)`

JakeWharton:
https://youtrack.jetbrains.com/issue/KT-15286 makes this a bit annoying, currently. Alternatively we could make these top-level functions and not worry about Java interop as much...

Closes #36.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -119,12 +119,13 @@ class ClassName private constructor(
   }
 
   companion object {
-    @JvmStatic fun get(clazz: Class<*>): ClassName {
-      require(!clazz.isPrimitive) { "primitive types cannot be represented as a ClassName" }
-      require(Void.TYPE != clazz) { "'void' type cannot be represented as a ClassName" }
-      require(!clazz.isArray) { "array types cannot be represented as a ClassName" }
+    @JvmStatic @JvmName("get")
+    fun Class<*>.asClassName(): ClassName {
+      require(!isPrimitive) { "primitive types cannot be represented as a ClassName" }
+      require(Void.TYPE != this) { "'void' type cannot be represented as a ClassName" }
+      require(!isArray) { "array types cannot be represented as a ClassName" }
       val names = mutableListOf<String>()
-      var c = clazz
+      var c = this
       while (true) {
         names += c.simpleName
         val enclosing = c.enclosingClass ?: break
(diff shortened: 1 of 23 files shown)


src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -69,29 +69,30 @@ class TypeVariableName private constructor(
...
-    @JvmStatic fun get(name: String): TypeVariableName {
-      return TypeVariableName.of(name, emptyList())
-    }
+    @JvmStatic @JvmName("get")
+    operator fun invoke(name: String) = TypeVariableName.of(name, emptyList())
 
     /** Returns type variable named `name` with `bounds`.  */
-    @JvmStatic fun get(name: String, vararg bounds: TypeName): TypeVariableName {
+    @JvmStatic @JvmName("get")
+    operator fun invoke(name: String, vararg bounds: TypeName): TypeVariableName {
```
swankjesse: can bounds have a default? Or is that not a thing for varargs?

src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -137,12 +138,10 @@ class ClassName private constructor(
       return ClassName(names)
     }
 
-    @JvmStatic fun get(clazz: KClass<*>): ClassName {
-      val qualifiedName = clazz.qualifiedName
-      if (qualifiedName == null) {
-        throw IllegalArgumentException("$clazz cannot be represented as a TypeName")
-      }
-      return ClassName.bestGuess(qualifiedName)
+    @JvmStatic @JvmName("get")
```
ronshapiro: Do you care about the JVM name here? For code in another JVM language to have a KClass sounds unlikely

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -315,10 +316,10 @@ class CodeBlock private constructor(
...
       when (o) {
         is TypeName -> return o
-        is TypeMirror -> return TypeName.get(o)
-        is Element -> return TypeName.get(o.asType())
-        is Type -> return TypeName.get(o)
-        is KClass<*> -> return TypeName.get(o)
+        is TypeMirror -> return o.asTypeName()
+        is Element -> return o.asType().asTypeName()
+        is Type -> return o.asTypeName()
+        is KClass<*> -> return o.asTypeName()
```
ronshapiro: Is `java.lang.Class` intentionally omitted here?

JakeWharton: Class is a Type and asTypeName will do the right thing for 'em

src/main/java/com/squareup/kotlinpoet/ClassName.kt
JakeWharton: True. It was just a mechanical refactor to put them on everything. In theory something could give you a KClass in Java just like you can have a Class in Kotlin

PR #108, opened 2017-06-05
Change arrayComponent to extension method.

JakeWharton:


src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -227,10 +227,10 @@ abstract class TypeName internal constructor(
       }
     }
 
-    /** Returns the array component of `type`, or null if `type` is not an array.  */
-    @JvmStatic fun arrayComponent(type: TypeName): TypeName? {
-      return if (type is ParameterizedTypeName && type.rawType == ARRAY)
-        type.typeArguments.single() else
+    /** Returns the array component of this [TypeName], or null if it is not an array.  */
+    @JvmStatic fun TypeName.arrayComponent(): TypeName? {
```
swankjesse: Weird to use an extension method for a class we declare?

JakeWharton: Oh, true. We do this in JavaPoet as well. Can change.

JakeWharton: Actually I think this used to be on `ArrayTypeName`?

swankjesse: Yep. For Kotlin arrays aren’t special, right?

JakeWharton: Damnit. Yes. I need to start thinking about what I'm doing rather than blindly coding.

PR #112, opened 2017-06-05
Make varargs a property of the parameter.

### LINKED ISSUE #110: Re-tool vararg support
Kotlin lets you have multiple in a single function parameter list.

JakeWharton:
This removes the array specialization from the type and allows for multiple and different positions in the list.

Closes #110.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -19,7 +19,7 @@ import com.squareup.kotlinpoet.ClassName.Companion.asClassName
 import com.squareup.kotlinpoet.FunSpec.Companion.CONSTRUCTOR
 import com.squareup.kotlinpoet.FunSpec.Companion.GETTER
 import com.squareup.kotlinpoet.FunSpec.Companion.SETTER
-import com.squareup.kotlinpoet.TypeName.Companion.arrayComponent
+import com.squareup.kotlinpoet.KModifier.VARARG
 import com.squareup.kotlinpoet.TypeName.Companion.asTypeName
 import com.squareup.kotlinpoet.TypeVariableName.Companion.asTypeVariableName
 import java.io.IOException
@@ -44,7 +44,6 @@ class FunSpec private constructor(builder: Builder) {
   val receiverType = builder.receiverType
   val returnType = builder.returnType
   val parameters = builder.parameters.toImmutableList()
-  val varargs = builder.varargs
   val exceptions = builder.exceptions.toImmutableList()
   val code = builder.code.build()
   val defaultValue = builder.defaultValue
@@ -53,17 +52,9 @@ class FunSpec private constructor(builder: Builder) {
     require(code.isEmpty() || !builder.modifiers.contains(KModifier.ABSTRACT)) {
       "abstract function ${builder.name} cannot have code"
     }
-    require(!builder.varargs || lastParameterIsArray(builder.parameters)) {
-      "last parameter of varargs function ${builder.name} must be an array"
-    }
     require(name != SETTER || parameters.size == 1) {
       "$name must have exactly one parameter"
     }
-  }
-
-  private fun lastParameterIsArray(parameters: List<ParameterSpec>): Boolean {
-    return parameters.isNotEmpty()
-        && parameters[parameters.size - 1].type.arrayComponent() != null
   }
 
   internal fun parameter(name: String): ParameterSpec? {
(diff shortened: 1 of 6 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -313,19 +297,17 @@ class FunSpec private constructor(builder: Builder) {
...
-    fun addParameter(name: String, type: TypeName, vararg modifiers: KModifier)
-        = addParameter(ParameterSpec.builder(name, type, *modifiers).build())
+    @JvmOverloads
+    fun addParameter(name: String, type: TypeName, vararg modifiers: KModifier, varargs: Boolean = false)
+        = addParameter(ParameterSpec.builder(name, type, *modifiers).varargs(varargs).build())
 
-    fun addParameter(name: String, type: Type, vararg modifiers: KModifier)
-        = addParameter(name, type.asTypeName(), *modifiers)
+    @JvmOverloads
+    fun addParameter(name: String, type: Type, vararg modifiers: KModifier, varargs: Boolean = false)
```
swankjesse: shouldn’t varargs just be one of the modifiers? `KModifier.VARARG` ?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -443,7 +425,12 @@ class FunSpec private constructor(builder: Builder) {
 
       funBuilder.returns(method.returnType.asTypeName())
       funBuilder.addParameters(ParameterSpec.parametersOf(method))
-      funBuilder.varargs(method.isVarArgs)
+      if (method.isVarArgs) {
```
swankjesse: nice.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1051,6 +1050,50 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun varargsNotLast() {
```
swankjesse: Anywhere varargs is cool

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
JakeWharton: Yeah I guess that's the easiest way. Removes a lot of specialization!

JakeWharton: You already had this!

JakeWharton:
PTAL

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1051,6 +1051,50 @@ class TypeSpecTest {
...
+        .build()
+    assertThat(toString(taqueria)).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |import java.lang.Runnable
+        |import kotlin.Boolean
+        |import kotlin.Int
+        |
+        |class Taqueria {
+        |  fun prepare(workers: Int, vararg jobs: Runnable, start: Boolean) {
```
Egorand: Multiple varargs are prohitibed, this function won't compile.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1051,6 +1051,50 @@ class TypeSpecTest {
...
+        .build()
+    assertThat(toString(taqueria)).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |import java.lang.Runnable
+        |import kotlin.Int
+        |import kotlin.String
+        |
+        |class Taqueria {
+        |  fun prepare(workers: Int, vararg jobs: Runnable, vararg names: String) {
```
Egorand: Multiple varargs are prohibited, this function won't compile.

src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -83,6 +77,7 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
     internal val annotations = mutableListOf<AnnotationSpec>()
     internal val modifiers = mutableListOf<KModifier>()
     internal var defaultValue: CodeBlock? = null
+    internal var varargs = false
```
swankjesse: Unused?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
JakeWharton: I swear this worked long ago back when we started using Kotlin. Removed the test.

Egorand: Would it make sense to check and throw if second vararg parameter is added?

JakeWharton: The compiler will enforce it so i'm not sure it's worth duplicating. Especially since it'd have to be done on every object that collects `ParameterSpec`s

Issue #100, opened 2017-05-28
Using @Throws annotation instead of Java throws syntax


geralt-encore:
I was checking out sources and stumbled upon usage of the invalid in Kotlin `throws` syntax. Before fixing it I decided to open the issue since it probably will require api changes. 
So there are several options:

- Completely remove `addException` apis from `FunSpec` since this syntax is not a part of the language and delegate it to `AnnotationSpec`

- Leave current api unchanged and convert supplied `exceptions` to `AnnotationSpec` internally

- Create a new api

What do you think? I would be happy to fix it.

swankjesse:
I think I like #2 best.

JakeWharton:
Since exceptions are extremely uncommon, I lean toward 1. Especially since
the library could be used to generate Kotlin targeting JS or native.

burntcookie90:
I think going with the first option works better, aligns with the actual language spec as well. "Morphing" from a thrown exception to an annotation would be a bit confusing.

vanniktech:
This can be closed, right?

PR #117, opened 2017-06-11
throws cleanup

burntcookie90:
PR for the first option discussed in #100

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -235,9 +235,5 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
     @JvmStatic fun builder(type: KClass<*>): Builder {
       return builder(type.asClassName())
     }
-
-    @JvmStatic fun throwsBuilder(): Builder {
-      return builder(Throws::class)
-    }
   }
 }
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -235,5 +235,9 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
     @JvmStatic fun builder(type: KClass<*>): Builder {
       return builder(type.asClassName())
     }
+
+    @JvmStatic fun throwsBuilder(): Builder {
+      return builder(Throws::class)
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -474,38 +474,60 @@ class TypeSpecTest {
...
-        |  fun throwOne() throws IOException {
+        |  @Throws(
+        |      exceptionClasses = IOException::class
+        |  )
+        |  fun throwOne() {
         |  }
         |
-        |  fun throwTwo() throws IOException, SourCreamException {
+        |  @Throws(
+        |      exceptionClasses = *arrayOf(IOException::class, SourCreamException::class)
```
JakeWharton: Is this star needed?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -474,38 +474,60 @@ class TypeSpecTest {
...
+        |  abstract fun abstractThrow()
         |
-        |  external fun nativeThrow() throws IOException
+        |  @Throws(
+        |      exceptionClasses = IOException::class
+        |  )
+        |  external fun nativeThrow()
         |}
         |""".trimMargin())
+
```
JakeWharton: Nit: revert

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -474,38 +474,60 @@ class TypeSpecTest {
     val taco = TypeSpec.classBuilder("Taco")
         .addModifiers(KModifier.ABSTRACT)
         .addFun(FunSpec.builder("throwOne")
-            .addException(IOException::class)
+            .addAnnotation(AnnotationSpec.throwsBuilder()
+                .addMember("exceptionClasses", "%T::class", IOException::class.asClassName())
```
JakeWharton: This can be 'value', right? So it emits nice and short.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -235,5 +235,9 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
     @JvmStatic fun builder(type: KClass<*>): Builder {
       return builder(type.asClassName())
     }
+
+    @JvmStatic fun throwsBuilder(): Builder {
```
JakeWharton: Not sure we need this. It's about the same as `builder<Throws>()` which we could add support for. Even `builder(Throws::class)` is short.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
burntcookie90: ah yeah, didn't realize the `value` name was special cased in the lib

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
burntcookie90: only if we were using the named parameter, switching it to `value` means we don't need to wrap in `*arrayOf()`

PR #118, opened 2017-06-12
primary constructors w/ modifiers and annotations

### LINKED ISSUE #116: Missing ability to generate internal constructor
Should be able to emit the following

```kotlin
class Foo @Inject internal constructor() { }
```

burntcookie90:
Adds the ability to emit primary constructors with modifiers and
annotations, using the `constructor` keyword.

Fixes #116

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -99,6 +99,26 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitWhereBlock(typeVariables)
 
         primaryConstructor?.let {
+          var useKeyword = false
+          var emittedAnnotations = false
+
+          if (it.annotations.isNotEmpty()) {
+            codeWriter.emit(" ")
+            codeWriter.emitAnnotations(it.annotations, true)
+            useKeyword = true
+            emittedAnnotations = true
+          }
+
+          if (it.modifiers.isNotEmpty()) {
+            if (!emittedAnnotations) codeWriter.emit(" ")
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -98,6 +98,24 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          var useKeyword = false
+          val emittedAnnotations = if (it.annotations.isNotEmpty()) {
+            codeWriter.emit(" ")
+            codeWriter.emitAnnotations(it.annotations, true)
+            useKeyword = true
+            true
+          } else false
+
+          if (it.modifiers.isNotEmpty()) {
+            if(!emittedAnnotations) codeWriter.emit(" ")
```
JakeWharton: Nit: space after if

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -98,6 +98,24 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitTypeVariables(typeVariables)
 
         primaryConstructor?.let {
+          var useKeyword = false
+          val emittedAnnotations = if (it.annotations.isNotEmpty()) {
```
JakeWharton: Just set this to the boolean inside the if. And then the two ifs check the boolean. No gross if expression nonsense

JakeWharton:
@burntcookie90 ~kick~ i mean ping

burntcookie90:
Whoopsie! Will get this in tomorrow, sorry :smile:

swankjesse:
Nice!

Issue #119, opened 2017-06-12
Cannot add property to companion object


mkpowers:
Attempting to add a property to a companion object causes this error: 

`Exception in thread "main" java.lang.IllegalStateException: cannot add property val ID: kotlin.String = "id" to COMPANION`

The exception is being thrown on line 408 of TypeSpec.kt:
`check(kind == Kind.CLASS || kind == Kind.ENUM || kind == Kind.ANNOTATION) {
        "cannot add property $propertySpec to $kind"
      }`

Is this intentional? If it is, I'm not sure how one is meant to add properties to a companion object.

Thanks! Love the library.

JakeWharton:
Probably a simple bug on our part. Thanks for reporting.

PR #121, opened 2017-06-13
#119 Allow to add properties to objects, companions and interfaces

geralt-encore:
https://github.com/square/kotlinpoet/issues/119


swankjesse:
Awesome.

PR #123, opened 2017-06-14
Fix Class not calling through to superclass constructor when superclass has a no-arg constructor.

### LINKED ISSUE #120: Class not calling through to superclass constructor when superclass has a no-arg constructor
`val typeBuilder = TypeSpec.classBuilder("Foo").superclass(Bar::class)`

This generates `class Foo : Bar`, which does not compile in Kotlin 1.1.2. I read through the documentation and the TypeSpecTests to try to get some insight, but I found nothing that indicates that I'm omitting a function call that adds the superclass constructor call.

vanniktech:
Fixes #120

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -113,6 +113,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
         val extendsTypes: List<TypeName>
         val implementsTypes: List<TypeName>
+
         if (kind == Kind.INTERFACE) {
           extendsTypes = superinterfaces
           implementsTypes = emptyList()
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -113,6 +113,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
         val extendsTypes: List<TypeName>
         val implementsTypes: List<TypeName>
```
JakeWharton: Is the distinction between these necessary? Seems like a carryover from JavaPoet that has no place here.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -111,20 +111,20 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
           codeWriter.emit(")")
         }
 
-        val extendsTypes: List<TypeName>
-        val implementsTypes: List<TypeName>
+        val superTypes: List<Pair<TypeName, String>>
+
         if (kind == Kind.INTERFACE) {
```
JakeWharton: This whole conditional can be dropped now

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -111,20 +111,20 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
           codeWriter.emit(")")
         }
 
-        val extendsTypes: List<TypeName>
-        val implementsTypes: List<TypeName>
+        val superTypes: List<Pair<TypeName, String>>
```
JakeWharton: Maybe consider `List<CodeBlock>` and do `CodeBlock.of("%T", type)` or `CodeBlock.of("%T()", type)`.

PR #126, opened 2017-06-15
Fix syntax for declaring multiple generic type constraints

### LINKED ISSUE #125: Incorrect syntax when declaring multiple constraints for a generic type parameter
`TypeVariableName("T", Comparator::class, Serializable::class)` declared for a function will produce the following signature:

```kotlin
fun <T : Comparator & Serializable> getComparator(): T
```

which doesn't compile. Correct Kotlin syntax is:

```kotlin
fun <T> getComparator(): T where T : Comparator, T : Serializable
```

Egorand:
Fixes #125. Will this need more unit tests? Can you think of any specific scenarios?

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -158,22 +158,44 @@ internal class CodeWriter @JvmOverloads constructor(
    * everywhere else bounds are omitted.
    */
   @Throws(IOException::class)
-  fun emitTypeVariables(typeVariables: List<TypeVariableName>) {
+  fun emitTypeVariables(
+      typeVariables: List<TypeVariableName>,
+      emitSignature: (() -> Unit)? = null) {
     if (typeVariables.isEmpty()) return
 
     emit("<")
+    val withMultipleBounds = mutableListOf<TypeVariableName>()
     var firstTypeVariable = true
     for (typeVariable in typeVariables) {
       if (!firstTypeVariable) emit(", ")
       emitCode("%L", typeVariable.name)
-      var firstBound = true
-      for (bound in typeVariable.bounds) {
-        emitCode(if (firstBound) " : %T" else " & %T", bound)
-        firstBound = false
+      if (typeVariable.bounds.size > 1) {
+        withMultipleBounds += typeVariable
+      } else if (typeVariable.bounds.size == 1) {
+        emitCode(" : %T", typeVariable.bounds[0])
       }
       firstTypeVariable = false
     }
     emit(">")
+    emitSignature?.invoke()
+    if (withMultipleBounds.isNotEmpty()) {
+      emit(" ")
+      emitWhereBlock(withMultipleBounds)
+    }
+  }
+
+  @Throws(IOException::class)
+  internal fun emitWhereBlock(typeVariables: List<TypeVariableName>) {
+    emit("where ")
+    var firstBound = true
+    for (typeVariable in typeVariables) {
+      for (bound in typeVariable.bounds) {
+        if (!firstBound) emit(", ")
+        emitCode("%L", typeVariable.name)
+        emitCode(" : %T", bound)
+        firstBound = false
+      }
+    }
   }
 
   @Throws(IOException::class)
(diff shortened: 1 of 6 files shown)


src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -158,22 +158,44 @@ internal class CodeWriter @JvmOverloads constructor(
...
+      if (typeVariable.bounds.size > 1) {
+        withMultipleBounds += typeVariable
+      } else if (typeVariable.bounds.size == 1) {
+        emitCode(" : %T", typeVariable.bounds[0])
       }
       firstTypeVariable = false
     }
     emit(">")
+    emitSignature?.invoke()
+    if (withMultipleBounds.isNotEmpty()) {
```
swankjesse: would this be simpler as two calls? One call `emitTypeVariables()` and another, `emitWhereBlock()` ? Custom control flow is a very powerful construct, and I prefer to use simpler constructs wherever we can

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -77,10 +77,39 @@ class FunSpec private constructor(builder: Builder) {
     }
 
     if (typeVariables.isNotEmpty()) {
-      codeWriter.emitTypeVariables(typeVariables)
-      codeWriter.emit(" ")
+      codeWriter.emitTypeVariables(typeVariables) {
```
swankjesse: ... this 8 line block would become a different 6 line block, with no repeated call to `emitSignature`:

```
if (typeVariables.isNotEmpty()) {
  codeWriter.emitTypeVariables(typeVariables);
  codeWriter.emit(" ");
}
emitSignature(codeWriter, enclosingName);
codeWriter.emitWhereBlock(typeVariables);
```

I think it’s more boring an easier to follow

src/test/java/com/squareup/kotlinpoet/TypeVariableNameTest.kt
```diff
@@ -0,0 +1,106 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import java.io.Serializable
+
+class TypeVariableNameTest {
```
swankjesse: Nice tests

src/test/java/com/squareup/kotlinpoet/TypeVariableNameTest.kt
```diff
@@ -0,0 +1,106 @@
...
+        .addTypeVariable(TypeVariableName("T", Serializable::class, Runnable::class))
+        .returns(TypeVariableName("T").asNullable())
+        .addStatement("return null")
+        .build()
+    assertThat(funSpec.toString()).isEqualTo("""
+      |fun <T> foo(): T? where T : java.io.Serializable, T : java.lang.Runnable = null
+      |""".trimMargin())
+  }
+
+  @Test fun twoTypeVariablesTwoBoundsEach() {
```
swankjesse: supercool

Egorand:
Thanks for the feedback @swankjesse!
I wasn't sure about the API, passing `emitSignature()` as a block looked a bit fishy. The solution you suggest is good, but the downside is that currently every spec that needs to print type variables has to call both `emitTypeVariables()` and `emitWhereBlock()` - you can see that I had to modify it in both `FunSpec` and `TypeSpec`. I updated the KDoc to make this dependency clear, but I think it still isn't ideal. Any ideas?

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -166,16 +168,35 @@ internal class CodeWriter @JvmOverloads constructor(
...
     emit(">")
   }
 
+  /**
+   * Emit a `where` block containing type bounds for each type variable that has at least two
+   * bounds.
+   */
+  @Throws(IOException::class)
+  fun emitWhereBlock(typeVariables: List<TypeVariableName>) {
+    if (typeVariables.isEmpty()) return
```
swankjesse: this clause isn’t necessary, right?

swankjesse:
Because this is just implementation details – not public API – we can start with this approach which is simpler but requires more careful use. If we want to change our minds later, we can.

For public APIs its much more interesting!

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
Egorand: Yep, not really

Issue #124, opened 2017-06-15
Don't emit empty class bodies


JakeWharton:
```kotlin
class Foo {
}
```

geralt-encore:
Isn't  it already implemented [here](https://github.com/square/kotlinpoet/blob/master/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt#L81)? `companionObject`, `enumConstants` and `initializerBlock` should be probably added as well. Should I make a PR?

JakeWharton:
It doesn't work. Try adding a primary constructor and a property initialized to the constructor.

You get

```
class Foo(val bar: Bar) {
}
```

vanniktech:
This can be closed, right?

PR #127, opened 2017-06-15
Don't emit empty class bodies

Egorand:
Fix for #124


Egorand:
Indeed, this fails with a property declared inside the primary constructor, as pointed out by @JakeWharton. I'll take a look later and submit the fix.

Issue #128, opened 2017-06-15
Add Iterable<CodeBlock>.joinToCode helper.


JakeWharton:
This helps build argument lists the like. Should include optional prefix and suffix strings (and maybe blocks?).

PR #129, opened 2017-06-15
Iterable<CodeBlock>.joinToCode

geralt-encore:
https://github.com/square/kotlinpoet/issues/128

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -382,17 +382,5 @@ class CodeBlock private constructor(
 
 @JvmOverloads
 fun Iterable<CodeBlock>.joinToCode(separator: CharSequence = ", ", prefix: CharSequence = "",
-                                   suffix: CharSequence = "" ): CodeBlock {
-  val builder = StringBuilder()
-
-  builder.append(prefix)
-  forEachIndexed { index, block ->
-    if (index != 0) {
-      builder.append(separator)
-    }
-    builder.append(block.toString())
-  }
-  builder.append(suffix)
-
-  return CodeBlock.of(builder.toString())
-}
+                                   suffix: CharSequence = "" ) =
+    CodeBlock.of(map { it.toString() }.joinToString(separator, prefix, suffix))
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -379,3 +379,20 @@ class CodeBlock private constructor(
         = c == '%' || c == '>' || c == '<' || c == '[' || c == ']' || c == 'W'
   }
 }
+
+@JvmOverloads
+fun Iterable<CodeBlock>.joinToCode(separator: CharSequence = ", ", prefix: CharSequence = "",
```
Egorand: Would something like `this.map { it.toString() }.joinToString()` work?

geralt-encore: It should work. Didn't think about it in a first place...

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -379,3 +379,8 @@ class CodeBlock private constructor(
...
+                                   suffix: CharSequence = "" ) =
+    CodeBlock.of(map { it.toString() }.joinToString(separator, prefix, suffix))
```
JakeWharton: Calling `toString` prevents the ability to import types in these blocks.

It should be something like
```kotlin
val blocks = toTypedArray<CodeBlock>()
return CodeBlock.of(repeat("%L", blocks.size).joinToString(separator, prefix, suffix), *blocks)
```

geralt-encore: I see. Wasn't sure about it. Thanks for pointing out!

geralt-encore: `toTypedArray` is available for `Collection` but not for `Iterable`. I've changed extension type to `Collection` for now. If it is needed I can do revert it back to `Iterable` and convert to array manually.

JakeWharton:
Merged manually as c1abb0f614705e7ffa5b3b444e1438051dcfb152

PR #137, opened 2017-06-22
Fix generating arrays as AnnotationSpec params

### LINKED ISSUE #135: Incorrect array generation for AnnotationSpec params
Example:

```kotlin
@Test fun defaultAnnotation() {
  val name = AnnotationSpecTest.IsAnnotated::class.java.canonicalName
  val element = compilation.elements.getTypeElement(name)
  val annotation = AnnotationSpec.get(element.annotationMirrors[0])

  val taco = TypeSpec.classBuilder("Taco")
      .addAnnotation(annotation)
      .build()
  assertThat(toString(taco)).isEqualTo("""
        |package com.squareup.tacos
        |
        |import com.squareup.kotlinpoet.AnnotationSpecTest
        |import java.lang.Override
        |import kotlin.Double
        |import kotlin.Float
        |
        |@AnnotationSpecTest.HasDefaultsAnnotation(
        |    o = AnnotationSpecTest.Breakfast.PANCAKES,
        |    p = 1701,
        |    f = 11.1,
        |    m = {
        |        9,
        |        8,
        |        1
        |    },
        |    l = Override::class,
        |    j = @AnnotationSpecTest.AnnotationA,
        |    q = @AnnotationSpecTest.AnnotationC("bar"),
        |    r = {
        |        Float::class,
        |        Double::class
        |    }
        |)
        |class Taco
        |""".trimMargin())
}
```

`{}` won't work in Kotlin. It has to be `arrayOf()`, `intArrayOf()` or something different, hence `AnnotationSpec` should be aware of the type of the parameter. Would it make sense to capture this info through `addMember()`?

```kotlin
fun addMember(name: String, type: ClassName, format: String, vararg args: Any)

fun addMember(name: String, type: ClassName, codeBlock: CodeBlock)
```

Based on the type, and in case there are multiple values for the same name, the builder will be able to generate correct `CodeBlock` for the array and place it as the first element in the members list.

Please share your thoughts. I might be able to work out a PR this week.

Egorand:
Fixes #135

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -87,11 +87,9 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
     codeWriter.emit(whitespace)
     if (values.size > 1) {
       codeWriter.indent(2)
-      var first = true
       for (i in 1 until values.size) {
-        if (!first) codeWriter.emit(memberSeparator)
+        if (i > 1) codeWriter.emit(memberSeparator)
         codeWriter.emitCode(values[i])
-        first = false
       }
       codeWriter.unindent(2)
       codeWriter.emit(whitespace + ")")
@@ -227,7 +225,7 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
       is Int -> IntArray::class.asTypeName()
       is Long -> LongArray::class.asTypeName()
       is Short -> ShortArray::class.asTypeName()
-      else -> kotlin.Array<Any>::class.asTypeName()
+      else -> ParameterizedTypeName.get(ARRAY, ANY)
     }
   }
 
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -16,6 +16,7 @@
 package com.squareup.kotlinpoet
 
 import com.squareup.kotlinpoet.ClassName.Companion.asClassName
+import com.squareup.kotlinpoet.TypeName.Companion.asTypeName
 import java.io.IOException
 import java.io.StringWriter
 import java.lang.reflect.Array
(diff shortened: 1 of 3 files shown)


src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -236,8 +236,9 @@ class TypeSpecTest {
         .addFun(FunSpec.builder("fooBar")
             .addModifiers(KModifier.PUBLIC, KModifier.ABSTRACT)
             .addAnnotation(AnnotationSpec.builder(headers)
-                .addMember("value", "%S", "Accept: application/json")
-                .addMember("value", "%S", "User-Agent: foobar")
+                .addMemberArrayElement("value", Array<String>::class, "%S",
```
JakeWharton: For what it's worth, the generic being specified here is lost. We only receive `KClass<Array>` not `KClass<Array<String>>`.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -82,16 +83,19 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
...
-    var first = true
-    for (codeBlock in values) {
-      if (!first) codeWriter.emit(memberSeparator)
-      codeWriter.emitCode(codeBlock)
-      first = false
+    codeWriter.emitCode(values[0])
+    codeWriter.emit(whitespace)
+    if (values.size > 1) {
+      codeWriter.indent(2)
+      var first = true
```
JakeWharton: This isn't needed. You can use `i > 1` in the conditional.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -171,11 +208,51 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
...
+    private fun Any.toArrayTypeName() = when (this) {
+      is Boolean -> BooleanArray::class.asTypeName()
+      is Byte -> ByteArray::class.asTypeName()
+      is Char -> CharArray::class.asTypeName()
+      is Double -> DoubleArray::class.asTypeName()
+      is Float -> FloatArray::class.asTypeName()
+      is Int -> IntArray::class.asTypeName()
+      is Long -> LongArray::class.asTypeName()
+      is Short -> ShortArray::class.asTypeName()
+      else -> kotlin.Array<Any>::class.asTypeName()
```
JakeWharton: This won't work. You'll get a raw `Array` type not a generic. This has to be a `ParameterizedTypeName(ARRAY, ANY)`.

src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -298,28 +317,89 @@ class AnnotationSpecTest {
...
         |        Float::class,
         |        Double::class
-        |    }
+        |    )
+        |)
+        |class Taco
+        |""".trimMargin())
+  }
+
+  @Test fun arrayParams() {
```
JakeWharton: nice test

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
Egorand: Right. We don't really need the generic type to generate code though, so passing `Array<Any>::class` should always be fine.

Egorand: Since we never have to specify the generic type, we only distinguish between primitive type arrays and all other arrays. I was thinking about better API to avoid passing unnecessary params, is this a good idea?

JakeWharton:
This is no longer needed as of 1.2 so I propose we leave the behavior as-is.

JakeWharton:
Or, well, looks like they use `[` and `]` so that'll need changed.

PR #143, opened 2017-06-27
Recognize imports when emitting nullable types

### LINKED ISSUE #140: Type is written as fully qualified when made nullable
The following code:
```kotlin
val stringType = ClassName.run { String::class.asClassName() }
KotlinFile.builder("somePackage", "SomeFile.kt")
        .addProperty(PropertySpec.builder("prop1", stringType).build())
        .addProperty(PropertySpec.builder("prop2", stringType.asNullable()).build())
        .build()
        .toString()
```
generates:
```kotlin
package somePackage

import kotlin.String

val prop1: String

val prop2: kotlin.String?
```

The type for prop2 became `kotlin.String?` rather than `String?`

Egorand:
Fixes #140

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -70,7 +70,7 @@ class ClassName private constructor(
    * Returns the top class in this nesting group. Equivalent to chained calls to
    * [ClassName.enclosingClassName] until the result's enclosing class is null.
    */
-  fun topLevelClassName() = ClassName(names.subList(0, 2), names.size == 2 && nullable)
+  fun topLevelClassName() = ClassName(names.subList(0, 2))
 
   fun reflectionName(): String {
     // trivial case: no nested names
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -109,7 +109,7 @@ class ClassName private constructor(
   }
 
   /** Returns the simple name of this class, like `"Entry"` for [Map.Entry].  */
-  fun simpleName() = names[names.size - 1]
+  fun simpleName() = names[names.size - 1].let { if (nullable) "$it?" else it }
```
JakeWharton: This isn't correct. We need this method to return the actual class name, nullable or not.

src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -70,7 +70,7 @@ class ClassName private constructor(
    * Returns the top class in this nesting group. Equivalent to chained calls to
    * [ClassName.enclosingClassName] until the result's enclosing class is null.
    */
-  fun topLevelClassName() = ClassName(names.subList(0, 2))
+  fun topLevelClassName() = ClassName(names.subList(0, 2), names.size == 2 && nullable)
```
JakeWharton: I don't think nullability should be propagated here.

Egorand:
Thanks for the feedback @JakeWharton, indeed the problem seems to be in how `ClassName` gets looked up in the list of imports. Since an import covers both nullable and non-nullable types, we have to ignore nullability during look up. Please check last commit.

Issue #147, opened 2017-06-28
Nested annotations require parenthesis even if empty


JakeWharton:
```kotlin
@Foo(Bar())
// not
@Foo(Bar)
```

Egorand:
#151 addresses this as well, can be closed

JakeWharton:
Thanks!

PR #151, opened 2017-06-28
Fix annotation syntax when used as params

### LINKED ISSUE #146: Nested annotations shouldn't emit @
```kotlin
@Foo(things = [Bar(), Baz()])
// not
@Foo(things = [@Bar(), @Baz()])
```

Egorand:
Fixes #146 and #147

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -317,7 +317,7 @@ internal class CodeWriter @JvmOverloads constructor(
     if (o is TypeSpec) {
       o.emit(this, null)
     } else if (o is AnnotationSpec) {
-      o.emit(this, true)
+      o.emit(this, inline = true, asParameter = true)
     } else if (o is CodeBlock) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -317,7 +317,7 @@ internal class CodeWriter @JvmOverloads constructor(
     if (o is TypeSpec) {
       o.emit(this, null)
     } else if (o is AnnotationSpec) {
-      o.emit(this, true)
+      o.emit(this, inline = true, asParameter = true)
```
JakeWharton: Hmm I'm a bit unsure about this behavior as it could affect code in places we don't expect. As we don't have any tests which currently exercise this behavior I'm going to opt for the fix instead of being overly defensive. We may need to revisit this should a use-case come up.

Egorand: Agreed, it's too generic here, however I didn't find a better option and was just relying on tests.

PR #152, opened 2017-06-29
Convert boxed primtive types to kotlin types

### LINKED ISSUE #139: Incorrect generation for boxed primitives when used as generics
If you read the type `SomeClass<Integer>` and write it you get `SomeClass<Integer>` rather than `SomeClass<Int>`. The generated code compiles but complains that `Integer` should not be used.

ansman:
This fixes #139

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -121,6 +121,16 @@ class ClassName private constructor(
   companion object {
     @JvmStatic @JvmName("get")
     fun Class<*>.asClassName(): ClassName {
+      when (this) {
+        java.lang.Boolean::class.java -> return BOOLEAN
+        java.lang.Byte::class.java -> return BYTE
+        java.lang.Character::class.java -> return CHAR
+        java.lang.Short::class.java -> return SHORT
+        java.lang.Integer::class.java -> return INT
+        java.lang.Long::class.java -> return LONG
+        java.lang.Float::class.java -> return FLOAT
+        java.lang.Double::class.java -> return DOUBLE
+      }
       require(!isPrimitive) { "primitive types cannot be represented as a ClassName" }
       require(Void.TYPE != this) { "'void' type cannot be represented as a ClassName" }
       require(!isArray) { "array types cannot be represented as a ClassName" }
@@ -203,6 +213,18 @@ class ClassName private constructor(
       }
       names += getPackage(this).qualifiedName.toString()
       names.reverse()
+      if (names.size == 2 && names[0] == "java.lang") {
+        when (names[1]) {
+          "Boolean" -> return BOOLEAN
+          "Byte" -> return BYTE
+          "Character" -> return CHAR
+          "Short" -> return SHORT
+          "Integer" -> return INT
+          "Long" -> return LONG
+          "Float" -> return FLOAT
+          "Double" -> return DOUBLE
+        }
+      }
       return ClassName(names)
     }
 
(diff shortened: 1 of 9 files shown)


ansman:
One outstanding question, should `java.lang.String` be converted to `kotlin.String`?

ansman:
Also, I'm not sure about the location of these tests.

src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -203,6 +213,18 @@ class ClassName private constructor(
       }
       names += getPackage(this).qualifiedName.toString()
       names.reverse()
+      if (names.size == 2 && names[0] == "java.lang") {
+        when (names[1]) {
```
swankjesse: This seems like the wrong place to do this work. I think a more conventional mechanism would be for ClassName to have a method like `toKotlinClassName()` ?

swankjesse: ... also that’d make it possible to support other types, like MutableList?

ansman: Yeah, it felt a bit misplaced and I like an instance method.
That would make it less automatic which is fine as long as it's clearly documented I guess.

I wondered about things like MutableList but it's ambiguous if List should be MutableList or List, perhaps there is a good way to solve that though.

ansman: Oh, I see what you mean. Like you could have `toMutable` or something like that perhaps.

ansman:
@swankjesse I added it as a function on `TypeName`. There are a couple of outstanding questions, I put some in the code but the other one is regarding the collection types.
Perhaps a `asMutable`/`asImmutable` function could work? It would just be a bit confusing to call that on `Int` for example.

src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -48,6 +48,23 @@ class ClassName private constructor(
 
   override fun asNonNullable() = ClassName(names, false, annotations)
 
+  override fun toKotlinType(): ClassName = when (packageName()) {
+    "java.lang" -> when (simpleName()) {
```
ansman: Are there more types that should be handled here?

JakeWharton: There's probably hundreds to the point where we're not going to be able to maintain a list.

swankjesse:
Unclear what action to take on this. Closing for now.

ansman:
@swankjesse Any reason for not merging it?

src/test/java/com/squareup/kotlinpoet/ClassNameTest.kt
```diff
@@ -22,6 +22,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.fail
 import org.junit.Rule
 import org.junit.Test
+import javax.lang.model.type.TypeMirror
```
Egorand: This import looks unused

ansman: Fixed

uluckas:
Is there any good reason not to fix the boxed types thing, while thinking about how to handle other kotlin types?

Here is one more usecase:
I try to generate an override for an existing method
```kotlin
    open fun getI(i: CompletableDeferred<Int>)
```

And due to this issue, my generated code will be:
```kotlin
    override fun getI(i: CompletableDeferred<Integer>)
```

which won't compile.

JakeWharton:
That's a different issue. We should be reading the Kotlin type like the
compiler would.

Issue #150, opened 2017-06-28
Annotation use site target support


JakeWharton:
```kotlin
@field:Foo
```

JakeWharton:
Probably just a nullable enum on AnnotationSpec?

Egorand:
@JakeWharton should work. Should the following be supported?

```kotlin
class Example {
  @set:[Inject VisibleForTesting]
  var collaborator: Collaborator
}
```

https://kotlinlang.org/docs/reference/annotations.html#annotation-use-site-targets

JakeWharton:
That's an output optimization at best. Not something we express in the API.

We can get away with `@set:Inject @set:VisibleForTesting` for now.

PR #155, opened 2017-06-29
Support annotation use site targets

Egorand:
Implements #150.

#153 provides cleaner emission logic, once it's merged I'll update the PR.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -20,7 +20,6 @@ import java.io.IOException
 import java.io.StringWriter
 import java.lang.reflect.Array
 import java.util.Arrays
-import java.util.Locale
 import java.util.Objects
 import javax.lang.model.element.AnnotationMirror
 import javax.lang.model.element.AnnotationValue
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -20,6 +20,7 @@ import java.io.IOException
 import java.io.StringWriter
 import java.lang.reflect.Array
 import java.util.Arrays
+import java.util.Locale
```
JakeWharton: I don't want to add any new dependencies on the JVM. An easy way around this is to add the raw output string to each enum value.

Egorand: Makes sense. `KModifier` uses the same approach to printing, worth opening an issue to fix that as well?

JakeWharton: Sure. Or just do a PR if you have time.

I think long-term we'll have a core KotlinPoet and then platform-specific helpers for things like java.lang.Class, TypeMirror, TypeElement, Filer, etc. in a kotlinpoet-jvm artifact.

Egorand: Pushed the fix

JakeWharton:
I'll rebase on you

Egorand:
Great, thanks!

PR #159, opened 2017-06-30
Convert NameAllocatorTest to Kotlin.

JakeWharton:


src/test/java/com/squareup/kotlinpoet/NameAllocatorTest.kt
```diff
@@ -0,0 +1,107 @@
...
+    val nameAllocator = NameAllocator()
+    assertThat(nameAllocator.newName("1ab", 1)).isEqualTo("_1ab")
+  }
+
+  @Test fun characterMappingInvalidStartIsInvalidPart() {
+    val nameAllocator = NameAllocator()
+    assertThat(nameAllocator.newName("&ab", 1)).isEqualTo("_ab")
+  }
+
+  @Test fun javaKeyword() {
```
swankjesse: will wanna be kotlinKeyword later

PR #168, opened 2017-06-30
Support importing from top-level file.

### LINKED ISSUE #160: Support adding imports for top-level properties and functions.
`KotlinFile.Builder.addStaticImport` needs a way to import top-level properties.

NightlyNexus:
Closes #160.
Totally unsure if this is remotely the right thing to do...


JakeWharton:
Looks fine. When you do this you're taking fate into your own hands with regard to disambiguation, but naming conventions should mostly save you here.

Issue #154, opened 2017-06-29
Add ability to annotate a file


Egorand:
Add support for 

```kotlin
@file:JvmName("TacoUtils")

package com.squareup.tacos
```

Egorand:
I can work on a PR during the weekend.

JakeWharton:
Sounds good. We'll look to do a 0.4 next week with all these changes so
this would be nice to have.

JakeWharton:
f2297ef330a538e971168d53a436818f13a0b9b5

PR #169, opened 2017-06-30
Support file annotations

Egorand:
#154

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
+++ b/src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
@@ -50,6 +50,7 @@ private val NULL_APPENDABLE = object : Appendable {
  * aliases.
  */
 class KotlinFile private constructor(builder: KotlinFile.Builder) {
+  val fileAnnotations = builder.fileAnnotations.toImmutableList()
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/KotlinFile.kt
```diff
@@ -194,6 +202,24 @@ class KotlinFile private constructor(builder: KotlinFile.Builder) {
       require(SourceVersion.isName(fileName)) { "not a valid file name: $fileName" }
     }
 
+    fun addFileAnnotation(annotationSpec: AnnotationSpec) = apply {
+      fileAnnotations += annotationSpec.ensureFileUseSiteTarget()
+    }
+
+    private fun AnnotationSpec.ensureFileUseSiteTarget() = when (useSiteTarget) {
```
JakeWharton: This is only used in one place so I think we can just inline it into the `addFileAnnotation(AnnotationSpec)` method.

Also I would throw an exception if any other use-site target is already specified (with a test for that).

JakeWharton:
Fixed and merged manually as f2297ef330a538e971168d53a436818f13a0b9b5

Egorand:
Cool, thanks for finalizing!

PR #170, opened 2017-06-30
Replace Java keyword checks with Kotlin.

### LINKED ISSUE #164: Check for Kotlin keywords.
FunSpec.builder("void") fails. FunSpec.builder("fun") will generate.
I don't know of a Kotlin version of `SourceVersion.isName`.

Since it'll generate code that won't compile, I don't know if this check adds much value, anyway.

NightlyNexus:
Closes #164, as far as I can tell.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -114,3 +114,55 @@ internal fun stringLiteralWithQuotes(value: String): String {
     return result.toString()
   }
 }
+
+internal fun isIdentifier(name: String): Boolean {
+  return IDENTIFIER_REGEX.matches(name)
+}
+
+internal fun isKeyword(name: String): Boolean {
+  return KEYWORDS.contains(name)
+}
+
+internal fun isName(name: String): Boolean {
+  return name.split("\\.").none { isKeyword(name) }
+}
+
+private val IDENTIFIER_REGEX
+    = ("((\\p{gc=Lu}+|\\p{gc=Ll}+|\\p{gc=Lt}+|\\p{gc=Lm}+|\\p{gc=Lo}+|\\p{gc=Nl}+)+" +
+    "\\d*" +
+    "\\p{gc=Lu}*\\p{gc=Ll}*\\p{gc=Lt}*\\p{gc=Lm}*\\p{gc=Lo}*\\p{gc=Nl}*)" +
+    "|" +
+    "(`[^\n\r`]+`)")
+    .toRegex()
+
+// https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/renderer/KeywordStringsGenerated.java
+private val KEYWORDS = setOf(
+    "package",
+    "as",
+    "typealias",
+    "class",
+    "this",
+    "super",
+    "val",
+    "var",
+    "fun",
+    "for",
+    "null",
+    "true",
+    "false",
+    "is",
+    "in",
+    "throw",
+    "return",
+    "break",
+    "continue",
+    "object",
+    "if",
+    "try",
+    "else",
+    "while",
+    "do",
+    "when",
+    "interface",
+    "typeof"
+)
--- a/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -16,7 +16,6 @@
 package com.squareup.kotlinpoet
 
 import java.io.IOException
-import javax.lang.model.SourceVersion
 import javax.lang.model.element.Element
 import javax.lang.model.element.ElementKind
 import javax.lang.model.element.NestingKind.MEMBER
@@ -40,7 +39,7 @@ class ClassName private constructor(
 
   init {
     for (i in 1 until names.size) {
-      require(SourceVersion.isName(names[i])) { "part ${names[i]} is keyword" }
+      require(isName(names[i])) { "part ${names[i]} is keyword" }
     }
   }
 
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -17,7 +17,6 @@ package com.squareup.kotlinpoet
 
 import java.io.IOException
 import java.util.EnumSet
-import javax.lang.model.SourceVersion
 
 /** Sentinel value that indicates that no user-provided package has been set.  */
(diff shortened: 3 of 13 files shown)


src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -113,3 +114,47 @@ internal fun stringLiteralWithQuotes(value: String): String {
...
   }
 }
+
+internal fun isKeyword(name: String): Boolean {
+  return KEYWORDS.contains(name)
+}
+
+internal fun isName(name: String): Boolean {
+  return NAME_REGEX.split(name, 0).toTypedArray()
+      .none { !isIdentifier(it) || isKeyword(name) }
```
JakeWharton: I want to drop the Java check altogether.

NightlyNexus: Ah, right.
i'll give look at it and ask more here.

NightlyNexus:
am i going down the right path here for replacing isIdentifier?
do i need to find the list of illegal characters and hardcode them in here?
```kotlin
internal fun isIdentifier(name: String): Boolean {
  val first = name.codePointAt(0)
  if (first !in '\u001f'.toInt()..'\u007f'.toInt() || first == '.'.toInt() // TODO: illegal characters
      || first in '0'.toInt()..'9'.toInt()) {
    return false
  }
  var i = if (first >= 0x010000) 2 else 1
  while (i < name.length) {
    val part = name.codePointAt(i)
    if (first !in '\u001f'.toInt()..'\u007f'.toInt() || first == '.'.toInt()) {
      return false
    }
    i += if (part >= 0x010000) 2 else 1
  }
  return true
}
```

JakeWharton:
Honestly I was just thinking we guard against all the keywords for now and that'd be it. I don't know enough about Kotlin's restrictions on naming, but that looks alright.

NightlyNexus:
Oh, gotcha. and leave [this](https://github.com/square/kotlinpoet/blob/9b44af2ba31b4fbda76cd23d68fda13e4b53be5d/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt#L28) for now?

I also don't know the restrictions.
But, a quick look at https://jetbrains.github.io/kotlin-spec/ makes it look easy enough.
I'm going to try squeezing in the replacement into this change in a bit.

JakeWharton:
Ideally that one also changes to be Kotlin-specific eventually.

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -113,3 +113,60 @@ internal fun stringLiteralWithQuotes(value: String): String {
     return result.toString()
   }
 }
+
+internal fun isIdentifier(name: String): Boolean {
+  return IDENTIFIER_NAME_REGEX.matches(name) || IDENTIFIER_NAME_WITH_BACKTICKS_REGEX.matches(name)
```
NightlyNexus: oh, actually could merge the regexes.

NightlyNexus:
added some more tests. sorry for that notification spam.

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -113,3 +113,58 @@ internal fun stringLiteralWithQuotes(value: String): String {
...
+    "object",
+    "if",
+    "try",
+    "else",
+    "while",
+    "do",
+    "when",
+    "interface",
+    "typeof"
+)
```
JakeWharton: What about `public` and the like?

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -113,3 +113,58 @@ internal fun stringLiteralWithQuotes(value: String): String {
...
+internal fun isIdentifier(name: String): Boolean {
+  return IDENTIFIER_REGEX.matches(name)
+}
+
+internal fun isKeyword(name: String): Boolean {
+  return KEYWORDS.contains(name)
+}
+
+internal fun isName(name: String): Boolean {
+  return NAME_REGEX.split(name, 0).toTypedArray().none { isKeyword(name) }
```
JakeWharton: There's an overload of `split` that takes a `CharSequence` instead of a `Regex`. Can we just use that with `"."`?

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -113,3 +113,58 @@ internal fun stringLiteralWithQuotes(value: String): String {
...
+}
+
+private val NAME_REGEX = "\\.".toRegex()
+private val IDENTIFIER_REGEX
+    = ("((\\p{gc=Lu}+|\\p{gc=Ll}+|\\p{gc=Lt}+|\\p{gc=Lm}+|\\p{gc=Lo}+|\\p{gc=Nl}+)+" +
+    "\\d*" +
+    "\\p{gc=Lu}*\\p{gc=Ll}*\\p{gc=Lt}*\\p{gc=Lm}*\\p{gc=Lo}*\\p{gc=Nl}*)" +
```
JakeWharton: This is intense! Is there an explanation about where it came from? Can we accomplish this with imperative code that would be more readable (and maintainable)?

src/main/java/com/squareup/kotlinpoet/Util.kt
NightlyNexus: `public` is allowed. like `fun public() {}`

src/main/java/com/squareup/kotlinpoet/Util.kt
NightlyNexus: yep!

src/main/java/com/squareup/kotlinpoet/Util.kt
NightlyNexus: https://jetbrains.github.io/kotlin-spec/#_lexical_grammar
identifier-start (line 131) and identifier-part (132 and 133).

Yeah, it's too crazy. I'll look at it again and see if I can think of anything.

src/main/java/com/squareup/kotlinpoet/Util.kt
JakeWharton: Huh. Cool.

src/main/java/com/squareup/kotlinpoet/Util.kt
NightlyNexus: I'm not sure what to do here.
We could duplicate Character.isLetter as a helper and use a loop.
Any ideas?
I could also revert this and only replace `isKeyword` and `isName` until we think of something sane.

swankjesse:
Thanks!

PR #171, opened 2017-06-30
Use StringBuilder over Java-specific StringWriter.

JakeWharton:


JakeWharton:
TBR. Merging as I don't want to conflict myself locally since I'm about to touch all these files. You already reviewed the JavaPoet version of this so I'm counting that 😀 .

PR #174, opened 2017-07-01
Random code fixes and improvements

Egorand:
- Mostly small code style issues fixed by auto-formatting
- Replaced a number of statements to use more Kotlin features, where it makes sense

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -55,12 +55,8 @@ class FunSpec private constructor(builder: Builder) {
     }
   }
 
-  internal fun parameter(name: String): ParameterSpec? {
-    for (parameter in parameters) {
-      if (parameter.name == name) return parameter
-    }
-    return null
-  }
+  internal fun parameter(name: String): ParameterSpec? =
+      parameters.firstOrNull { it.name == name }
 
   @Throws(IOException::class)
   internal fun emit(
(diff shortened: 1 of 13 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -55,12 +55,8 @@ class FunSpec private constructor(builder: Builder) {
     }
   }
 
-  internal fun parameter(name: String): ParameterSpec? {
-    for (parameter in parameters) {
-      if (parameter.name == name) return parameter
-    }
-    return null
-  }
+  internal fun parameter(name: String): ParameterSpec? =
```
JakeWharton: Return type can be omitted here

Egorand: Thought about keeping it to make it more clear that it's nullable, but it's probably clear anyway

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -56,14 +54,9 @@ class FunSpecTest {
...
-    for (executableElement in elements) {
-      if (executableElement.simpleName.toString() == name) {
-        return executableElement
-      }
-    }
-    throw IllegalArgumentException(name + " not found in " + elements)
-  }
+  private fun findFirst(elements: Collection<ExecutableElement>, name: String) =
+      elements.firstOrNull { it.simpleName.toString() == name } ?:
+          throw IllegalArgumentException(name + " not found in " + elements)
```
NightlyNexus: could also be `requireNotNull(elements.firstOrNull { it.simpleName.toString() == name }) { name + " not found in " + elements }`

Egorand: Yep, good one! I guess I will submit another small PR

PR #175, opened 2017-07-01
Random code fixes #2

Egorand:
Follow up to #174

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -55,9 +55,8 @@ class FunSpecTest {
   }
 
   private fun findFirst(elements: Collection<ExecutableElement>, name: String) =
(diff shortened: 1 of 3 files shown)


src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -55,8 +55,9 @@ class FunSpecTest {
   }
 
   private fun findFirst(elements: Collection<ExecutableElement>, name: String) =
-      elements.firstOrNull { it.simpleName.toString() == name } ?:
-          throw IllegalArgumentException(name + " not found in " + elements)
+      requireNotNull(elements.firstOrNull { it.simpleName.toString() == name }) {
+        "$name not found in $elements"
+      }
```
JakeWharton: I don't think this one is better. Can you revert this one?

Egorand: Sure

Issue #157, opened 2017-06-29
Split platform-specific helper into their own artifacts.


JakeWharton:
Thinking `kotlinpoet` and `kotlinpoet-jvm`, the latter of which has all the `java.*` things like `Class` things, `TypeElement` things, `TypeMirror` things, and #105.

JakeWharton:
This would let someone use the library for Kotlin generation from a native tool or JS tool on node.js, should they want to do so.

JakeWharton:
I think a good start is at least getting them into separate files. Have 90% of this locally.

Egorand:
There's a lot of `@JvmField` and `@JvmStatic` usages to help provide more Java-friendly API. Would be good to remove them from the main library, and supply Java-friendly API with `kotlinpoet-jvm`.

Egorand:
@JakeWharton have something to push?

Egorand:
Closing this in favor of #304

PR #179, opened 2017-07-04
Move extensions to top-level functions.

JakeWharton:
IntelliJ IDEA cannot currently discover them when embedded inside an companion object. Plus this makes better imports in the consuming code.

Refs #157.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -13,9 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:JvmName("TypeNames")
+
 package com.squareup.kotlinpoet
 
-import com.squareup.kotlinpoet.ClassName.Companion.asClassName
 import java.io.IOException
 import java.lang.reflect.GenericArrayType
 import java.lang.reflect.ParameterizedType
@@ -51,7 +52,7 @@ import kotlin.reflect.KClass
  * --------------------------
  *
  * In an annotation processor you can get a type name instance for a type mirror by calling
- * [TypeName.get]. In reflection code, you can use [TypeName.get].
+ * [asTypeName]. In reflection code, you can use [asTypeName].
 
  * Defining new types
  * ------------------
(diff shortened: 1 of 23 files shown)


src/main/java/com/squareup/kotlinpoet/WildcardTypeName.kt
```diff
@@ -121,3 +118,9 @@ class WildcardTypeName private constructor(
     }
   }
 }
+
+@JvmName("get")
+fun javax.lang.model.type.WildcardType.asWildcardTypeName() = Companion.get(this, mutableMapOf())
```
Egorand: Why not `WildcardTypeName.get(this, mutableMapOf())`?

JakeWharton: Probably just the result of automated refactoring. Will fix.

src/main/java/com/squareup/kotlinpoet/NameAllocator.kt
```diff
@@ -84,7 +84,7 @@ class NameAllocator private constructor(
   /**
    * Return a new name using `suggestion` that will not be a Java identifier or clash with other
    * names. The returned value can be queried multiple times by passing `tag` to
-   * [NameAllocator.get].
+   * [NameAllocator.fromType].
```
NightlyNexus: it's still `get`. accidental change?

src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -51,7 +52,7 @@ import kotlin.reflect.KClass
  * --------------------------
  *
  * In an annotation processor you can get a type name instance for a type mirror by calling
- * [TypeName.get]. In reflection code, you can use [TypeName.get].
+ * [TypeName.fromTypeMirror]. In reflection code, you can use [TypeName.fromType].
```
NightlyNexus: the static method names are still all `get`.

src/main/java/com/squareup/kotlinpoet/NameAllocator.kt
JakeWharton: Just IntelliJ being stupid when it's trying to be really smart.

src/main/java/com/squareup/kotlinpoet/WildcardTypeName.kt
Egorand: Cool, that `Companion` import looked weird

src/main/java/com/squareup/kotlinpoet/TypeName.kt
JakeWharton: They're `get` in Java but `asTypeName` in Kotlin

PR #181, opened 2017-07-09
Improved handling of single expression functions

### LINKED ISSUE #180: Support expression bodies + control flow
I was trying to do this:

```kotlin
FunSpec.builder("toppingPrice")
  .addParameter("topping", String::class)
  .beginControlFlow("return when (topping)")
  .addStatement("%S -> 1", "beef")
  .addStatement("%S -> 2", "lettuce")
  .addStatement("%S -> 3", "cheese")
  .addStatement("else -> throw IllegalToppingException(topping)")
  .endControlFlow()
  .build()
```

expecting this:

```kotlin
fun toppingPrice(topping: String) = when (topping) {
  "beef" -> 1
  "lettuce" -> 2
  "cheese" -> 3
  else -> throw IllegalToppingException(topping)
}
```

but got this:

```kotlin
fun toppingPrice(topping: String) {
  return when (topping) {
    "beef" -> 1
    "lettuce" -> 2
    "cheese" -> 3
    else -> throw IllegalToppingException(topping)
  }
}
```

Is this a valid use case for `beginControlFlow()`?

Egorand:
Fixes #180 and extends support for single expression functions. 

Still WIP, there's a test with broken formatting which I'll try to fix. Would love to get some feedback on the approach.


Egorand:
So the problem was in emitting the following incorrectly indented code:

```kotlin
|package com.squareup.tacos
|
|class Taco {
|  val NAME: Thing.Thang<Foo, Bar> = object : Thing.Thang<Foo, Bar>() {
|    override fun call(final thung: Thung<in Foo>): Thung<in Bar> =
|        object : SimpleThung<Bar>(thung) {
|      override fun doSomething(bar: Bar) {
|        /* code snippets */
|      }
|    }
|  }
|}
```

Fixed by removing `%W` before the expression body. This is not ideal, as long lines won't get wrapped, but I'm looking at how `PropertySpec` emits its initializer:

```kotlin
if (withInitializer && initializer != null) {
    ...
    codeWriter.emit(" = ")
  }
  codeWriter.emitCode("%[%L%]", initializer)
}
```

which, if changed to `emitCode(" =%W")`, would lead to similar indentation problems. Is this worth opening a separate issue for?

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -138,6 +138,10 @@ class CodeBlock private constructor(
     internal val formatParts = mutableListOf<String>()
     internal val args = mutableListOf<Any?>()
 
+    fun isEmpty() = formatParts.isEmpty()
+
+    fun isNotEmpty() = !isEmpty()
```
JakeWharton: :+1: to these two methods. I'm pretty sure there's a few places in the codebase where they can be used.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -78,13 +79,8 @@ class FunSpec private constructor(builder: Builder) {
       return
     }
 
-    val asExpressionBody = code.withoutPrefix(EXPRESSION_BODY_PREFIX)
```
JakeWharton: Can we remove this method from `CodeBlock` then?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -41,6 +41,7 @@ class FunSpec private constructor(builder: Builder) {
   val parameters = builder.parameters.toImmutableList()
   val exceptions = builder.exceptions.toImmutableList()
   val code = builder.code.build()
+  val isSingleExpression = builder.isSingleExpression
```
JakeWharton: I don't think "single" adds any value here. Kotlin calls these "expression bodies" so I think either `isExpression` or `isExpressionBody` is sufficient. I would also think we rename `code` to `body` here...

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
JakeWharton: Ah I see it used below..

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -342,6 +359,8 @@ class FunSpec private constructor(builder: Builder) {
     internal const val CONSTRUCTOR = "constructor()"
     internal const val GETTER = "get()"
     internal const val SETTER = "set()"
+    internal const val RETURN = "return "
+    @JvmField internal val RETURN_CODE_BLOCK = CodeBlock.of(RETURN)
```
JakeWharton: Why the annotation?

Egorand: I would usually use it in this context to make compiler generate a simple `public static final` field, but in this case it probably just adds clutter. I'll remove it.

JakeWharton: For Android apps and libraries such a micro-optimization is probably useful, but for libraries we want to be as generic and idiomatic as possible. Plus we want to reduce our dependence on JVM-specific bits to hopefully allow this library to work on node.js or via native code someday.

Egorand: Makes sense. There's a number of places where this technique is used on `internal` members, I guess some cleanup would be good.

JakeWharton: Yeah there's a lot of artifacts from since this codebase started from porting JavaPoet.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -296,43 +294,62 @@ class FunSpec private constructor(builder: Builder) {
         = addParameter(name, type.asTypeName(), *modifiers)
 
     fun addCode(format: String, vararg args: Any) = apply {
-      code.add(format, *args)
+      body.add(stripReturn(format), *args)
+    }
+
+    private fun stripReturn(format: String) = when {
```
JakeWharton: I'm a little concerned about this method because of it's side-effect in setting the builder boolean. Can we call it something like `expressionBodyHandler` instead?

Egorand:
@swankjesse I like the idea, with that we should be able to do `body.trim().withoutPrefix("return ")`, which would solve the problem of matching `%[` in statements. I also agree that there's value in keeping the model unchanged. Let me experiment with it when I have more time, and I'll push the results.

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -108,6 +108,25 @@ class CodeBlock private constructor(
     return CodeBlock(resultFormatParts, resultArgs)
   }
 
+  /**
+   * Returns a copy of the code block without leading and trailing no-arg placeholders
+   * (`%W`, `%<`, `%>`, `%[`, `%]`).
+   */
+  internal fun trim(): CodeBlock {
```
swankjesse: Should we require `%>`  and `%[` be balanced?

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -374,6 +397,7 @@ class CodeBlock private constructor(
     @JvmField internal val LOWERCASE = Regex("[a-z]+[\\w_]*")
     internal const val ARG_NAME = 1
     internal const val TYPE_NAME = 2
+    internal val NO_ARG_PLACEHOLDERS = setOf("%W", "%<", "%>", "%[", "%]")
```
swankjesse: nit: reorder `%<` and `%>` ? Typically we indent with `>` then unindent with `<` not the opposite!

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -236,7 +233,7 @@ class FunSpec private constructor(builder: Builder) {
           Modifier.ABSTRACT -> this.modifiers += KModifier.ABSTRACT
           Modifier.FINAL -> this.modifiers += KModifier.FINAL
           Modifier.NATIVE -> this.modifiers += KModifier.EXTERNAL
-          Modifier.DEFAULT -> {}
+          Modifier.DEFAULT -> Unit
```
swankjesse: Neat

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -443,4 +440,4 @@ internal val String.isConstructor: Boolean
 internal val String.isAccessor: Boolean
   get() = this == GETTER || this == SETTER
 
-internal val EXPRESSION_BODY_PREFIX = CodeBlock.of("%[return ")
+internal val EXPRESSION_BODY_PREFIX = CodeBlock.of("return ")
```
swankjesse: much nicer

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -404,6 +404,52 @@ class CodeBlockTest {
...
+        .isEqualTo(CodeBlock.of("return this"))
+  }
+
+  @Test fun trimMultipleNoArgPlaceholders() {
+    assertThat(
+        CodeBlock.of("%[return if (x > %L) %S %Welse %S%]", 1, "a", "b").trim())
+        .isEqualTo(CodeBlock.of("return if (x > %L) %S %Welse %S", 1, "a", "b"))
+  }
+
+  @Test fun trimOnlyNoArgPlaceholders() {
```
swankjesse: Nice.

swankjesse:
This worked out very nicely!

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Egorand: So far I haven't encountered a use case for that.

PR #184, opened 2017-07-16
Don't emit setter parameter type

### LINKED ISSUE #183: Don't emit setter parameter type
```kotlin
println(FunSpec.setterBuilder()
    .addParameter("value", String::class)
    .addStatement("parseTacoFromString(value)")
    .build())
```

will result in

```kotlin
set(value: kotlin.String) {
  parseTacoFromString(value)
}
```

where `kotlin.String` is unnecessary.

Egorand:
Fixes #183

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -31,10 +31,13 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
   val defaultValue = builder.defaultValue
 
   @Throws(IOException::class)
-  internal fun emit(codeWriter: CodeWriter) {
+  internal fun emit(codeWriter: CodeWriter, emitType: Boolean = true) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -31,10 +31,13 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
   val defaultValue = builder.defaultValue
 
   @Throws(IOException::class)
-  internal fun emit(codeWriter: CodeWriter) {
+  internal fun emit(codeWriter: CodeWriter, emitType: Boolean = true) {
```
JakeWharton: I think I would prefer this being called `includeType` as the "emit" prefix is semi-redundant with the name of the method.

Egorand: Yep, good point, will fix

swankjesse:
Thanks!

Issue #138, opened 2017-06-25
Update KDoc


Egorand:

Egorand:
KDoc is up-to-date with version 0.4.0, closing the issue

PR #185, opened 2017-07-17
Remove writeTo(Filer) from KDoc.

hotchemi:
already removed at https://github.com/square/kotlinpoet/pull/93


JakeWharton:
These are generated automatically. We're going to release 0.4 shortly and will update all the Javadoc (#138).

PR #189, opened 2017-08-07
Add declaration-site variance support

### LINKED ISSUE #188: Support declaration-site variance
There seems to be no way of generating the following:

```kotlin
class B<out E : A>
```

I suggest parameterizing `TypeSpec`'s `addTypeVariable()` to be able to specify variance.

Egorand:
Fixes #188

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
@@ -74,43 +74,24 @@ class TypeVariableName private constructor(
       return TypeVariableName(name, bounds.filter { it != ANY }, variance)
     }
 
-    /** Returns type variable named `name` without bounds.  */
-    @JvmStatic @JvmName("get")
-    operator fun invoke(name: String) = TypeVariableName.of(name, emptyList(), variance = null)
-
     /** Returns type variable named `name` with `variance` and without bounds.  */
-    @JvmStatic @JvmName("get")
-    operator fun invoke(name: String, variance: KModifier) =
+    @JvmStatic @JvmName("get") @JvmOverloads
+    operator fun invoke(name: String, variance: KModifier? = null) =
         TypeVariableName.of(name, emptyList(), variance)
 
-    /** Returns type variable named `name` with `bounds`.  */
-    @JvmStatic @JvmName("get")
-    operator fun invoke(name: String, vararg bounds: TypeName) =
-        TypeVariableName.of(name, bounds.toList(), variance = null)
-
     /** Returns type variable named `name` with `variance` and `bounds`.  */
-    @JvmStatic @JvmName("get")
-    operator fun invoke(name: String, variance: KModifier, vararg bounds: TypeName) =
+    @JvmStatic @JvmName("get") @JvmOverloads
+    operator fun invoke(name: String, vararg bounds: TypeName, variance: KModifier? = null) =
         TypeVariableName.of(name, bounds.toList(), variance)
 
-    /** Returns type variable named `name` with `bounds`.  */
-    @JvmStatic @JvmName("get")
-    operator fun invoke(name: String, vararg bounds: KClass<*>) =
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -65,32 +66,52 @@ class TypeVariableName private constructor(
   }
 
   companion object {
-    internal fun of(name: String, bounds: List<TypeName>): TypeVariableName {
+    internal fun of(name: String, bounds: List<TypeName>, variance: KModifier?): TypeVariableName {
+      require(variance == null || variance == KModifier.IN || variance == KModifier.OUT) {
+        "$variance is an invalid variance modifier, the only allowed values are in and out!"
```
swankjesse: neat

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -65,32 +66,52 @@ class TypeVariableName private constructor(
...
     @JvmStatic @JvmName("get")
-    operator fun invoke(name: String, vararg bounds: KClass<*>): TypeVariableName {
-      return TypeVariableName.of(name, bounds.map { it.asTypeName() })
-    }
+    operator fun invoke(name: String, vararg bounds: KClass<*>) =
+        TypeVariableName.of(name, bounds.map { it.asTypeName() }, variance = null)
+
+    /** Returns type variable named `name` with `variance` and `bounds`.  */
+    @JvmStatic @JvmName("get")
+    operator fun invoke(name: String, variance: KModifier, vararg bounds: KClass<*>) =
```
swankjesse: there’s no way to use default values to save a bunch of overloads? This seems excessive!

src/test/java/com/squareup/kotlinpoet/TypeVariableNameTest.kt
```diff
@@ -101,4 +102,30 @@ class TypeVariableNameTest {
...
+    assertThat(typeSpec.toString()).isEqualTo("""
+      |class Taco<out E : kotlin.Number>
+      |""".trimMargin())
+  }
+
+  @Test fun invalidVariance() {
+    try {
+      TypeVariableName("E", KModifier.FINAL)
+      fail()
+    } catch (e: IllegalArgumentException) {
```
swankjesse: nit: rename `expected` ?

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
Egorand: True. Since those methods target Java, it might be tricky to get proper overloads, but I'll give it a go.

swankjesse: Should figure out exactly how much we want these APIs to be consumable from Java. Since this is a brand new API I’m not certain it’s a requirement!

Egorand: I like the idea of #157, then we can move all Java-specific stuff outside of the core library.

Egorand: I tinkered with the factory methods a bit and pushed an alternative version. Now, to pass both `bounds` and `variance` in Kotlin, you have to use named arguments:

```kotlin
TypeVariableName("E", Number::class, variance = KModifier.IN)
```
In my opinion, that's not ideal, but since `variance` is an optional parameter, I think it's fine. Additionally, I generally like using named arguments, as it improves readability in most cases.

If you want to do the same in Java, you'll have to pass an array instead of varargs:

```java
TypeVariableName.get("name", new TypeName[]{null}, KModifier.OUT);
```
Given that we're probably not aiming at perfectly Java-friendly API, I think it's alright. What do you think?

swankjesse: I like the new code better.

PR #191, opened 2017-08-18
Fix failure on adding interface function as a default method, also allowed empty body with no return type

prestongarno:


prestongarno:
I use tabs, sorry about the reverts

prestongarno:
I use tabs, sorry about all the reverts

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -483,8 +484,12 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
     fun addFun(funSpec: FunSpec) = apply {
       if (kind == Kind.INTERFACE) {
-        requireExactlyOneOf(funSpec.modifiers, KModifier.ABSTRACT)
-        requireExactlyOneOf(funSpec.modifiers, KModifier.PUBLIC, KModifier.PRIVATE)
+        if(funSpec.body.isEmpty()) {
```
Egorand: Please add a unit test

prestongarno: will do

prestongarno:
Didn't know if this should be checked at the TypeSpec#addFun of at the end at `type.build()`.  There's checks in both places but they're inconsistent. `addFun` method doesn't apply when adding functions through iterator `addFunctions` & it's how I'm currently getting around the error

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -483,8 +483,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
     fun addFun(funSpec: FunSpec) = apply {
       if (kind == Kind.INTERFACE) {
-        requireExactlyOneOf(funSpec.modifiers, KModifier.ABSTRACT)
-        requireExactlyOneOf(funSpec.modifiers, KModifier.PUBLIC, KModifier.PRIVATE)
+
+        if(funSpec.body.isEmpty()) {
```
Egorand: nit: missing space after `if`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -483,8 +483,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          require(!funSpec.modifiers.contains(KModifier.PRIVATE), {"Abstract $kind function can not be private"})
```
Egorand: nit: it would be more consistent to have lambda outside the function parentheses:

```kotlin
require(!funSpec.modifiers.contains(KModifier.PRIVATE)) {
  "Abstract $kind function can not be private"
}
```

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -483,8 +483,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          requireNotNull(funSpec.returnType,
```
Egorand: It's actually fine to suppress the type, it will always default to `Unit`. When I add `Unit` explicitly, IDE warns me that it's redundant.

Also here, same thing about placing lambda outside the function parentheses.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -837,6 +839,46 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test
```
Egorand: nit: put `@Test` on the same line with `fun`

Egorand:
@prestongarno yeah, I see that function checks are applied inconsistently throughout the `Builder` methods, that's worth a separate PR.

PR #193, opened 2017-08-18
Replace kotlin-test with kotlin-test-junit

Egorand:
IDEA says "kotlin-test-junit is better with junit"


JakeWharton:
If this ever becomes a multiplatform project we can switch back

PR #194, opened 2017-08-18
Fix inconsistent FunSpec validation

Egorand:
Checks were only performed in `addFun()` and not in `addFunctions()`.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -478,7 +478,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addFunctions(funSpecs: Iterable<FunSpec>) = apply {
-      this.funSpecs += funSpecs
+      funSpecs.forEach { addFun(it) }
     }
 
     fun addFun(funSpec: FunSpec) = apply {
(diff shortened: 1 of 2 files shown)


Egorand:
@JakeWharton `addFuns()`?

JakeWharton:
I'm think we switch to just `addFunction`. We don't say add `val`, we say `addProperty`, so there's no need to use 'fun' here.

JakeWharton:
Can be separate PR.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -478,15 +478,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addFunctions(funSpecs: Iterable<FunSpec>) = apply {
-      this.funSpecs += funSpecs
+      funSpecs.forEach { addFun(it) }
```
JakeWharton: BTW this could be a method ref
```kotlin
funSpecs.forEach(this::addFun)
```

Egorand: Nope, since `forEach` expects `(FunSpec) -> Unit`, and `addFun` is `(FunSpec) -> Builder`

JakeWharton: Ah, there's a bug to fix that.

PR #199, opened 2017-08-22
Delegate superinterfaces as constructor parameters

### LINKED ISSUE #198: Support adding a superinterface for TypeSpec by constructor parameter delegation
I can't find anything about this in the docs, but I'm currently accomplishing this by string replace after building the types. Something like 

`TypeSpec::addSuperinterface(parameter: ParameterSpec)`

to generate a class with the `: <Type> by <constructor parameter>` signature

prestongarno:
resolves #198 - adding ability to declare superinterface implementation on a class by delegate to constructor parameter. I think overloading `addSuperinterface` with `ParameterSpec` and keeping track of them as a separate `TypeSpec.Builder` property guards against the most errors

Added multiple test cases, only thing which is pretty much impossible is checking all other constructors of a class. So far it just requires a non-null primary constructor on call to `build()` and adds the delegate `ParameterSpec` if not done explicitly.


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -134,7 +136,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         val types = listOf(superclass).filter { it != ANY }.map {
           CodeBlock.of("%T(%L)", it, superclassConstructorParameters.joinToCode())
         }
-        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) }
+        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) } + superinterfaceDelegates.map { CodeBlock.of("${it.type} by ${it.name}") }
```
Egorand: Type of the superinterface and type of the parameter aren't necessarily the same, take this case for example:

```kotlin
class A : Consumer<Byte> {
  override fun accept(t: Byte) {
  }
}

class B(val a : A): Consumer<Byte> by a
```

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -336,6 +338,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     internal var superclass: TypeName = ANY
     internal val superclassConstructorParameters = mutableListOf<CodeBlock>()
     internal val superinterfaces = mutableListOf<TypeName>()
+    internal val superinterfaceDelegates = mutableListOf<ParameterSpec>()
```
Egorand: All we need to emit the `by` clause is the name of the parameter, it feels redundant to me to store `ParameterSpec`s for this purpose

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -441,6 +444,14 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     fun addSuperinterface(superinterface: KClass<*>)
         = addSuperinterface(superinterface.asTypeName())
 
+    fun addSuperinterface(superinterface: ParameterSpec) = apply {
```
Egorand: Not sure I like the API, specifying a superinterface using a `ParameterSpec` feels confusing.. And, as I mentioned, the types won't necessarily match.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -519,6 +530,24 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         "anonymous type has too many supertypes"
       }
 
+      if (superinterfaceDelegates.isNotEmpty()) {
+        requireNotNull(primaryConstructor) {
```
Egorand: Nice check

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -519,6 +530,24 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          "delegate superinterfaces ${superinterfaceDelegates.map { it.type }} " +
+              "require value parameters ${superinterfaceDelegates.map { it.name }} by primary constructor" }
+
+        superinterfaceDelegates.forEach {
+          require(!superinterfaces.contains(it.type)) {
+            "class '$name' already implements ${it.type}, delegating to constructor parameter '${it.name}' not allowed" }
+        }
+
+        if(primaryConstructor?.parameters?.containsAll(superinterfaceDelegates) == false)
```
Egorand: nit: missing space after `if`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -519,6 +530,24 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+            primaryConstructor = primaryConstructor!!.toBuilder()
+                .addParameters(superinterfaceDelegates)
+                .build()
+        // re-assigning ^^^ allowed?? otherwise need to require non-null ctor at time of declaring delegate
```
Egorand: This validation can be made in `build()`, when we're sure all the args have been provided

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -18,10 +18,7 @@ package com.squareup.kotlinpoet
 import com.google.common.collect.ImmutableMap
 import com.google.common.truth.Truth.assertThat
 import com.google.testing.compile.CompilationRule
-import com.squareup.kotlinpoet.KModifier.ABSTRACT
-import com.squareup.kotlinpoet.KModifier.INTERNAL
-import com.squareup.kotlinpoet.KModifier.PRIVATE
-import com.squareup.kotlinpoet.KModifier.VARARG
+import com.squareup.kotlinpoet.KModifier.*
```
Egorand: Better keep separate imports

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2580,6 +2579,105 @@ class TypeSpecTest {
     }
   }
 
+  @Test fun delegateByCtorParameter() {
+    val delegate = ParameterSpec.builder( "cheese",
+        ParameterizedTypeName.get(
```
Egorand: `ParameterizedTypeName.get(Consumer::class, Byte::class)` works

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2580,6 +2579,105 @@ class TypeSpecTest {
...
+    try {
+      TypeSpec.classBuilder("Taco")
+          .addSuperinterface(delegate)
+          .build()
+      fail()
+    } catch (expected: IllegalArgumentException) {
+      assertThat(expected)
+          .hasMessage("delegate superinterfaces [kotlin.String] require value parameters [cheese] by primary constructor")
+    }
+  }
```
Egorand: nit: empty line after `}` missing

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: What if we had a `superinterfacesWithDelegates = mutableMapOf<TypeName, String>()` instead?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: I thought about that too, maybe just rename to something ilke `implementByDelegate` or `superinterfaceBy`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: There needs to be at least something which keeps track of them. What if someone reverts the typespec to a builder, iterates over all of the functions' parameters, and changes the name of each parameter? I can't think of any other way except direct reference at the moment

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Maybe a simple `addSuperinterface(superinterface: KClass<*>, delegateParameterName: String)` would do?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
prestongarno: Jetbrains being helpful. Should probably keep my settings in a repository now with all the different ones I use - tough to keep track

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: You can validate that in `build()`, check that for every delegate there indeed is a primary constructor parameter with that name

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: Good call. I was pretty sure that there'd be something more to this because it's dealing with types. Recurse up the type hierarchy is the only way I think.  There should be already something like this in kotlinpoet for the rest of the type names though? I think I saw it when writing this

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: delegate can be a `val` or simply a dummy parameter that's not doing anything.  I'm all for keeping it simple, but there's always going to be an edge case... Modifiers on the input argument too? It's a bit more complicated because it's affecting multiple aspects of the type

prestongarno: `addSuperinterface(superinterface: KClass<*>, delegateParameterName: String)` is much more straightforward, but there will need to be assertions made about what the parameter will be generated as (and consequently - restrictions)

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: True, this is code generation and not a commpiler, after all

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: But you'll still have to pass the actual `ParameterSpec` with the primary constructor definition. To print the `by` part, we only need to know the name of that `ParameterSpec`.

prestongarno: > you'll still have to pass the actual ParameterSpec with the primary constructor definition.

Ok, now I see where you're coming from.  I don't think I explained my thinking behind the way I wrote it.  

When I thought of `superinterfaceBy(param: ParamSpec)`,  I thought it carried the implication that it'd be generated as a param of the primary constructor - even though the function call is on the `TypeSpec` with no accompanying `FunSpec`.  Since there's no distinction between a primary constructor and functions (at the moment), my reasoning was that the only way to support a configurable property (which is also delegated to by a type) was to leave it up to the caller to know that there needed to be a primary ctor for that type - and the delegate parameter would be added whether explicitly added or not. That's why I checked for a non-null constructor on the call to `build()`, and then the ParamSpec is added to the ctor if not done already. 

I'm fine with changing it though. My use case (which led to the pull request) has a default value for the parameter, which is why I went at the problem from that angle.  Maybe there needs to be a more specific type like `ConstructorSpec` in order to make that intuitive, but it's not my call to change up the type hierarchy

Egorand: Yeah, I see your point. I don't feel good about implicitly modifying the primary constructor though, at the very least it's not intuitively clear that `superinterfaceBy()` would do this, and we'll have to resolve duplicate `ParamSpec`s coming from this method and the primary constructor's `FunSpec`. In my opinion, it's safer to leave primary constructor creation unchanged, and add a way to map the name of a specific primary constructor param to a superinterface we're adding. But I might be missing some edge cases here.

@JakeWharton @swankjesse what are your thoughts on the API?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
JakeWharton: github.com/square/java-code-styles

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: You won't be able to re-use `ParameterSpec` not only because of the aforementioned different type situation, but also because you can do things like
```kotlin
interface B
class Foo: B
class A : B by Foo()
```

Egorand: `addSuperinterface(superinterface: KClass<*>, delegate: CodeBlock)`?

JakeWharton: Seems reasonable. Might also want the `String` overload for the parameter name since it's the common case.

prestongarno: Sounds good.  I'll change it up.

And wow, today I learned that the `by ...` expression of a class declaration could be an arbitrary expression returning an interface instance - makes sense though.

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -397,6 +397,7 @@ class CodeBlock private constructor(
     internal val LOWERCASE = Regex("[a-z]+[\\w_]*")
     internal const val ARG_NAME = 1
     internal const val TYPE_NAME = 2
+    internal val empty = of("")
```
Egorand: Since it's a constant, the name should be all uppercase

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -34,6 +34,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val superclass = builder.superclass
   val superclassConstructorParameters = builder.superclassConstructorParameters.toImmutableList()
   val superinterfaces = builder.superinterfaces.toImmutableList()
+  val delegateSuperInterfaces = builder.delegateSuperInterfaces.toImmutableMap()
```
Egorand: Not sure I like the name. Since that map maps superinterfaces to their delegates, I think a better name would be `superinterfaceDelegates`, or `delegatedSuperinterfaces`.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,23 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       this.superinterfaces += superinterfaces
     }
 
-    fun addSuperinterface(superinterface: TypeName) = apply {
-      superinterfaces += superinterface
+    fun addSuperinterface(superinterface: TypeName, delegate: CodeBlock = CodeBlock.empty) = apply {
+      if (delegate === CodeBlock.empty)
```
Egorand: This check will fail if somebody passes a different `CodeBlock.of("")`. Would it be better to use `==`, or check via `delegate.isEmpty()`?

Egorand:
Looks good! As @JakeWharton mentioned, you might also want to add an overload with `delegate: String` as the second param, as the most frequent case is delegating to a constructor parameter.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: Checked with `===` and made the `CodeBlock.EMPTY` internal because I thought it'd be consistent with the way the current API handles `CodeBlock` with zero checks except the for the pre-defined symbols  (i.e. kotlinpoet doesn't stop you from shooting yourself in the foot). Although an empty codeblock is a pretty simple case

prestongarno: I changed it to delegate.isEmpty(), so it simply adds the regular superinterface on an empty CodeBlock (doesn't fail it), assuming that's what you meant

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,34 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       this.superinterfaces += superinterfaces
     }
 
-    fun addSuperinterface(superinterface: TypeName) = apply {
-      superinterfaces += superinterface
+    fun addSuperinterface(superinterface: TypeName, delegate: CodeBlock = CodeBlock.EMPTY) = apply {
+      if (delegate.isEmpty())
```
JakeWharton: Curly braces are _always_ required for blocks unless the statement is on the same line as the conditional (which you're only allowed to do if there's no `else`).

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,34 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+    fun addSuperinterface(superinterface: KClass<*>, delegate: CodeBlock = CodeBlock.EMPTY)
+        = addSuperinterface(superinterface.asTypeName(), delegate)
 
-    fun addSuperinterface(superinterface: KClass<*>)
-        = addSuperinterface(superinterface.asTypeName())
+    fun addSuperinterface(constructorParameterName: String) = apply {
+      requireNotNull(primaryConstructor) {
+        "delegating to constructor parameter requires not-null constructor" }
+      requireNotNull(primaryConstructor?.parameter(constructorParameterName)?.also {
+        param -> addSuperinterface(param.type, CodeBlock.of(param.name))
```
JakeWharton: The side-effecting use of `also` here seems really egregious. It's hard to follow this code. Can we split the validation and the effect into separate statements?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -18,10 +18,10 @@ package com.squareup.kotlinpoet
 import com.google.common.collect.ImmutableMap
 import com.google.common.truth.Truth.assertThat
 import com.google.testing.compile.CompilationRule
-import com.squareup.kotlinpoet.KModifier.ABSTRACT
 import com.squareup.kotlinpoet.KModifier.INTERNAL
-import com.squareup.kotlinpoet.KModifier.PRIVATE
+import com.squareup.kotlinpoet.KModifier.ABSTRACT
```
JakeWharton: Seems like we lost a bit of alphabetization here... ABSTRACT should be up where it used to be.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2580,7 +2581,119 @@ class TypeSpecTest {
...
+        .addSuperinterface(ParameterizedTypeName.get(Consumer::class, String::class), CodeBlock.of("{ println(it) }"))
+        .build()
+
+    val expect = """
+        |package com.squareup.tacos
+        |
+        |import java.util.function.Consumer
+        |import kotlin.String
+        |
+        |class Guac(somethingElse: String) : Consumer<String> by { println(it) }
```
JakeWharton: This works? That's crazy!

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,34 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
-        = addSuperinterface(superinterface.asTypeName())
+    fun addSuperinterface(superinterface: Type, delegate: CodeBlock = CodeBlock.EMPTY)
+        = addSuperinterface(superinterface.asTypeName(), delegate)
+
+    fun addSuperinterface(superinterface: KClass<*>, delegate: CodeBlock = CodeBlock.EMPTY)
+        = addSuperinterface(superinterface.asTypeName(), delegate)
 
-    fun addSuperinterface(superinterface: KClass<*>)
-        = addSuperinterface(superinterface.asTypeName())
+    fun addSuperinterface(constructorParameterName: String) = apply {
```
JakeWharton: I'm still not sure about this behavior. I was thinking that we'd do `KClass` + `String` since most often you won't need to power of a `CodeBlock` to declare a delegate.

JakeWharton:
Sorry for all the constant comments. You picked a hard feature of the language!

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
prestongarno: turns out it doesn't, only this does `by Consumer({ println(it) })`

prestongarno:
No worries, better than learning kotlin on my own. And not quite as hard as coming up with regexes and `String.replace` statements I've been using for this! Looking forward to cleaning my project up

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,38 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+        superinterfaces += superinterface
+      } else {
+        require(kind == Kind.CLASS) {
+          "delegation only allowed for classes (found $kind '$name')" }
+        require(!superinterface.nullable) {
+          "expected non-nullable type but was '${superinterface.asNonNullable()}'"}
+        require(delegatedSuperInterfaces[superinterface] == null) {
+          "'$name' can not delegate to $superinterface by $delegate with existing declaration by " +
+              "${delegatedSuperInterfaces[superinterface]}" }
+        delegatedSuperInterfaces.put(superinterface, delegate)
```
Egorand: nit: can be `delegatedSuperInterfaces[superinterface] = delegate`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -431,15 +437,38 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+    fun addSuperinterface(superinterface: KClass<*>, constructorParameterName: String) =
+        addSuperinterface(superinterface.asTypeName(), constructorParameterName)
+
+    fun addSuperinterface(superinterface: TypeName, constructorParameter: String) = apply {
+      requireNotNull(primaryConstructor) {
+        "delegating to constructor parameter requires not-null constructor" }
+      val parameter = primaryConstructor?.parameter(constructorParameter)
+      requireNotNull(parameter) {
+        "no such constructor parameter '$constructorParameter' to delegate to for type '$name'" }
+      addSuperinterface(parameter!!.type, CodeBlock.of(parameter.name))
```
Egorand: We should pass `superinterface` here instead of `parameter!!.type`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -467,7 +467,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       val parameter = primaryConstructor?.parameter(constructorParameter)
       requireNotNull(parameter) {
         "no such constructor parameter '$constructorParameter' to delegate to for type '$name'" }
-      addSuperinterface(parameter!!.type, CodeBlock.of(parameter.name))
+      require(parameter!!.type == superinterface) {
```
Egorand: This check will break things like:

```kotlin
class A : Runnable
class B(a: A) : Runnable by a
```

prestongarno: I think about these things after I push, my bad for making this painful o_0

Egorand: No worries! By the way, I think it would be great to have a unit test for this specific case

prestongarno: will do!

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2693,6 +2693,29 @@ class TypeSpecTest {
...
+            .build())
+        .addSuperinterface(ParameterizedTypeName.get(ClassName.bestGuess("TypeBuilder"),
+            ClassName.bestGuess("EntityBuilder"),
+            ClassName.bestGuess("Entity")), "argBuilder")
+        .build()
+
+    assertThat(toString(type)).isEqualTo("""
+          |package com.squareup.tacos
+          |
+          |class EntityBuilder(argBuilder: Payload<EntityBuilder, Entity> = Payload.create()) : TypeBuilder<EntityBuilder, Entity> by argBuilder
```
prestongarno: Don't know exactly where this is handled, but carriage returns aren't inserted here for this long class signature (it normally formats them correctly when adding regular supertypes, if I recall correctly)

Egorand: Normally this is done by adding a `%W` modifier into the `CodeBlock`, that will trigger the `CodeWriter` to emit either a whitespace, or new line if the current one is too long. Can we do this:

```kotlin
if (superTypes.isNotEmpty()) {
  codeWriter.emitCode(superTypes.joinToCode(prefix = " : ", separator = ",%W"))
}
```

Can you please check if that works for you?

prestongarno: Seems it's the other way around, I used:

`codeWriter.emitCode(superTypes.joinToCode(separator = ",%W", prefix = " : "))`

I checked the [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting), and it says that the supertypes list/constructor invocation should be on the same line as the closing parentheses, so I think technically the right way to format this case is by breaks on parameter:

     class EntityBuilder(
          argBuilder: Payload<EntityBuilder, Entity> = Payload.create()
    ) : TypeBuilder<EntityBuilder, Entity> by argBuilder

It looks like it'd need context/lookaround when formatting the entire type with constructor, name, etc. to do that esp. since prefixing all superinterfaces with a break is allowed if there's a superclass ctor invocation, so for now I just added the separator as shown above which formats this example from the tests correctly:

        class Guac(somethingElse: String) : RandomIface by Factory.byStrategy(FacadeImpl),
            Runnable by Runnable({ Logger.debug("Hello world") })

Egorand: Looks good! In any case, I don't feel like proper styling and wrapping is that important, since this is generated code. Wrapping long lines doesn't hurt though.

Egorand:
@swankjesse @JakeWharton any extra feedback on this PR?

JakeWharton:
I'll look again today.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -336,6 +341,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     internal var superclass: TypeName = ANY
     internal val superclassConstructorParameters = mutableListOf<CodeBlock>()
     internal val superinterfaces = mutableListOf<TypeName>()
+    internal val delegatedSuperInterfaces = mutableMapOf<TypeName, CodeBlock>()
```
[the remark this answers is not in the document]
JakeWharton: I wouldn't care about this if it was internal only, but it means we expose both the list and the map publicly once you build the object, which is where I take most of the issue.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -134,9 +136,12 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
           CodeBlock.of("%T(%L)", it, superclassConstructorParameters.joinToCode())
         }
-        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) }
+        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) } +
+            delegatedSuperInterfaces.entries.map {
+              (type, init) -> CodeBlock.of("%T by $init", type) }
+
         if (superTypes.isNotEmpty()) {
-          codeWriter.emitCode(superTypes.joinToCode(prefix = " : "))
+          codeWriter.emitCode(superTypes.joinToCode(separator = ",%W", prefix = " : "))
```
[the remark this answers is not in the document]
Egorand: Makes sense!

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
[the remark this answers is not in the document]
Egorand: The name should be more descriptive then, what about `superinterfacesWithDelegates`?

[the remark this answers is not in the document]
JakeWharton: I'd be tempted to keep `superinterfaces` and add documentation to this property describing what the values represent.

Egorand:
@prestongarno would be great to finalize this PR, it's just a couple small things away from being ready. Do you need any help?

prestongarno:
@Egorand My bad, forgot about this since I just used my fork as dependency. I made the changes to what I think @JakeWharton was talking about, but if there's anything missing/lacking let me know and if you want to I wouldn't mind if you went ahead and finalized them, thanks!

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -133,9 +137,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
             CodeBlock.of("%T", it)
           }
         }
-        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) }
+        val superTypes = types + superinterfaces.entries.map { (type, init) ->
```
Egorand: nit: please place the closing bracket on its own line: 

```kotlin
...map { (type, init) ->
  // do smth
}
```

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -133,9 +137,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+              if(init == null)  CodeBlock.of("%T", type) else CodeBlock.of("%T by $init", type) }
```
Egorand: nit: space after `if` missing

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -423,18 +429,42 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addSuperinterfaces(superinterfaces: Iterable<TypeName>) = apply {
-      this.superinterfaces += superinterfaces
+      this.superinterfaces.putAll(superinterfaces.map {
+        Pair(it, null) })
```
Egorand: nit: same observation about the closing brackets

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -524,7 +554,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       }
 
       val superclassIsAny = superclass == ANY
-      val interestingSupertypeCount = (if (superclassIsAny) 0 else 1) + superinterfaces.size
+      val interestingSupertypeCount = (if (superclassIsAny) 0 else 1) + this.superinterfaces.size
```
Egorand: nit: using `this` here doesn't seem to add any value

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -32,7 +32,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val primaryConstructor = builder.primaryConstructor
   val superclass = builder.superclass
   val superclassConstructorParameters = builder.superclassConstructorParameters.toImmutableList()
-  val superinterfaces = builder.superinterfaces.toImmutableList()
+  /**
+   * Map of superinterfaces - entries with a null value represent a regular superinterface (with
+   * no delegation), while non-null [com.squareup.kotlinpoet.CodeBlock] values represent delegates
+   * for the corresponding [com.squareup.kotlinpoet.TypeSpec] interface (key) value */
```
swankjesse: nit: typically we do `*/` on its own line

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -423,18 +430,43 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addSuperinterfaces(superinterfaces: Iterable<TypeName>) = apply {
-      this.superinterfaces += superinterfaces
+      this.superinterfaces.putAll(superinterfaces.map {
```
swankjesse: off-topic: this seems like an inefficient idiom. I think it’s fine for this project but it makes me anxious because it creates a lot of temporary stuff for syntactic benefit

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -423,18 +430,43 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+        Pair(it, null) 
+      })
     }
 
-    fun addSuperinterface(superinterface: TypeName) = apply {
-      superinterfaces += superinterface
+    fun addSuperinterface(superinterface: TypeName, delegate: CodeBlock = CodeBlock.EMPTY) = apply {
+      if (delegate.isEmpty()) {
+        this.superinterfaces.put(superinterface, null)
```
swankjesse: Should we use `CodeBlock.EMPTY` instead of null in the map?

swankjesse: (seems simpler to have a single value that means “no delegate” rather than two)

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
swankjesse: Fix the test?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2645,6 +2648,126 @@ class TypeSpecTest {
...
+        |import java.lang.Runnable
+        |import kotlin.String
+        |
+        |class Guac(somethingElse: String) : RandomIface by Factory.byStrategy(FacadeImpl),
+        |    Runnable by Runnable({ Logger.debug("Hello world") })
+        |""".trimMargin()
+
+    assertThat(toString(type)).isEqualTo(expect)
+  }
+  @Test fun testNoSuchParameterDelegate() {
```
swankjesse: nit: blank line between tests

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2645,6 +2648,126 @@ class TypeSpecTest {
...
+    try {
+      TypeSpec.classBuilder("Taco")
+          .primaryConstructor(FunSpec.constructorBuilder()
+              .addParameter("other", String::class)
+              .build())
+          .addSuperinterface(KFunction::class, "notOther")
+          .build()
+    } catch (expected: IllegalArgumentException) {
```
swankjesse: please use `assertThrows()`

```
   TypeSpec.classBuilder()
       ...
}
```

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2645,6 +2648,126 @@ class TypeSpecTest {
...
+      assertThat(expected).hasMessage("no such constructor parameter 'notOther' to delegate to for type 'Taco'")
+    }
+  }
+
+  @Test fun failAddParamDelegateWhenNullCtor() {
+    try {
```
swankjesse: same

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2645,6 +2648,126 @@ class TypeSpecTest {
...
+      TypeSpec.classBuilder("Taco")
+          .addSuperinterface(Runnable::class, "etc")
+          .build()
+    } catch (expected: IllegalArgumentException) {
+      assertThat(expected).hasMessage("delegating to constructor parameter requires not-null constructor")
+    }
+  }
+
+  @Test fun testAddedDelegateByParamName() {
```
swankjesse: cool!

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2677,8 +2800,8 @@ class TypeSpecTest {
       |)
       |""".trimMargin())
   }
-
```
swankjesse: nit: don’t remove this?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: Actually that might be cleaner. I'll push formatting fix but wait for other feedback on this (since I've already re-done a few times) - would it be possible to do something more specific than just `CodeBlock`s and nulls?

prestongarno: something that could have standard lib delegate & operator function support too?

Egorand: @swankjesse I remember we've discussed sentinels on an earlier PR. I think `null`s play better in Kotlin (might not be true for Java) as the compiler ensures you don't miss the checks. If your sentinel is `CodeBlock.EMPTY` you have to be careful to check against it everywhere, and it feels super error-prone.

swankjesse: Agreed in general. In this case the code is already actively checking if the code block is empty and mapping that to null. It seems bad to use empty code blocks as a sentinel in the API but nulls in the implementation.

I don't feel strongly here and to unblock this, let's keep it as is!

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2645,6 +2648,125 @@ class TypeSpecTest {
...
+
+  @Test fun testNoSuchParameterDelegate() {
+    assertThrows<IllegalArgumentException> {
+      TypeSpec.classBuilder("Taco")
+          .primaryConstructor(FunSpec.constructorBuilder()
+              .addParameter("other", String::class)
+              .build())
+          .addSuperinterface(KFunction::class, "notOther")
+          .build()
+    }.hasMessage("no such constructor parameter 'notOther' to delegate to for type 'Taco'")
```
Egorand: `hasMessage()` is deprecated in the newest version of Truth, please use `hasMessageThat().isEqualTo()`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -133,9 +139,12 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
             CodeBlock.of("%T", it)
           }
         }
-        val superTypes = types + superinterfaces.map { CodeBlock.of("%T", it) }
+        val superTypes = types + superinterfaces.entries.map { (type, init) ->
+            if (init == null)  CodeBlock.of("%T", type) else CodeBlock.of("%T by $init", type)
```
swankjesse: I think there’s a small problem here. This calls `$init.toString()` eagerly but we really want to embed the codeblock using `%L`. Otherwise imports will be wrong if the codeblock contains ’em.

swankjesse: ie. `CodeBlock.of(%T by %L, type, init)`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -423,18 +432,43 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+    fun addSuperinterface(superinterface: KClass<*>, constructorParameterName: String) =
+        addSuperinterface(superinterface.asTypeName(), constructorParameterName)
+
+    fun addSuperinterface(superinterface: TypeName, constructorParameter: String) = apply {
+      requireNotNull(primaryConstructor) {
+        "delegating to constructor parameter requires not-null constructor" }
+      val parameter = primaryConstructor?.parameter(constructorParameter)
+      requireNotNull(parameter) {
+        "no such constructor parameter '$constructorParameter' to delegate to for type '$name'" }
+      addSuperinterface(superinterface, CodeBlock.of(constructorParameter))
```
swankjesse: This is academic, but strictly speaking we shouldn’t use `CodeBlock.of()` to turn a literal name into a codeblock. The problem is that the literal name could have a metacharacter and we want to escape it.

Better would be `CodeBlock.of("%L", constructorParameter)`. That way if you pass in the _string_ `%[` that’s what we’ll echo out. As-is that’ll mess up our indentation and it’ll be difficult to diagnose why.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2643,6 +2646,123 @@ class TypeSpecTest {
...
+    assertThat(toString(type)).isEqualTo(expect)
+  }
+
+  @Test fun testMultipleDelegates() {
+    val type = TypeSpec.classBuilder("Guac")
+        .primaryConstructor(FunSpec.constructorBuilder()
+            .addParameter("somethingElse", String::class)
+            .build())
+        .addSuperinterface(ClassName.bestGuess("RandomIface"), CodeBlock.of("Factory.byStrategy(FacadeImpl)"))
+        .addSuperinterface(Runnable::class, CodeBlock.of("Runnable({ Logger.debug(\"Hello world\") })"))
```
swankjesse: I don’t think this is a good test because it shows a type `Logger` being embedded directly rather than with the proper syntax `%T`. We expect our users to read our test cases and use them as examples, and this is one where we’re taking a shortcut we don’t want to take.

The big consequence is that imports don’t work.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2643,6 +2646,123 @@ class TypeSpecTest {
...
+        |import kotlin.String
+        |
+        |class Guac(somethingElse: String) : RandomIface by Factory.byStrategy(FacadeImpl),
+        |    Runnable by Runnable({ Logger.debug("Hello world") })
+        |""".trimMargin()
+
+    assertThat(toString(type)).isEqualTo(expect)
+  }
+
+  @Test fun testNoSuchParameterDelegate() {
```
swankjesse: Nice test

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2643,6 +2646,123 @@ class TypeSpecTest {
...
+      TypeSpec.classBuilder("Taco")
+          .primaryConstructor(FunSpec.constructorBuilder()
+              .addParameter("other", String::class)
+              .build())
+          .addSuperinterface(KFunction::class, "notOther")
+          .build()
+    }.hasMessageThat().isEqualTo("no such constructor parameter 'notOther' to delegate to for type 'Taco'")
+  }
+
+  @Test fun failAddParamDelegateWhenNullCtor() {
```
swankjesse: nit: I don’t like custom abbreviations like `Ctor` because it’s difficult to guess what abbreviation will be used, and that makes grepping more difficult

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
prestongarno: I was wondering why the imports didn't work consistently

Egorand:
Hey @prestongarno !

It'd be great to have this merged for 0.8.0! Could you please rebase and go over Jesse's latest comments and see if there's anything to improve? 

Thanks again!

Egorand:
Merged manually. Thanks @prestongarno !

prestongarno:
Was about to push changes, but thanks for taking care of that! And I'm glad I could help.

Egorand:
@prestongarno forgot to ask you to sign the CLA, could you do that please? http://squ.re/sign-the-cla

prestongarno:
Done.

Egorand:
Thanks!

PR #202, opened 2017-08-24
Add support for named companion objects

### LINKED ISSUE #201: Support named companion objects
At the moment it's not possible to create a companion object `TypeSpec` with a name, there's only one factory method:

```kotlin
@JvmStatic fun companionObjectBuilder() = Builder(Kind.COMPANION, null, null)
```

Egorand:
Closes #201

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -90,7 +90,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitAnnotations(annotations, false)
         codeWriter.emitModifiers(modifiers, setOf(PUBLIC))
         codeWriter.emit(kind.declarationKeyword)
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -534,6 +534,8 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
     @JvmStatic fun companionObjectBuilder() = Builder(Kind.COMPANION, null, null)
 
+    @JvmStatic fun companionObjectBuilder(name: String) = Builder(Kind.COMPANION, name, null)
```
JakeWharton: Should this be nullable and have a default of null (and `@JvmOverloads`). It's not very idiomatic Kotlin to have manual overloads like this.

Egorand: Yeah, you're right, will fix.

PR #203, opened 2017-08-26
Rename addFun() to addFunction() in KotlinFile

Egorand:
I accidentally broke README with #195 by replacing `addFun` with `addFunction`, as `KotlinFile` also had a method with this name. Since we renamed it in `FunSpec`, I think it makes sense to rename it in `KotlinFile` for consistency.


JakeWharton:
I'm tempted to rename FunSpec to FunctionSpec... Must. Resist.

Egorand:
Since there's no way to `addFun()` anymore, we have to keep all remaining fun in place. So no.

PR #204, opened 2017-08-26
Add minimal DSL support

Egorand:
Since adding this was trivial, I decided to submit it for review. Although the test doesn't showcase it very well, I think in some cases this approach might make the code less verbose. And as the priority right now is correctness, I'm fine with just keeping the PR around, might be a source of ideas for upcoming attempts to provide DSL-like API.


JakeWharton:
I'm not sure we want to add this if we're only going to break it later to create an actual DSL with less builder ceremony. Also this doesn't offer much more value than using `apply { }`.

Egorand:
Agree. The only advantage over `apply` is not needing to call `build()`, which is not that big of a deal. Closing.

PR #207, opened 2017-08-31
Support constructor delegation

### LINKED ISSUE #205: Secondary constructors
I want to have a class with multiple constructors, like the following:
```kotlin
class Test(val a: String?, val b: String?, val c: String?) {
    constructor(map: Map<String, String>) : this(map["a"], map["b"], map["c"])
}
```
While I am able to add a secondary constructor using FunSpec.constructorBuilder() and TypeSpec.Builder.addFun(), I can't add the `: this(...)` to the end of the constructor definition.

Egorand:
Fixes #205

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -39,6 +39,7 @@ class FunSpec private constructor(builder: Builder) {
   val receiverType = builder.receiverType
   val returnType = builder.returnType
   val parameters = builder.parameters.toImmutableList()
+  val delegateConstructorParameters = builder.delegateConstructorParameters.toImmutableList()
   val exceptions = builder.exceptions.toImmutableList()
   val body = builder.body.build()
 
@@ -73,7 +74,10 @@ class FunSpec private constructor(builder: Builder) {
     emitSignature(codeWriter, enclosingName)
     codeWriter.emitWhereBlock(typeVariables)
 
-    if (modifiers.contains(KModifier.ABSTRACT) || modifiers.contains(KModifier.EXTERNAL)) {
+    val isEmptyConstructor = isConstructor && delegateConstructorParameters.isNotEmpty() &&
+        body.isEmpty()
+    if (modifiers.contains(KModifier.ABSTRACT) || modifiers.contains(KModifier.EXTERNAL) ||
+        isEmptyConstructor) {
       codeWriter.emit("\n")
       return
     }
@@ -112,6 +116,11 @@ class FunSpec private constructor(builder: Builder) {
 
     if (returnType != null) {
       codeWriter.emitCode(": %T", returnType)
+    }
+
+    if (delegateConstructorParameters.isNotEmpty()) {
+      codeWriter.emitCode(delegateConstructorParameters
+          .joinToCode(prefix = " : this(", suffix = ")"))
     }
 
     if (exceptions.isNotEmpty()) {
@@ -169,6 +178,7 @@ class FunSpec private constructor(builder: Builder) {
     builder.typeVariables += typeVariables
     builder.returnType = returnType
     builder.parameters += parameters
+    builder.delegateConstructorParameters += delegateConstructorParameters
     builder.exceptions += exceptions
     builder.body.add(body)
     return builder
@@ -182,6 +192,7 @@ class FunSpec private constructor(builder: Builder) {
     internal var receiverType: TypeName? = null
     internal var returnType: TypeName? = null
     internal val parameters = mutableListOf<ParameterSpec>()
+    internal val delegateConstructorParameters = mutableListOf<CodeBlock>()
     internal val exceptions = mutableSetOf<TypeName>()
     internal val body = CodeBlock.builder()
 
@@ -281,6 +292,14 @@ class FunSpec private constructor(builder: Builder) {
       check(name != GETTER) { "$name cannot have parameters" }
       check(name != SETTER || parameters.size == 0) { "$name can have only one parameter" }
       parameters += parameterSpec
+    }
+
+    fun addDelegateParameters(vararg parameters: String) =
+        addDelegateParameters(*parameters.map { CodeBlock.of(it) }.toTypedArray())
+
+    fun addDelegateParameters(vararg parameters: CodeBlock) = apply {
+      check(name.isConstructor) { "only constructors can have delegate parameters!"}
+      delegateConstructorParameters += parameters
     }
 
     fun addParameter(name: String, type: TypeName, vararg modifiers: KModifier)
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -301,12 +301,6 @@ class FunSpec private constructor(builder: Builder) {
       callConstructor("this", *args.map { CodeBlock.of(it) }.toTypedArray())
     }
 
-    private fun callConstructor(constructor: String, vararg args: CodeBlock) {
-      check(name.isConstructor) { "only constructors can delegate to other constructors!" }
-      delegateConstructor = constructor
-      delegateConstructorArguments += args
-    }
-
     fun callThisConstructor(vararg args: CodeBlock = emptyArray()) = apply {
       callConstructor("this", *args)
     }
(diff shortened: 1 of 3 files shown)


Egorand:
@wooky hope the API makes sense, would appreciate some feedback

wooky:
Looks good! 👍

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -114,6 +118,11 @@ class FunSpec private constructor(builder: Builder) {
       codeWriter.emitCode(": %T", returnType)
     }
 
+    if (delegateConstructorParameters.isNotEmpty()) {
+      codeWriter.emitCode(delegateConstructorParameters
+          .joinToCode(prefix = " : this(", suffix = ")"))
```
JakeWharton: What about calling `super`?

Egorand: I can extend the API to allow for specifying which constructor to delegate to, `super` or `this`.

Egorand:
I wonder if `ConstructorSpec` could be a good idea since there are more differences between constructors and functions in Kotlin than in Java.

JakeWharton:
It's worth looking into, yeah. I believe that was the case in JavaPoet before we merged them into one.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -73,7 +75,10 @@ class FunSpec private constructor(builder: Builder) {
     emitSignature(codeWriter, enclosingName)
     codeWriter.emitWhereBlock(typeVariables)
 
-    if (modifiers.contains(KModifier.ABSTRACT) || modifiers.contains(KModifier.EXTERNAL)) {
+    val isEmptyConstructor = isConstructor && delegateConstructorParameters.isNotEmpty() &&
+        body.isEmpty()
+    if (modifiers.contains(KModifier.ABSTRACT) || modifiers.contains(KModifier.EXTERNAL) ||
```
JakeWharton:
```kotlin
ABSTRACT in modifiers || EXTERNAL in modifiers || isEmptyConstructor
```

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -283,6 +297,26 @@ class FunSpec private constructor(builder: Builder) {
       parameters += parameterSpec
     }
 
+    fun callThisConstructorWithParameters(vararg parameters: String) =
```
JakeWharton: This name is a bit much. Can we do any better? Maybe `delegateConstructor`?

Egorand: `this` and `super` still have to be in the name, what about `delegateToThisConstructor` and `delegateToSuperConstructor`?

JakeWharton: We could expose an enum that lets you pick between the two delegates.

Egorand: Can do. I hope to try `ConstructorSpec` soon, part of the problem with naming comes from the need to clarify that the methods should be used with the `constructorBuilder()`.

JakeWharton:
I think I'm okay with this for now, under the assumption we'll explore a constructor-specific spec in the future.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -169,11 +179,17 @@ class FunSpec private constructor(builder: Builder) {
...
     builder.returnType = returnType
     builder.parameters += parameters
+    builder.delegateConstructor = delegateConstructor
+    builder.delegateConstructorParameters += delegateConstructorParameters
     builder.exceptions += exceptions
     builder.body.add(body)
     return builder
   }
 
+  enum class Constructor(internal val keyword: String) {
```
swankjesse: Not sure about this name. ConstructorCall ?

swankjesse: DelegateConstructor?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -283,6 +301,17 @@ class FunSpec private constructor(builder: Builder) {
...
 
+    fun delegateConstructor(constructor: Constructor, vararg parameters: String) =
+        delegateConstructor(constructor, *parameters.map { CodeBlock.of(it) }.toTypedArray())
+
+    fun delegateConstructor(
+        constructor: Constructor,
+        vararg parameters: CodeBlock = emptyArray()) = apply {
+      check(name.isConstructor) { "only constructors can have delegate parameters!" }
+      delegateConstructor = constructor
+      delegateConstructorParameters += parameters
```
swankjesse: The `+=` call seems like it’s broken for repeated calls? Ie. it adds, rather than replaces?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -256,6 +258,49 @@ class FunSpecTest {
...
+
+    assertThat(funSpec.toString()).isEqualTo("""
+      |constructor(list: kotlin.collections.List<kotlin.Int>) : this(list[0], list[1])
+      |""".trimMargin())
+  }
+
+  @Test fun superConstructorDelegate() {
+    val funSpec = FunSpec.constructorBuilder()
+        .addParameter("list", ParameterizedTypeName.get(List::class, Int::class))
+        .delegateConstructor(SUPER, "list[0]", "list[1]")
```
swankjesse: would the API be simpler if there were just two methods, `thisConstructor` and `superConstructor` ?

swankjesse: Then there’s no `Constructor` type, which seems like a simplification to me

swankjesse: Or should these args go on `constructorBuilder()` above?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -256,6 +258,49 @@ class FunSpecTest {
...
+        .addParameter("a", Int::class)
+        .delegateConstructor(THIS)
+        .build()
+
+    assertThat(funSpec.toString()).isEqualTo("""
+      |constructor(a: kotlin.Int) : this()
+      |""".trimMargin())
+  }
+
+  @Test fun addingDelegateParametersToNonConstructorForbidden() {
```
swankjesse: Nice

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
Egorand: Indeed, good call. I guess it would make sense to store them in a set, will fix.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
Egorand: Would it be clear enough if we drop the `delegate` part (i.e. `delegateThisConstructor()` vs `thisConstructor()`)? I agree with having two methods vs the enum param, that was what the initial version looked like (albeit a lot more verbose), and I guess I like it better. 

For the `constructorBuilder()` args, we'll need to have different versions of it for `this` and `super`, and I can't really think of good names. `thisConstructorBuilder()` and `superConstructorBuilder()`? And keeping the original `constructorBuilder()` with no args?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
Egorand: On a closer look, using a set doesn't make sense, as you can pass a single `CodeBlock` as an argument for multiple parameters. In this case, appending to the list rather than replacing seems perfectly fine to me. Do we want API to allow replacing or removing the args?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -283,6 +297,28 @@ class FunSpec private constructor(builder: Builder) {
       parameters += parameterSpec
     }
 
+    fun callThisConstructor(vararg args: String) = apply {
+      callConstructor("this", *args.map { CodeBlock.of(it) }.toTypedArray())
+    }
+
+    private fun callConstructor(constructor: String, vararg args: CodeBlock) {
```
JakeWharton: Can you move this to the bottom of all four methods. Overloads have to stay next to each other and I'd prefer the this/super functions also be grouped.

Egorand: Yep, done

PR #211, opened 2017-09-02
Add header and impl to modifiers enum

Egorand:


src/main/java/com/squareup/kotlinpoet/KModifier.kt
```diff
@@ -61,7 +61,11 @@ enum class KModifier(
   // Type modifiers.
   IN("in", Target.VARIANCE_ANNOTATION),
   OUT("out", Target.VARIANCE_ANNOTATION),
-  VARARG("vararg", Target.PARAMETER);
+  VARARG("vararg", Target.PARAMETER),
+
+  // Multiplatform modules.
+  HEADER("header", Target.CLASS),
+  IMPL("impl", Target.CLASS);
```
JakeWharton: This is also valid in a function and probably a property

Egorand: Is there any documentation btw?

Egorand:
These modifiers are not mentioned in https://github.com/yole/kotlin-style-guide/issues/3

src/main/java/com/squareup/kotlinpoet/KModifier.kt
JakeWharton: Uhh. I don't think there's much if any. I can point you to a bunch of example usage though.

Egorand: Sure, that'd be great

JakeWharton: https://github.com/JakeWharton/Reagent has the referenced atomic thing from the issue.

The others are private. Can invite later when I'm not mobile. Beyond that you can search "Kotlin multiplatform Gradle" and find a few others' samples.

Egorand: Thanks!

JakeWharton:
Damn I forgot to squash

PR #217, opened 2017-09-06
Tests and fixes for crossplatform code

Egorand:
Some changes are necessary to make this work, since in a default class all methods have to have a body. Solution implemented here:

- Add `KModifier.HEADER` to the list of implicit function modifiers in `Kind.CLASS`
- Mark `TypeSpec` and all `FunSpec`s with `KModifier.HEADER`

An alternative solution would be:

- Introduce a new `Kind` called `HEADER_CLASS` with `KModifier.ABSTRACT` as implicit function modifier
- Mark `FunSpec`s with `KModifier.ABSTRACT`

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -405,3 +405,5 @@ fun Collection<CodeBlock>.joinToCode(separator: CharSequence = ", ", prefix: Cha
   val placeholders = Array(blocks.size, { "%L" })
   return CodeBlock.of(placeholders.joinToString(separator, prefix, suffix), *blocks)
 }
+
+internal fun CodeBlock?.isNullOrEmpty() = this == null || isEmpty()
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -405,5 +405,3 @@ fun Collection<CodeBlock>.joinToCode(separator: CharSequence = ", ", prefix: Cha
   val placeholders = Array(blocks.size, { "%L" })
   return CodeBlock.of(placeholders.joinToString(separator, prefix, suffix), *blocks)
 }
-
-internal fun CodeBlock?.isNullOrEmpty() = this == null || isEmpty()
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -286,6 +286,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         setOf(KModifier.PUBLIC),
         setOf(KModifier.PUBLIC)),
 
+    HEADER_CLASS(
+        "class",
+        setOf(KModifier.PUBLIC, KModifier.HEADER),
+        setOf(KModifier.PUBLIC, KModifier.HEADER)),
+
     OBJECT(
         "object",
         setOf(KModifier.PUBLIC),
@@ -381,7 +386,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun primaryConstructor(primaryConstructor: FunSpec?) = apply {
-      check(kind.isOneOf(Kind.CLASS, Kind.ENUM, Kind.ANNOTATION)) {
+      check(kind.isOneOf(Kind.CLASS, Kind.HEADER_CLASS, Kind.ENUM, Kind.ANNOTATION)) {
         "$kind can't have initializer blocks"
       }
       if (primaryConstructor != null) {
@@ -399,7 +404,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     private fun ensureCanHaveSuperclass() {
-      check(kind.isOneOf(Kind.CLASS, Kind.OBJECT, Kind.COMPANION)) {
+      check(kind.isOneOf(Kind.CLASS, Kind.HEADER_CLASS, Kind.OBJECT, Kind.COMPANION)) {
         "only classes can have super classes, not $kind"
       }
     }
@@ -443,10 +448,18 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addProperties(propertySpecs: Iterable<PropertySpec>) = apply {
-      this.propertySpecs += propertySpecs
+      propertySpecs.map(this::addProperty)
     }
 
     fun addProperty(propertySpec: PropertySpec) = apply {
+      if (kind == Kind.HEADER_CLASS) {
+        require(propertySpec.initializer.isNullOrEmpty()) {
+          "properties in header classes can't have initializers"
+        }
+        require(propertySpec.getter == null && propertySpec.setter == null) {
+          "properties in header classes can't have getters and setters"
+        }
+      }
       propertySpecs += propertySpec
     }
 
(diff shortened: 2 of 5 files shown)


JakeWharton:
Needs rebase.

JakeWharton:
If a class is a header then all functions cannot have a body and all properties cannot have an initializer or get/set functions. So you shouldn't need to specify it and we should error if you try to add a FunSpec or PropertySpec with those things.

Egorand:
Looks like secondary constructors are allowed, but constructor delegation isn't. Is there any practical value in having them?

JakeWharton:
In having what? Secondary constructors? They're needed. Here's another real example:
```kotlin
header open class IoException : Exception {
  constructor()
  constructor(message: String)
}
```

Egorand:
Will add tests for that

Egorand:
@JakeWharton anything else you can think of?

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,167 @@
...
+      |header open class IoException : Exception {
+      |  constructor()
+      |
+      |  constructor(message: String)
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun initBlockInHeaderForbidden() {
+    try {
```
JakeWharton: can you use `assertThrows` for all these?

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,167 @@
...
+      TypeSpec.headerClassBuilder("AtomicRef<V>")
```
JakeWharton: nit: delete `<V>` from all of these?

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        .addFunction(FunSpec.constructorBuilder()
+            .addParameter("message", String::class)
+            .build())
+        .build()
+    val fileSpec = FileSpec.builder("", "Test")
+        .addType(headerSpec)
+        .build()
+
+    assertThat(fileSpec.toString()).isEqualTo("""
+      |import java.lang.Exception
```
JakeWharton: This is weird! It should be `kotlin.Exception`. I'll have to look into that...

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,167 @@
...
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.test.fail
+
+class CrossplatformTest {
+
+  @Test fun crossplatform() {
+    val headerTypeParam = TypeVariableName("V")
+    val headerType = "AtomicRef<V>"
```
JakeWharton: `<V>` shouldn't be in the name.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -437,16 +444,25 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
     fun addProperty(propertySpec: PropertySpec) = apply {
+      if (kind == Kind.HEADER_CLASS) {
+        require(propertySpec.initializer.isNullOrEmpty()) {
+          "properties in header classes can't have initializers"
+        }
+        require(propertySpec.getter == null && propertySpec.setter == null) {
+          "properties in header classes can't have getters and setters"
+        }
+      }
```
JakeWharton: very nice

JakeWharton:
Also please make sure you squash before merging. I forgot last time.

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
Egorand: `kotlin.Exception` is a type alias, so there's only `java.lang.Exception` at runtime. Guess the only way to fix this is to have some manual replacement logic.

JakeWharton: Yeah I mean I get why it's happening, but reflection-wise it's weird and it means there's weird implications for code generation now.

JakeWharton: I'd rather see it behave like `Any` which gets turned into `Object` but you don't see it.

Egorand:
Got blocked a bit by #227, gonna fix that tomorrow and finish the squash

JakeWharton:
Just sent PR

Egorand:
Should be complete now

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -405,3 +405,5 @@ fun Collection<CodeBlock>.joinToCode(separator: CharSequence = ", ", prefix: Cha
   val placeholders = Array(blocks.size, { "%L" })
   return CodeBlock.of(placeholders.joinToString(separator, prefix, suffix), *blocks)
 }
+
+internal fun CodeBlock?.isNullOrEmpty() = this == null || isEmpty()
```
swankjesse: Ick?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -289,6 +289,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         setOf(KModifier.PUBLIC),
         setOf(KModifier.PUBLIC)),
 
+    HEADER_CLASS(
```
swankjesse: Nice.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -439,16 +444,25 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
     }
 
     fun addProperties(propertySpecs: Iterable<PropertySpec>) = apply {
-      this.propertySpecs += propertySpecs
+      propertySpecs.map(this::addProperty)
     }
 
     fun addProperty(propertySpec: PropertySpec) = apply {
+      if (kind == Kind.HEADER_CLASS) {
+        require(propertySpec.initializer.isNullOrEmpty()) {
```
swankjesse: Is a null check sufficient? Is it ever empty? (Or vice-versa!)

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,153 @@
...
+ * limitations under the License.
+ */
+
+package com.squareup.kotlinpoet
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import java.util.concurrent.atomic.AtomicReference
+
+class CrossplatformTest {
```
swankjesse: Awesome

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,153 @@
...
+    val fileSpec = FileSpec.builder("", "Test")
+        .addType(headerSpec)
+        .addTypeAlias(implSpec)
+        .build()
+
+    assertThat(fileSpec.toString()).isEqualTo("""
+      |import java.util.concurrent.atomic.AtomicReference
+      |import kotlin.Boolean
+      |
+      |header internal class AtomicRef<V>(value: V) {
```
swankjesse: I'm surprised the access modifier splits `header` and `class`. I think of it as one keyword!

swankjesse: (though if this matches the jetbrains' order, keep it!)

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
```diff
@@ -0,0 +1,153 @@
...
+      |header open class IoException : Exception {
+      |  constructor()
+      |
+      |  constructor(message: String)
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun initBlockInHeaderForbidden() {
+    assertThrows<IllegalStateException> {
```
swankjesse: This is great, @jakewharton!

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
JakeWharton: They haven't defined the order of `header` or `impl` in relation to other keywords yet.

Egorand: It would indeed be more convenient to emit `header class` together, also `internal header class` reads better IMO

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Currently `initializer` can be both, since it's optional and we don't prohibit passing in an empty `CodeBlock`. I'll see if there are valid use-cases for the latter, and if not - I'd go with disallowing an empty `CodeBlock` as `initializer` in `PropertySpec`.

src/test/java/com/squareup/kotlinpoet/CrossplatformTest.kt
Egorand: Actually, we don't want any initializers, so I guess a null-check is sufficient. Pushed cb3c5e4

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Actually, we don't want any initializers, so I guess a null-check is sufficient. Pushed cb3c5e4

PR #219, opened 2017-09-06
Rename KotlinFile to FileSpec.

JakeWharton:


JakeWharton:
Now how long until someone files a bug that the README is wrong for 0.4.0?

PR #222, opened 2017-09-07
A bunch of code cleanups mostly for style.

JakeWharton:
Wherein I make a PR exactly like those I hate receiving: code style cleanups for every file!


JakeWharton:
Got carried away while looking for something useful to do...

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -302,14 +292,11 @@ internal class CodeWriter constructor(
...
   private fun emitLiteral(o: Any?) {
-    if (o is TypeSpec) {
-      o.emit(this, null)
-    } else if (o is AnnotationSpec) {
-      o.emit(this, inline = true, asParameter = true)
-    } else if (o is CodeBlock) {
-      emitCode(o)
-    } else {
-      emit(o.toString())
+    when (o) {
```
Egorand: Could've also been `= when(o) {`. Do you prefer braces when the return type is `Unit`?

JakeWharton: Not necessarily. I'm not sure why I didn't do it for this one. Will look later.

src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -88,12 +88,12 @@ class ClassName internal constructor(
         packageName + "." + names[1]
     }
     // concat top level class name and nested names
-    val builder = StringBuilder()
-    builder.append(topLevelClassName())
-    for (name in simpleNames().subList(1, simpleNames().size)) {
-      builder.append('$').append(name)
+    return buildString {
```
swankjesse: 🆒

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -311,30 +315,26 @@ class CodeBlock private constructor(
...
-        is TypeSpec -> return o.name!!
-        else -> throw IllegalArgumentException("expected name but was " + o)
-      }
+    private fun argToName(o: Any?) = when (o) {
+      is CharSequence -> o.toString()
+      is ParameterSpec -> o.name
+      is PropertySpec -> o.name
+      is FunSpec -> o.name
+      is TypeSpec -> o.name!!
+      else -> throw IllegalArgumentException("expected name but was " + o)
```
swankjesse: Nit: template string

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -445,7 +432,6 @@ internal class CodeWriter constructor(
       out.append(line)
       trailingNewline = false
     }
-    return this
```
swankjesse: Is there a benefit to returning self in Kotlin? Is it idiomatic?

JakeWharton: There is little benefit because of `apply { }` blocks. I was actually considering migrating to `CodeWriter` uses `apply` as a follow-up.

JakeWharton: Just tried it. It works in some areas and doesn't in others.

Things like
```kotlin
if (index > 0) codeWriter.emit(",").emitWrappingSpace()
```
are forced to become
```kotlin
if (index > 0) {
  codeWriter.apply {
    emit(", ")
    emitWrappingSpace()
  }
}
```

swankjesse: I like the first one!

PR #224, opened 2017-09-07
Do not invoke superclass ctor when no primary ctor.

### LINKED ISSUE #223: Fix superclass emission when primary constructor is missing
Test case:

```kotlin
@Test fun classExtendsNoPrimaryConstructor() {
  val typeSpec = TypeSpec.classBuilder("IoException")
      .superclass(Exception::class)
      .addFunction(FunSpec.constructorBuilder().build())
      .build()

  assertThat(toString(typeSpec)).isEqualTo("""
      |package com.squareup.tacos
      |
      |import java.lang.Exception
      |
      |class IoException : Exception() {
      |  constructor() {
      |  }
      |}
      |""".trimMargin())
}
```

The output fails to compile on the `Exception()` part with `Supertype initialization is impossible without primary constructor`. Correct syntax is `Exception` without braces.

JakeWharton:
And when secondary ctors are present.

Closes #223.

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
@@ -2498,6 +2498,52 @@ class TypeSpecTest {
     }
   }
 
+  @Test fun classExtendsNoPrimaryConstructor() {
+    val typeSpec = TypeSpec.classBuilder("IoException")
+        .superclass(Exception::class)
+        .addFunction(FunSpec.constructorBuilder().build())
+        .build()
+
+    assertThat(toString(typeSpec)).isEqualTo("""
+      |package com.squareup.tacos
+      |
+      |import java.lang.Exception
+      |
+      |class IoException : Exception {
+      |  constructor() {
+      |  }
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun classExtendsNoPrimaryOrSecondaryConstructor() {
+    val typeSpec = TypeSpec.classBuilder("IoException")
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2498,6 +2498,51 @@ class TypeSpecTest {
...
+      |""".trimMargin())
+  }
+
+  @Test fun classExtendsNoPrimaryConstructorButSuperclassParams() {
+    try {
+      TypeSpec.classBuilder("IoException")
+          .superclass(Exception::class)
+          .addSuperclassConstructorParameter("%S", "hey")
+          .addFunction(FunSpec.constructorBuilder().build())
+          .build()
```
vanniktech: Missing a fail

JakeWharton: Bah we need a lambda version of this assertion so I don't have to bother

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2498,6 +2498,52 @@ class TypeSpecTest {
...
+      fail()
+    } catch (e: IllegalArgumentException) {
```
Egorand: nit: `expected`

JakeWharton: We usually only use that name when the value is unused in the `catch` block. Another reason to move to a `assertThrows` helper.

PR #225, opened 2017-09-07
Use a helper for asserting when code throws.

JakeWharton:


src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -37,53 +36,33 @@ class CodeBlockTest {
...
   @Test fun statementBeginningCannotBeIndexed() {
-    try {
+    assertThrows<IllegalArgumentException> {
       CodeBlock.builder().add("%1[", "taco").build()
-      fail()
-    } catch (exp: IllegalArgumentException) {
-      assertThat(exp).hasMessage("%%, %>, %<, %[, %], and %W may not have an index")
-    }
-
+    }.hasMessage("%%, %>, %<, %[, %], and %W may not have an index")
```
swankjesse: Nice

PR #228, opened 2017-09-08
Add type variable support to typealias.

### LINKED ISSUE #227: Support type variables for TypeAliasSpec
Currently can't properly generate:

```kotlin
impl typealias AtomicRef<V> = AtomicReference<V>
```

JakeWharton:
Closes #227.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
@@ -27,10 +27,13 @@ class TypeAliasSpec private constructor(builder: TypeAliasSpec.Builder) {
   val name = builder.name
   val type = builder.type
   val modifiers = builder.modifiers.toImmutableSet()
+  val typeVariables = builder.typeVariables.toImmutableList()
 
   internal fun emit(codeWriter: CodeWriter) {
     codeWriter.emitModifiers(modifiers)
-    codeWriter.emitCode("typealias %L = %T", name, type)
+    codeWriter.emitCode("typealias %L", name)
+    codeWriter.emitTypeVariables(typeVariables)
+    codeWriter.emitCode(" = %T", type)
     codeWriter.emit("\n")
   }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/TypeAliasSpecTest.kt
```diff
@@ -37,6 +37,18 @@ class TypeAliasSpecTest {
     }
   }
 
+  @Test fun typeVariable() {
+    val v = TypeVariableName("V")
+    val typeAliasSpec = TypeAliasSpec
+        .builder("Word", ParameterizedTypeName.get(List::class.asClassName(), v))
```
JakeWharton: It's a bit weird to reference the variable in the supertype first, but we can look at that later...

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -47,14 +50,16 @@ class TypeAliasSpec private constructor(builder: TypeAliasSpec.Builder) {
...
+    builder.modifiers += modifiers
+    builder.typeVariables += typeVariables
     return builder
   }
 
   class Builder internal constructor(
       internal val name: String,
       internal val type: TypeName) {
-    internal var modifiers: MutableSet<KModifier> = mutableSetOf()
+    internal val modifiers = mutableSetOf<KModifier>()
```
vanniktech: indent is off here

JakeWharton: It's not a param

vanniktech: whops

JakeWharton: The problem is that `) {` should be on its own line

JakeWharton: ...for the precise reason you just discovered

vanniktech: definitely agree, want me to fix those?

JakeWharton: I didn't fix here to minimize the diff, but once this is approved and merged a PR to fix it would be great!

JakeWharton: Or fix it and i'll merge and rebase this since no one seems to be around...

PR #229, opened 2017-09-09
Put ) { on it's own line.

vanniktech:


vanniktech:
This should be done in #232

PR #232, opened 2017-09-10
Improve parameter list formatting

### LINKED ISSUE #218: Format multiline constructors (and methods) according to style guide
https://kotlinlang.org/docs/reference/coding-conventions.html

```kotlin
data class Person(
    override val id: Int, 
    override val name: String,
    override val surname: String
)
```

instead of 

```kotlin
data class Person(override val id: Int, override val name: String,
    override val surname: String)
```

as a developer its desirable because it makes testing generated code easier - if only one param changes the error diff is less cluttered.

Egorand:
Fixes #218

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -138,3 +138,23 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
         = builder(name, type.asTypeName(), *modifiers)
   }
 }
+
+internal fun List<ParameterSpec>.emit(
+    codeWriter: CodeWriter,
+    emitBlock: (ParameterSpec) -> Unit = { it.emit(codeWriter) }
+) {
+
+  fun emit(prefix: String, separator: String, suffix: String) {
+    codeWriter.emitCode(prefix)
+    forEachIndexed { index, parameter ->
+      if (index > 0) codeWriter.emitCode(separator)
+      emitBlock(parameter)
+    }
+    codeWriter.emitCode(suffix)
+  }
+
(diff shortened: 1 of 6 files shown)


Egorand:
@AlecStrong I went for a simpler solution here to wrap if there are more than 2 parameters. Would be good to actually wrap long lines only, but that seems to require being able to calculate the parameter list length and tweak formatting on the go - haven't figured out how to do that yet.

AlecStrong:
sounds reasonable to me, thanks!

src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -138,3 +138,23 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
         = builder(name, type.asTypeName(), *modifiers)
   }
 }
+
+internal fun List<ParameterSpec>.emit(
+    codeWriter: CodeWriter,
+    emitBlock: (ParameterSpec) -> Unit = { it.emit(codeWriter) }
+) {
+
+  fun emit(prefix: String, separator: String, suffix: String) {
```
swankjesse: Whoa, cool

swankjesse: Though this might be simpler as just a when clause with cases for 0, 1, 2 and else!

swankjesse: (no need to define prefix and suffix, each case could just be inline.)

Egorand: Makes sense! I pushed a08146c, gonna squash if nothing to change

PR #240, opened 2017-09-19
Improve annotation logic

### LINKED ISSUE #215: Annotation logic needs to change.
They're syntactically like calling a constructor and less of a weird special syntax like Java.

e.g., we should be able to emit:
```kotlin
@Deprecated("Nope", ReplaceWith("Yep"))
```

Egorand:
Closes #215 

- we don't emit member names by default anymore, but it can be enabled via `emitName` param in `addMember()`
- there are cases when specifying parameter names is mandatory: `Foo(vararg val a: Int, val value: Int)`
- reflecting from `Annotation` and `AnnotationMirror` will still emit all names except for `value` - should there be an option to switch it off for the whole annotation?


src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -266,10 +266,10 @@ class TypeSpecTest {
         |import kotlin.collections.Map
         |
         |interface Service {
-        |  @Headers([
+        |  @Headers(
         |      "Accept: application/json",
         |      "User-Agent: foobar"
-        |  ])
+        |  )
```
JakeWharton: This seems like a regression

Egorand: It's not. Since the param is `vararg value: String` it's fine to pass strings vararg-style.

JakeWharton: Do we still have tests which exercise the annotation literal syntax then?

Egorand:
I see, so we could just drop `name` and accept plain `CodeBlock`s - will try it out.

JakeWharton:
Or just a single CodeBlock for the whole body between the parenthesis. I
mean you're already basically doing that for nested annotations with
"%T(%N)".

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
Egorand: Yes, and I added a few as well:

```kotlin
@Test fun valueAndOtherVarargParameterEmitNames() {
  val annotationSpec = AnnotationSpec.builder(ValueAndOtherVarargParam::class)
      .addMember("a", CodeBlock.of("%S", "a"), emitName = true)
      .addMember("a", CodeBlock.of("%S", "b"), emitName = true)
      .addMember("a", CodeBlock.of("%S", "c"), emitName = true)
      .addMember("value", "1", emitName = true)
      .build()

  assertThat(toString(annotationSpec)).isEqualTo("""
    |package com.squareup.tacos
    |
    |import com.squareup.kotlinpoet.AnnotationSpecTest
    |
    |@AnnotationSpecTest.ValueAndOtherVarargParam(
    |    a = [
    |        "a",
    |        "b",
    |        "c"
    |    ],
    |    value = 1
    |)
    |class Taco
    |""".trimMargin())
}
```

Egorand:
True! I'm gonna leave the reflection use-case unchanged, just that the arg names will move under that single `CodeBlock`.

Egorand:
@JakeWharton I reworked the API, the model is a `List<CodeBlock>` now. Reflecting and importing from `AnnotationMirror` were a bit hard to rework, as I said I went with keeping the parameter name emission, to make sure the output is correct in all cases. There's less flexibility now, it's not possible to edit members on a reflected annotation anymore, check deleted test cases.  

The old commit is still in the tree, in case we'd want to get back to the previous solution. I'll squash everything when it's good to merge.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -127,59 +101,66 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
...
     }
 
     fun build() = AnnotationSpec(this)
+
+    companion object {
+
+      /**
+       * Creates a [CodeBlock] with parameter `format` depending on the given `value` object.
+       * Falls back to `"%L"` literal format if the class of the given `value` object is not
+       * supported.
```
swankjesse: not supported –> not known ?

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -131,6 +131,13 @@ class CodeBlock private constructor(
     }
   }
 
+  /**
+   * Returns a copy of the code block with selected format parts replaced, similar to
+   * [java.lang.String.replaceAll].
+   */
+  internal fun replaceAll(oldValue: String, newValue: String) =
+      CodeBlock(formatParts.map { it.replace(oldValue, newValue) }, args)
```
swankjesse: Once this completes, we could have a format part that is empty, or one that doesn’t match its arguments.

This is safe as an internal function but really dangerous in general. Want to add a warning to that effect?

swankjesse:
Kotlin’s syntax is much more consistent. Win!

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Egorand: Nice catch, I'll definitely mention that in the comment.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
Egorand: In fact it's "is not a special case". I can try to rephrase the doc.

Egorand:
@swankjesse true! I'm a little concerned about the black magic that was required to ensure proper formatting in inline vs non-inline cases, but the API looks nicer now. 

Pushed another comment with improved docs, please squash when merging.

PR #245, opened 2017-09-27
Support inline properties

Egorand:
[Inline properties (since 1.1)](https://kotlinlang.org/docs/reference/inline-functions.html#inline-properties)

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
@@ -32,6 +32,27 @@ class PropertySpecTest {
     assertThat(prop.toString()).isEqualTo("val foo: kotlin.String by Delegates.notNull()\n")
   }
 
+  @Test fun inline() {
+    val prop = PropertySpec.varBuilder("foo", String::class)
+        .addModifiers(KModifier.INLINE)
+        .getter(FunSpec.getterBuilder()
+            .addModifiers(KModifier.INLINE)
+            .addStatement("return %S", "foo")
+            .build())
+        .setter(FunSpec.setterBuilder()
+            .addModifiers(KModifier.INLINE)
+            .addParameter("value", String::class)
+            .build())
+        .build()
+    
+    assertThat(prop.toString()).isEqualTo("""
+      |inline var foo: kotlin.String
+      |  inline get() = "foo"
+      |  inline set(value) {
+      |  }
+      |""".trimMargin())
+  }
+
   @Test fun equalsAndHashCode() {
     val type = Int::class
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -32,6 +32,27 @@ class PropertySpecTest {
...
+        .setter(FunSpec.setterBuilder()
+            .addModifiers(KModifier.INLINE)
+            .addParameter("value", String::class)
+            .build())
+        .build()
+    
+    assertThat(prop.toString()).isEqualTo("""
+      |inline var foo: kotlin.String
+      |  inline get() = "foo"
+      |  inline set(value) {
```
JakeWharton: Do we want to skip emitting these when the enclosing property is marked as `inline`?

swankjesse: Yeah. My inclination is to only allow the modifier on `get()` and `set()`, but to promote it from there to the declaration itself if it’s on both `get()` and `set()`. That also avoids the problem of declaring inline on a property that has a backing field.

Egorand: Will it be clear from the API standpoint that you have to mark both getter and setter to have the property marked `inline`? I'd go with allowing it on both, but optimizing emission, in cases when either the property is `inline`, or both accessors are `inline`, as Jake suggested.

In case the backing field is used, marking either of the accessors or the property with `inline` will break compilation.

swankjesse: I think that `inline` on the property itself is syntactic sugar; it’s a more compact way of expressing that the accessors are inline.

Thus far in KotlinPoet we like to make our users specify the non-compact form but we emit the compact form where it applies. That way there’s only one way to do in in the API.

Egorand: Agree! Will submit another PR tomorrow

JakeWharton: Does that mean we prevent you from passing `inline` on the property? That feels really weird and I don't think it's the same as the expression body or primary constructor property behavior which would otherwise require new API.

swankjesse: I think we do prevent you from passing it on the property.

I’m imagining that KotlinPoet is kind of a cleaned-up, normalized AST and I don’t like that `inline` can be defined in multiple places. I like models where there’s exactly one way to represent a value.

PR #246, opened 2017-09-27
Support reified types

### LINKED ISSUE #244: Support parameterized functions.
`FunSpec.builder("foo").addParameter("foo", TypeVariableName.of("T", emptyList(), null)).build()`
makes
```kotlin
fun foo(foo: T) {
}
```

Should it make the following? Is there a way to make the following at all yet?
```kotlin
fun <T> foo(foo: T) {
}
```

Egorand:
Fixes #244


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -50,6 +51,9 @@ class FunSpec private constructor(builder: Builder) {
     require(name != SETTER || parameters.size == 1) {
       "$name must have exactly one parameter"
     }
+    require(INLINE in modifiers || typeVariables.none { it.reified }) {
+      "only type parameters of inline functions can be reified!"
+    }
```
NightlyNexus: oh, that is nice.

swankjesse:
I love this.

PR #248, opened 2017-09-28
Escape packageName components in FileSpec which are also Kotlin keywords

### LINKED ISSUE #247: Check packageName components for keywords
The following will generate an invalid package name because of the `in` keyword in one of its components:

```kotlin
@Test fun keywordInPackageName() {
  val source = FileSpec.builder("com.squareup.in", "California")
      .build()
      |package com.squareup.in
      |
      |""".trimMargin())
}
```

Would it be advisable to automatically adjust the emitted code to include the grave accents around keywords (i.e. ``` package com.squareup.`in` ```)? If not, should we extend the [existing sanity checks in FileSpec.Builder](https://github.com/square/kotlinpoet/blob/4b0e6e2b714bfbee559498c4f14db2c4e6605a41/src/main/java/com/squareup/kotlinpoet/FileSpec.kt#L177) to validate the `packageName` components at initialization time as well, and fail quickly on reserved keywords?

mannodermaus:
Fixes #247

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -165,8 +165,10 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
   }
 
   class Builder internal constructor(
-      internal val packageName: String,
+      packageName: String,
       internal val name: String) {
+    internal val packageName = packageName.split('.')
+        .joinToString(".") { escapeIfKeyword(it) }
     internal val annotations = mutableListOf<AnnotationSpec>()
     internal val comment = CodeBlock.builder()
     internal val memberImports = sortedSetOf<String>()
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -96,8 +96,11 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
 
     codeWriter.pushPackage(packageName)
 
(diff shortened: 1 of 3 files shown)


mannodermaus:
Let me know if `escapeIfKeyword(String)` should be an extension property instead, I went with my gut feeling, looking at the remaining code in `Util.kt` for reference.

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -165,8 +165,10 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
   }
 
   class Builder internal constructor(
-      internal val packageName: String,
+      packageName: String,
       internal val name: String) {
+    internal val packageName = packageName.split('.')
+        .joinToString(".") { escapeIfKeyword(it) }
```
JakeWharton: I think we only want to do this inside of `writeTo`, since currently we'd expose this escaped version which might break someone using it for something else.

mannodermaus:
Deferred the actual replacement to `FileSpec#emit()`.

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -133,6 +133,8 @@ internal fun stringLiteralWithQuotes(value: String): String {
   }
 }
 
+internal fun escapeIfKeyword(value: String): String = if (value.isKeyword) "`$value`" else value
```
Egorand: nit: return type can be omitted

mannodermaus: True. Updated!

PR #250, opened 2017-10-02
Improve inline properties feature

Egorand:
Based on the discussion on #245

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -38,9 +38,12 @@ class PropertySpec private constructor(builder: Builder) {
       implicitModifiers: Set<KModifier>,
       withInitializer: Boolean = true,
       inline: Boolean = false) {
+    val isInlineProperty = getter?.modifiers?.contains(KModifier.INLINE) ?: false &&
+        setter?.modifiers?.contains(KModifier.INLINE) ?: false
+    val propertyModifiers = if (isInlineProperty) modifiers + KModifier.INLINE else modifiers
     codeWriter.emitKdoc(kdoc)
     codeWriter.emitAnnotations(annotations, false)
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -149,12 +156,22 @@ class PropertySpec private constructor(builder: Builder) {
...
     }
 
     fun setter(setter: FunSpec) = apply {
       require(setter.name == SETTER) { "${setter.name} is not a setter" }
       check(this.setter == null) { "setter was already set" }
       this.setter = setter
+      checkPromoteInlineToProperty()
+    }
+
+    private fun checkPromoteInlineToProperty() {
```
swankjesse: Can you do this at render time instead?

swankjesse: Ie. keep `modifiers` as-is as a property of this class, but add `inline` when printing it? That way if anyone uses KotlinPoet as a basic AST they get `inline` where they expect it

JakeWharton: And (more importantly in my opinion) they don't see `inline` where they didn't specify it.

Egorand: Agree, good point. Force-pushed an updated version.

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -52,14 +55,19 @@ class PropertySpec private constructor(builder: Builder) {
       codeWriter.emitCode("%[%L%]", initializer)
     }
     if (!inline) codeWriter.emit("\n")
+    val implicitAccessorModifiers = if (isInlineProperty) {
```
swankjesse: Yeah, I really like this

swankjesse:
I really like where this landed. Thank you.

PR #254, opened 2017-10-10
Add support for renamed imports

### LINKED ISSUE #253: Add support for renamed imports
Should be able to generate the following:

```kotlin
import java.lang.String as JString
import kotlin.String as KString

val a: JString = JString("a")
val b: KString = "b"
```

Egorand:
Closes #253


src/main/java/com/squareup/kotlinpoet/ImportSpec.kt
```diff
@@ -0,0 +1,67 @@
...
+   */
+  fun renameTo(name: String) = ImportSpec(className, memberName, name)
+
+  override fun toString() = import
+
+  override fun compareTo(other: ImportSpec) = import.compareTo(other.import)
+
+  companion object {
+
+    fun forType(typeName: String) = ImportSpec(typeName)
```
vanniktech: JvmStatic for Java users?

JakeWharton:
I do not like this approach. Import aliasing shouldn't affect the types you specify for the various constructs. It should only affect the rendered output of a `FileSpec` and not the rest of the model.

src/main/java/com/squareup/kotlinpoet/ImportSpec.kt
Egorand: Yes, possible. I'll probably remove this class though.

Egorand:
@JakeWharton I pushed an alternative solution that's limited to `FileSpec`, guess that's what you meant.

JakeWharton:
I don't think the specs should know or care about type aliases. Here's the code that I want to write:
```kotlin
  @Test fun aliasedImports() {
        .addAliasedImport(java.lang.String::class.java, "JString")
        .addAliasedImport(String::class, "KString")
        .addProperty(PropertySpec.builder("a", java.lang.String::class.java)
            .initializer("%T(%S)", java.lang.String::class.java, "a")
            .build())
        .addProperty(PropertySpec.builder("b", String::class)
            .initializer("%S", "b")
            .build())
        .build()
      |package com.squareup.tacos
      |
      |import java.lang.String as JString
      |import kotlin.String as KString
      |
      |val a: JString = JString("a")
      |
      |val b: KString = "b"
      |""".trimMargin())
  }
```

Egorand:
Oh, I see, that looks nice. I'll try to get that to work.

JakeWharton:
Yeah. It basically becomes an implementation detail of CodeWriter to
rewrite the types.

src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -324,6 +324,42 @@ class FileSpecTest {
...
+
+  @Test fun enumAliasedImport() {
+    val minsAlias = "MINS"
+    val source = FileSpec.builder("com.squareup.tacos", "Taco")
+        .addAliasedImport(TimeUnit::class.asClassName(), "MINUTES", minsAlias)
+        .build()
+    assertThat(source.toString()).isEqualTo("""
+      |package com.squareup.tacos
+      |
+      |import java.util.concurrent.TimeUnit.MINUTES as MINS
```
swankjesse: For follow up - test that this gets folded in when it’s referenced?

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -37,7 +37,7 @@ private fun extractMemberName(part: String): String {
 internal class CodeWriter constructor(
     out: Appendable,
     private val indent: String = "  ",
-    private val memberImports: Set<String> = emptySet(),
+    private val memberImports: Map<String, String?> = emptyMap(),
```
swankjesse: should this be `String?` with null as the sentinel? Or say, a special string as we do with `NO_PACKAGE` ?

PR #258, opened 2017-10-13
Fix import aliasing for enum members

Egorand:
Followup to #254 

Also adds the `NO_ALIAS` sentinel


JakeWharton:
This null-object pattern feels out of place in Kotlin to me here and for the package. `null` is a perfectly good sentinel value in Kotlin because you're forced to handle it. If we're aiming to improve readability then changing `Map<String, String?>` to a `List<Import>` with `data class Import(qualifiedName: String, alias: String?)` seems like a better improvement.

Egorand:
@JakeWharton agree, makes sense. @swankjesse what do you say?

Egorand:
I added the `Import` class since I really like the idea

PR #260, opened 2017-10-15
Add LambdaSpec to simplify constructing lambda expressions

### LINKED ISSUE #257: Generating lambdas with parameters
Is there some function for generating `{ foo, bar -> ... }` similar to `beginControlFlow`? If not what should it be?

Egorand:
Closes #257 

Not sure if this kind of solution is too much for the problem, but it seemed like a straightforward one. Feedback is super welcome.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/LambdaSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/LambdaSpec.kt
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


JakeWharton:
I'm extremely hesitant to add a `*Spec` for this. Thus far they've only been for structure, not contents. We should be looking at trying to solve the more general problem of the different types of blocks and how to indent them. Examples that come to mind are `when` cases, `if` or `try` used as an expression, lambdas, etc.

Egorand:
Yeah, I see. Indeed, there should be a way to leave the current API intact and have the internals figure out the correct indentation strategy for specific types of statements. Gonna close this for now.

PR #261, opened 2017-10-16
Fix formatting of property initializers and multiline strings

### LINKED ISSUE #259: Improve formatting of property initializers
There are different formatting rules for property initializers, few examples below:

```kotlin
val a = "Hello !"
    .trim()
    .toLong()

val b = """
  a
  b
  c
  """

val c = { x: Int, y: Int ->
  x + y
}
```

Current implementation wraps the entire initializer in a statement, applying double indentation to every new line, hence the last example will be formatted as:

```kotlin
val c = { x: Int, y: Int ->
      x + y
    }
```

Egorand:
Fixes #259

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -88,7 +88,9 @@ internal fun characterLiteralWithoutSingleQuotes(c: Char) = when {
 internal fun stringLiteralWithQuotes(value: String, indent: String = ""): String {
   if (value.contains("\n")) {
     val result = StringBuilder(value.length + 32)
-    result.append("\"\"\"\n" + indent + "|")
+    result.append("\"\"\"\n")
+    result.append(indent)
+    result.append('|')
     var i = 0
     while (i < value.length) {
       val c = value[i]
@@ -98,7 +100,9 @@ internal fun stringLiteralWithQuotes(value: String, indent: String = ""): String
         i += 2
       } else if (c == '\n') {
         // Add a '|' after newlines. This pipe will be removed by trimMargin().
-        result.append("\n" + indent + "|")
+        result.append('\n')
+        result.append(indent)
+        result.append('|')
       } else {
         result.append(c)
       }
@@ -106,7 +110,10 @@ internal fun stringLiteralWithQuotes(value: String, indent: String = ""): String
     }
     // If the last-emitted character wasn't a margin '|', add a blank line. This will get removed
     // by trimMargin().
-    if (!value.endsWith("\n")) result.append("\n" + indent)
+    if (!value.endsWith("\n")) {
+      result.append('\n')
+      result.append(indent)
+    }
     result.append("\"\"\".trimMargin()")
     return result.toString()
   } else {
--- a/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -85,10 +85,10 @@ internal fun characterLiteralWithoutSingleQuotes(c: Char) = when {
 }
 
 /** Returns the string literal representing `value`, including wrapping double quotes.  */
-internal fun stringLiteralWithQuotes(value: String): String {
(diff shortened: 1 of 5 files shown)


src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -85,10 +85,10 @@ internal fun characterLiteralWithoutSingleQuotes(c: Char) = when {
 }
 
 /** Returns the string literal representing `value`, including wrapping double quotes.  */
-internal fun stringLiteralWithQuotes(value: String): String {
+internal fun stringLiteralWithQuotes(value: String, indent: String = ""): String {
   if (value.contains("\n")) {
     val result = StringBuilder(value.length + 32)
-    result.append("\"\"\"\n|")
+    result.append("\"\"\"\n" + indent + "|")
```
JakeWharton: String concatenation shouldn't occur in calls to `append` on a `StringBuilder`. Use multiple calls to `append`

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -98,15 +98,15 @@ internal fun stringLiteralWithQuotes(value: String): String {
         i += 2
       } else if (c == '\n') {
         // Add a '|' after newlines. This pipe will be removed by trimMargin().
-        result.append("\n|")
+        result.append("\n" + indent + "|")
```
JakeWharton: String concatenation shouldn't occur in calls to `append` on a `StringBuilder`. Use multiple calls to `append`

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -98,15 +98,15 @@ internal fun stringLiteralWithQuotes(value: String): String {
...
       } else {
         result.append(c)
       }
       i++
     }
     // If the last-emitted character wasn't a margin '|', add a blank line. This will get removed
     // by trimMargin().
-    if (!value.endsWith("\n")) result.append("\n")
+    if (!value.endsWith("\n")) result.append("\n" + indent)
```
JakeWharton: String concatenation shouldn't occur in calls to `append` on a `StringBuilder`. Use multiple calls to `append`

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -200,14 +200,14 @@ class KotlinPoetTest {
   @Test fun rawStrings() {
     val source = FileSpec.get(tacosPackage, TypeSpec.classBuilder("Taco")
         .addFunction(FunSpec.builder("strings")
-            .addStatement("val a = %S", "\"\n")
-            .addStatement("val b = %S", "a\"\"\"b\"\"\"\"\"\"c\n")
-            .addStatement("val c = %S", """
+            .addCode("val a = %S\n", "\"\n")
```
JakeWharton: What's the behavior if this were to remain `addStatement`? Forcing people to change the method they're calling isn't appealing.

src/main/java/com/squareup/kotlinpoet/Util.kt
Egorand: Will do. Didn't want to go with `\n$indent|` as it's unreadable.

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
Egorand: Inside a statement, every new line gets double-indented, so in this case the formatting will be weird. 

Since formatting rules for multi-line expressions in Kotlin vary, I feel like `addStatement()` doesn't make much sense as is and has to be reworked.

JakeWharton:
The need to use `addCode` here is unintuitive and means that you can't simply pass along user-input strings without either always using `addCode` or checking/escaping newlines. I'd prefer `addStatement` to Just Work™ when given a multiline string.

Egorand:
Same as for #260, closing this in favor of a solution that doesn't break current API.

PR #263, opened 2017-10-24
Escape name in FunSpec which is also a Kotlin keyword

### LINKED ISSUE #262: Provide some escape hatch to allow generating funs using Java keywords as name
While Kotlin seems perfectly fine to use Java keywords in method names (as long as you escape them in backticks) it seems `FunSpec` is not.

I'm currently migrating an XML-based DSL to Kotlin (basically a 1:1 conversion) which makes use of Java keywords.

Would be lovely if I could somehow disable the check for reserved keyword.

helpermethod:
Fixes #262.


helpermethod:
Fixes #262.

PR #265, opened 2017-10-30
Upgrade all dependencies

Egorand:


src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -38,31 +38,31 @@ class CodeBlockTest {
   @Test fun indentCannotBeIndexed() {
     assertThrows<IllegalArgumentException> {
       CodeBlock.builder().add("%1>", "taco").build()
-    }.hasMessage("%%, %>, %<, %[, %], and %W may not have an index")
```
Egorand: `hasMessage()` is deprecated

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -38,31 +38,31 @@ class CodeBlockTest {
...
+    }.hasMessageThat().isEqualTo("%%, %>, %<, %[, %], and %W may not have an index")
```
Egorand: `hasMessage()` is deprecated

pom.xml
```diff
@@ -86,19 +86,19 @@
...
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>1.10.16</version>
+      <version>2.11.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jdt.core.compiler</groupId>
```
JakeWharton: Is this still used?

Egorand: Not sure, can check

Egorand: There's a single test in `FunSpecTest` that uses Mockito, maybe there's a way to rewrite it and get rid of the dependency

PR #266, opened 2017-10-30
Remove explicit Java types from Util.

NightlyNexus:


src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -15,54 +15,35 @@
...
-  for ((key, value) in this) {
-    if (value.isEmpty()) continue
-    result.put(key, value.toImmutableList())
-  }
-  return Collections.unmodifiableMap(result)
-}
-
-internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V>
-    = Collections.unmodifiableMap(LinkedHashMap(this))
+internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> = LinkedHashMap(this)
```
Egorand: That's clever! Since there are no mutators on `kotlin.collections.Map`, no need to bother with `unmodifiableMap()`

JakeWharton: It was for Java users

Egorand: Oh yeah, right

Egorand: But that's `internal` anyway?

JakeWharton: This is used for exposed types

NightlyNexus: Oh, right. Will revert, thanks!

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -15,28 +15,14 @@
...
-import java.util.LinkedHashMap
-import java.util.LinkedHashSet
 
 internal object NullAppendable : Appendable {
   override fun append(charSequence: CharSequence) = this
   override fun append(charSequence: CharSequence, start: Int, end: Int) = this
   override fun append(c: Char) = this
 }
 
-internal fun <K, V> Map<K, List<V>>.toImmutableMultimap(): Map<K, List<V>> {
```
NightlyNexus: i can revert here if it's going to be used later.

src/main/java/com/squareup/kotlinpoet/Util.kt
Egorand: Is this a valid use case for a multiplatform implementation?

- `kotlinpoet-common`

```kotlin
expect internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V>
```

- `kotlinpoet`

```kotlin
actual internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> = 
    LinkedHashMap(this)
```

- `kotlinpoet-jvm`

```kotlin
actual internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> = 
    Collections.unmodifiableMap(LinkedHashMap(this))
```

swankjesse: Yes. But the multiplatform stuff is probably not stable enough to bother. When it's more mature we can.

JakeWharton: Kotlin and Java are both JRE targets. It'd be weird to have separate artifacts for them.

PR #267, opened 2017-10-31
Correct statement formatting

### LINKED ISSUE #259: Improve formatting of property initializers

Egorand:
Closes #259

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -57,7 +57,7 @@ internal class CodeWriter constructor(
    * line of a statement is indented normally and subsequent wrapped lines are double-indented. This
    * is -1 when the currently-written line isn't part of a statement.
    */
-  var statementLine = -1
+  var statementFormatter: StatementFormatter? = null
 
   init {
     for ((className, _) in memberImports) {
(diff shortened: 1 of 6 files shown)


Egorand:
Wanna add more documentation but would be great to have some feedback on the approach

JakeWharton:
I'll try to look at this soon, but it's definitely going to miss the next release and I'm scrambling trying to finish slides so it might slip to early next week.

swankjesse:
I’m a little lost on the motivation on this. The problem is that we’re indenting multiline strings too much?

Egorand:
In general, our indentation logic for statements is dumb, as I believe Java's indenting logic is dumb as well. Kotlin apparently has more indentation rules.

swankjesse:
I'd like to stall this until we update to the new style guide including 4-space indents everywhere.

That’s at least 3 parts:

 * Fix Square’s java-code-styles to match the Android spec.
 * Fix KotlinPoet’s own code to follow the guide
 * Fix emitted code to follow the guide.

The new rules are not my favorite, but I’d really like to just leave the world of manually wrapping and formatting in the past.

JakeWharton:
For what it's worth, JetBrains own code base is 2 space so some level of
civil disobedience is allowed.

Egorand:
Shall we switch to printing 4 spaces?

swankjesse:
Yep. We want to move as quickly as possible to the way that Go programmers work, where humans don’t get to control the whitespace.

swankjesse:
Close this?

Egorand:
I'd agree that it's a minor bug, shall we close both the PR and the issue?

swankjesse:
Done!

PR #268, opened 2017-10-31
Support lambda extensions properly in output.

JakeWharton:
And expose the things inside `LambdaTypeName`.


JakeWharton:
The tests probably make no sense... I just copied and changed the receiver type

PR #270, opened 2017-11-03
Support named LambdaTypeName parameters

### LINKED ISSUE #269: LambdaTypeName parameter names
```kotlin
(name: String) -> Unit
```

This also works, but is weird
```kotlin
fun ((name: String) -> Unit).whatever() = Unit
```

Egorand:
Fixes #269


src/main/java/com/squareup/kotlinpoet/Parameter.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.reflect.Type
+import kotlin.reflect.KClass
+
+data class Parameter(val name: String? = null, val type: TypeName) {
+
+  constructor(name: String, type: KClass<*>) : this(name, type.asTypeName())
+
+  constructor(name: String, type: Type) : this(name, type.asTypeName())
+
+  fun toCodeBlock() = if (name != null) {
```
JakeWharton: `internal`?

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -55,18 +55,33 @@ class LambdaTypeName internal constructor(
...
+    /*
+     * Returns a lambda type with `returnType` and parameters listed in `parameters` and
+     * `namedParameters`.
+     */
     @JvmStatic fun get(
         receiver: TypeName? = null,
         parameters: List<TypeName> = emptyList(),
-        returnType: TypeName)
-        = LambdaTypeName(receiver, parameters, returnType)
+        namedParameters: List<Parameter> = emptyList(),
```
JakeWharton: It's weird to have two lists. I can't mix and match to get like `x: Long, y, z: Long`.

Egorand: Didn't wanna break existing API, and adding a new factory method with `parameters: List<Parameter>` wouldn't compile due to type erasure. If breaking is fine - I'd just get rid of `parameters: List<TypeName>`.

JakeWharton: Hmm we can use `@JvmName` to defeat the erasure 🙈

Egorand: Oh yeah, let me try that. Wonder how that would work for JS, no erasure - no problem?

Egorand: Well, I couldn't come up with a meaningful value for `@JvmName`, so I went the following way instead: the version with `parameters` and `namedParameters` stays, but is deprecated in favor of the version with `parameters: List<Parameter>`. We can remove the former when going 1.0. Does this make sense?

swankjesse: I’m not too concerned with backwards-compatibility at this point.

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -55,18 +55,51 @@ class LambdaTypeName internal constructor(
   }
 
   companion object {
-    /** Returns a lambda type with `returnType` and parameters of listed in `parameters`. */
+    /*
```
swankjesse: `/**` ?

src/main/java/com/squareup/kotlinpoet/Parameter.kt
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.squareup.kotlinpoet
+
+import java.lang.reflect.Type
+import kotlin.reflect.KClass
+
+data class Parameter(val name: String? = null, val type: TypeName) {
```
swankjesse: This competes with an existing class, `ParameterSpec`. I’m anxious about having both types that mean almost the exact same thing. Would it be simpler to just use `ParameterSpec` in `LambdaTypeName` ?

src/main/java/com/squareup/kotlinpoet/Parameter.kt
```diff
@@ -0,0 +1,42 @@
...
+
+  constructor(name: String, type: KClass<*>) : this(name, type.asTypeName())
+
+  constructor(name: String, type: Type) : this(name, type.asTypeName())
+
+  internal fun toCodeBlock() = if (name != null) {
```
swankjesse: off topic: would it be better to do this kind of expression without braces?
```
internal fun toCodeBlock() = if (name != null)
    CodeBlock.of("%L: %T", name, type) else
    CodeBlock.of("%T", type)
```

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -324,6 +325,51 @@ class KotlinPoetTest {
...
+  }
+
+  @Test fun extensionFunctionLambdaWithMultipleParams() {
+    val source = FileSpec.builder(tacosPackage, "Taco")
+        .addFunction(FunSpec.builder("whatever")
+            .returns(Unit::class)
+            .receiver(LambdaTypeName.get(
+                parameters = listOf(
+                    Parameter("name", String::class),
+                    Companion.ofType(Int::class),
```
swankjesse: off topic: referencing `Companion` like this seems out of place. Idiomatic?

src/main/java/com/squareup/kotlinpoet/Parameter.kt
Egorand: That's a good point. We'll need only name and type, should we just ignore any other params set on the `ParameterSpec`?

src/main/java/com/squareup/kotlinpoet/Parameter.kt
Egorand: Yeah, I think that'd be fine with the guidelines

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
Egorand: Nope, there seems to be a problem with auto-complete and I didn't notice it, gonna fix.

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
Egorand: Agree. So would a single factory method with `parameters: List<ParameterSpec>` be enough, given that we'll have versions with `vararg parameters: ParameterSpec` and `vararg parameters: TypeName`?

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
Egorand: Right

src/main/java/com/squareup/kotlinpoet/Parameter.kt
Egorand: Hmm, the problem is that `name` is mandatory for `ParameterSpec`, and optional for `LambdaTypeName`. Will switching from `name: String` to `name: String?` on `ParameterSpec` be justified for that purpose? Then we could have `ParameterSpec.ofType()` to create an unnamed parameter.

JakeWharton: That sounds a bit weird.

swankjesse: How about we permit an empty string as the name in those cases, and require a factory method to get instances like that:

```
  ParameterSpec string = ParameterSpec.unnamed(String::class)
```

Egorand: I'll give it a try

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -366,7 +365,11 @@ class KotlinPoetTest {
       |import kotlin.String
       |import kotlin.Unit
       |
-      |fun ((name: String, Int, age: Long) -> Unit).whatever(): Unit = Unit
+      |fun ((
```
Egorand: Formatting here is weird due to the fact that we currently wrap parameter lists with 2 or more elements. I'll open an issue to improve it.

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -17,7 +17,7 @@ package com.squareup.kotlinpoet
 
 class LambdaTypeName internal constructor(
     val receiver: TypeName? = null,
-    parameters: List<Parameter> = emptyList(),
+    parameters: List<ParameterSpec> = emptyList(),
```
Egorand: Question: should we have checks to verify that all `ParameterSpec`s here have only the name and the type and nothing else?

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -324,6 +324,55 @@ class KotlinPoetTest {
...
+      |import kotlin.Int
+      |import kotlin.Long
+      |import kotlin.String
+      |import kotlin.Unit
+      |
+      |fun ((
+      |    name: String,
+      |    Int,
+      |    age: Long
+      |) -> Unit).whatever(): Unit = Unit
```
Egorand: Formatting here is weird as we currently wrap parameter lists with 2 or more elements. I'll open an issue to improve it.

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
JakeWharton: We should check what's supported. I have a feeling annotations might work...

Egorand: Really? Let me see

Egorand: `Unsupported [annotation on parameter in function type]`

Egorand: Modifiers and default values are unsupported as well, hence just name and type

src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
JakeWharton: Yeah we need a smarter CodeWriter which potentially buffers elements so that they can be rewritten when a line break is required.

Egorand: #274

Issue #273, opened 2017-11-06
Track the Android Kotlin style guide


swankjesse:
We should track this both for generated code and also within KotlinPoet itself.
https://android.github.io/kotlin-guides/style.html

Here’s one guideline I found interesting:
 * Expression functions should not wrap to two lines. If an expression function grows to require wrapping, use a normal function body, a return declaration, and normal expression wrapping rules instead.

Egorand:
The style guide has a strong opinion on wildcard imports:

> Wildcard imports (of any type) are **not allowed**.

Should we disallow them in `addStaticImport()`?

swankjesse:
Yeah!

swankjesse:
We should attempt to be consistent with the style guide in code we emit. I don't think it's useful or feasible to strictly enforce the style guide.

For wildcard imports, we simply don’t implement them.

Egorand:
I think we're doing pretty good on adhering to the style guide, formatting issues should go into separate tickets.

PR #275, opened 2017-11-07
Disallow wildcard imports

Egorand:


Egorand:
@swankjesse oops, missed your last comment on #273. If we allow wildcards but do nothing about them it might seem like the library is broken. On a second thought, we might just leave it as is, since that behavior is already implemented.

swankjesse:
Definitely like this. I primarily want us to make styleguide obedient code easy. But it isn’t worth writing a lot of code just to enforce the style guide.

Issue #256, opened 2017-10-11
Lambda parameters dont line wrap


saik0:
```kotlin
val barParams = ParameterSpec.builder("f",
        LambdaTypeName.get(parameters = generateSequence { Unit::class.asTypeName() }.take(20).toList(),
                           returnType = Unit::class.asTypeName()))
        .build()

val file = FileSpec.builder("", "Foo")
        .addFunction(FunSpec.builder("bar")
                .addParameter(barParams)
                .build())
        .build()

file.writeTo(System.out)
```

outputs:
```kotlin
fun bar(f: (Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit, Unit) -> Unit) {
}
```

JakeWharton:
This should trigger one-per-line parameters:
```kotlin
fun bar(
    f: ...
) {
}
```
but I don't think we'll worry about wrapping the actual lambda signature.

saik0:
Fair enough

Egorand:
Currently parameter wrapping depends on the number of parameters and not on the length of the code, which is something we can improve. The difficult part is figuring out the length of the parameter code before it's emitted.

swankjesse:
Fixed.

```
package com.squareup.tacos

import kotlin.Unit

fun bar(f: (
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit,
    Unit
) -> Unit) {
}
```

PR #281, opened 2017-11-14
Smarter wrapping logic for long parameter lists

### LINKED ISSUE #274: Improve wrapping logic of parameter lists
Currently we wrap based on the number of parameters, not on the length:

```kotlin
val file = FileSpec.builder(tacosPackage, "Taco")
    .addFunction(FunSpec.builder("test")
        .addParameter("a", Int::class)
        .addParameter("b", Int::class)
        .addParameter("c", Int::class)
        .addStatement("return a + b + c")
        .build())
    .build()
```

prints this:

```kotlin
fun test(
    a: Int,
    b: Int,
    c: Int
) = a + b + c
```

when it could be this:

```kotlin
fun test(a: Int, b: Int, c: Int) = a + b + c
```

Egorand:
Fixes #274

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
@@ -17,6 +17,7 @@ package com.squareup.kotlinpoet
 
 import com.google.common.truth.Truth.assertThat
 import org.junit.Test
+import java.io.Serializable
 
 class KotlinPoetTest {
   private val tacosPackage = "com.squareup.tacos"
@@ -107,11 +108,7 @@ class KotlinPoetTest {
         |import kotlin.Boolean
         |import kotlin.String
         |
-        |class Taco(
-        |        val cheese: String,
-        |        var cilantro: String,
-        |        lettuce: String
-        |) {
+        |class Taco(val cheese: String, var cilantro: String, lettuce: String) {
         |    val lettuce: String = lettuce.trim()
         |
         |    val onion: Boolean = true
@@ -365,11 +362,7 @@ class KotlinPoetTest {
       |import kotlin.String
       |import kotlin.Unit
       |
-      |fun ((
-      |        name: String,
-      |        Int,
-      |        age: Long
-      |) -> Unit).whatever(): Unit = Unit
+      |fun ((name: String, Int, age: Long) -> Unit).whatever(): Unit = Unit
       |""".trimMargin())
   }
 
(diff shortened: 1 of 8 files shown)


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -552,4 +545,50 @@ class KotlinPoetTest {
...
+                    ParameterSpec.unnamed(Serializable::class),
+                    ParameterSpec.unnamed(Appendable::class),
+                    ParameterSpec.unnamed(Cloneable::class)),
+                returnType = Unit::class.asTypeName()))
+            .build())
+        .addParameter("i", Int::class)
+        .addStatement("return %T", Unit::class)
+        .build()
+    assertThat(source.toString()).isEqualTo("""
+      |fun veryLongFunctionName(
```
Egorand: Not sure what the style guide says about that, but having nested wrapping felt like an overkill. Hence, lambda types as function parameters are always single-line, even if they wrap.

Egorand: This behavior should also be consistent with the discussion on #256: we don't wrap the lambda parameters, but we do wrap the function parameters.

Egorand:
I think I'll update this PR to cover #291 as well. Any feedback here so far?

JakeWharton:
I certainly like the results. When do you emit a parameter spec with wrapping=False?

Egorand:
When you've got a parameterized `LambdaTypeName` as the type of the param:

```kotlin
fun veryLongFunctionName(
        veryLongParameterName: (java.io.Serializable, java.lang.Appendable, kotlin.Cloneable) -> kotlin.Unit,
        i: kotlin.Int
) = kotlin.Unit
```

JakeWharton:
Ah, nice. I think I'd prefer if we flipped the boolean so that it wraps by default and we opt-out only in the lambda special case.

Egorand:
Can do

Egorand:
Fixes #291

gilgoldzweig:
Any updates on this request?

Egorand:
@gilgoldzweig we'll most likely go with a solution for #532, which should handle this use-case as well.

gilgoldzweig:
Cool, is there any way to make it work now?

Egorand:
There's no workaround unfortunately, but if your function/constructor has more than 2 parameters we'll auto-wrap it.

serranojoanna2025-commits:
Pull Requests

PR #282, opened 2017-11-15
Replace java.util.EnumSet.copyOf() with own implementation

Egorand:


src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -131,9 +129,10 @@ internal class CodeWriter constructor(
    */
   fun emitModifiers(
       modifiers: Set<KModifier>,
-      implicitModifiers: Set<KModifier> = emptySet()) {
+      implicitModifiers: Set<KModifier> = emptySet()
+  ) {
     if (modifiers.isEmpty()) return
-    for (modifier in EnumSet.copyOf(modifiers)) {
+    for (modifier in modifiers.toEnumSet()) {
```
JakeWharton: This was being used for sorting. The new version doesn't seem to be explicitly sorting. Are we just getting lucky?

Egorand: The new implementation consumes enum values in order and picks the ones present in `modifiers` - pretty much what `EnumSet.copyOf()` does. There are a few test cases that would've caught incorrect ordering.

src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -23,14 +23,17 @@ internal object NullAppendable : Appendable {
...
+internal fun <T> Collection<T>.toImmutableList(): List<T> =
+    Collections.unmodifiableList(ArrayList(this))
 
-internal fun <T> Collection<T>.toImmutableSet(): Set<T>
-    = Collections.unmodifiableSet(LinkedHashSet(this))
+internal fun <T> Collection<T>.toImmutableSet(): Set<T> =
+    Collections.unmodifiableSet(LinkedHashSet(this))
+
+internal inline fun <reified T : Enum<T>> Collection<T>.toEnumSet(): Set<T> =
+    enumValues<T>().filter(this::contains).toSet()
```
NightlyNexus: could be `enumValues<T>().filterTo(mutableSetOf(), this::contains)` to avoid the extra ArrayList.

Egorand: Nice, I'll use it!

PR #283, opened 2017-11-20
Increase indent to 4 spaces

Egorand:
As discussed in #267


src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -284,3 +284,5 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     @JvmStatic fun builder(packageName: String, fileName: String) = Builder(packageName, fileName)
   }
 }
+
+internal val DEFAULT_INDENT = "    "
```
NightlyNexus: `const`?

Egorand: Oh yeah, nice!

PR #285, opened 2017-11-23
Support suspending lambda types

### LINKED ISSUE #284: Modifiers on lambda types
```kotlin
fun foo(bar: suspend (Foo) -> Bar)
```

Egorand:
Fixes #284

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
@@ -20,6 +20,7 @@ class LambdaTypeName internal constructor(
     parameters: List<ParameterSpec> = emptyList(),
     val returnType: TypeName = UNIT,
     nullable: Boolean = false,
+    val suspending: Boolean = false,
     annotations: List<AnnotationSpec> = emptyList()
 ) : TypeName(nullable, annotations) {
   val parameters = parameters.toImmutableList()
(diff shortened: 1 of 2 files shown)


Egorand:
One reason I didn't go this way is that the factory methods on `LambdaTypeName` are getting too big. Two `vararg` params are not allowed, so you'd probably have to pass a `setOf<KModifier>()`, which is inconvenient. I thought about adding a `Builder`, would that be justified? Then we could have `addModifiers()` and for now throw on anything other than `SUSPEND`.

swankjesse:
Could do like Nullable?

```
val type = LambdaTypeName.get(...).asSuspending()
```

Egorand:
Yeah, another option, not compatible with multiple modifiers though. I can replace the new factory method param with `asSuspending()` for now, and we can add smth like `withModifiers()` later if we need to and deprecate `asSuspending()` - what would you say?

swankjesse:
Yeah I think that's great.

Egorand:
Done

src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
```diff
@@ -20,6 +20,7 @@ class LambdaTypeName internal constructor(
     parameters: List<ParameterSpec> = emptyList(),
     val returnType: TypeName = UNIT,
     nullable: Boolean = false,
+    val suspending: Boolean = false,
```
swankjesse: Because this isn’t a `private` or `internal` val it’s public API right? I think that’s okay

Egorand: It's API, yes

Issue #131, opened 2017-06-16
Support KotlinJS external classes


marciodel:
In KotlinJS, declaring a class as external makes all members implicitly external.

Currently the implicit modifiers set for members are not dynamic.

JakeWharton:
I'm not sure we even need implicit modifiers like we do in JavaPoet. I
think the whole concept can be removed from the library.

JakeWharton:
Nevermind. I was thinking of something different.

We can probably do this, although implicit modifiers are an optimization that isn't strictly needed for correctness, just for cleaning up the generated code.

marciodel:
Nice!

For KotlinJs the implicit 'external' it is about correctness. Only top-level declaration can have 'external' modifier and it propagates for the inner declarations.

In fact I've modified the code to implement it. It is working fine. I intend to create a PR, ASAP.

cypressious:
Currently you also can't do `external` top-level properties.

Egorand:
Fixed with #287

PR #287, opened 2017-11-27
Improve external support

cypressious:
- Functions in external classes mustn't have bodies
- Support external properties

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -17,6 +17,7 @@ package com.squareup.kotlinpoet
 
 import com.squareup.kotlinpoet.KModifier.COMPANION
 import com.squareup.kotlinpoet.KModifier.EXPECT
+import com.squareup.kotlinpoet.KModifier.EXTERNAL
 import com.squareup.kotlinpoet.KModifier.PUBLIC
 import java.lang.reflect.Type
 import kotlin.reflect.KClass
@@ -47,13 +48,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val funSpecs = builder.funSpecs.toImmutableList()
   val typeSpecs = builder.typeSpecs.toImmutableList()
 
-  private val implicitPropertyModifiers = kind.implicitPropertyModifiers + when {
-    KModifier.EXPECT in modifiers -> setOf(KModifier.EXPECT)
-    else -> emptySet()
-  }
-
-  private val implicitFunctionModifiers = kind.implicitFunctionModifiers + when {
-    KModifier.EXPECT in modifiers -> setOf(KModifier.EXPECT)
+  private val implicitMemberModifiers = kind.implicitFunctionModifiers + when {
+    EXPECT in modifiers -> setOf(EXPECT)
+    EXTERNAL in modifiers -> setOf(EXTERNAL)
     else -> emptySet()
   }
 
@@ -74,7 +71,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     return builder
   }
 
-  internal fun emit(codeWriter: CodeWriter, enumName: String?) {
+  internal fun emit(codeWriter: CodeWriter, enumName: String?, isNestedExternal: Boolean = false) {
     // Nested classes interrupt wrapped line indentation. Stash the current wrapping state and put
     // it back afterwards when this type is complete.
     val previousStatementLine = codeWriter.statementLine
@@ -121,7 +118,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       } else {
         codeWriter.emitKdoc(kdoc)
         codeWriter.emitAnnotations(annotations, false)
-        codeWriter.emitModifiers(modifiers, setOf(PUBLIC))
+        codeWriter.emitModifiers(modifiers, if (isNestedExternal) setOf(PUBLIC, EXTERNAL) else setOf(PUBLIC))
         codeWriter.emit(kind.declarationKeyword)
         if (name != null) {
           codeWriter.emitCode(" %L", name)
@@ -210,7 +207,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
           continue
         }
         if (!firstMember) codeWriter.emit("\n")
-        propertySpec.emit(codeWriter, implicitPropertyModifiers)
+        propertySpec.emit(codeWriter, implicitMemberModifiers)
         firstMember = false
       }
 
@@ -233,7 +230,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       for (funSpec in funSpecs) {
         if (!funSpec.isConstructor) continue
         if (!firstMember) codeWriter.emit("\n")
-        funSpec.emit(codeWriter, name, implicitFunctionModifiers)
+        funSpec.emit(codeWriter, name, implicitMemberModifiers)
         firstMember = false
       }
 
@@ -241,18 +238,20 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       for (funSpec in funSpecs) {
         if (funSpec.isConstructor) continue
         if (!firstMember) codeWriter.emit("\n")
-        funSpec.emit(codeWriter, name, implicitFunctionModifiers)
+        funSpec.emit(codeWriter, name, implicitMemberModifiers)
         firstMember = false
       }
 
       // Types.
+      val areNestedExternal = EXTERNAL in modifiers || isNestedExternal
+
       for (typeSpec in typeSpecs) {
         if (!firstMember) codeWriter.emit("\n")
-        typeSpec.emit(codeWriter, null)
+        typeSpec.emit(codeWriter, null, isNestedExternal = areNestedExternal)
         firstMember = false
       }
 
-      companionObject?.emit(codeWriter, null)
+      companionObject?.emit(codeWriter, null, isNestedExternal = areNestedExternal)
 
       codeWriter.unindent()
       codeWriter.popType()
--- a/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
@@ -84,4 +84,15 @@ class PropertySpecTest {
     assertThat(a == b).isTrue()
     assertThat(a.hashCode()).isEqualTo(b.hashCode())
   }
+
+  @Test
+  fun externalTopLevel() {
+    val prop = PropertySpec.builder("foo", String::class)
+        .addModifiers(KModifier.EXTERNAL)
+        .build()
+
+    assertThat(prop.toString()).isEqualTo("""
+      |external val foo: kotlin.String
+      |""".trimMargin())
+  }
 }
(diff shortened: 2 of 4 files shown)


cypressious:
Resolves #131

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -547,6 +556,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     @JvmStatic fun expectClassBuilder(name: String) = Builder(Kind.EXPECT_CLASS, name, null).apply {
       addModifiers(KModifier.EXPECT)
     }
+    @JvmStatic fun externalClassBuilder(name: String) = Builder(Kind.EXTERNAL_CLASS, name, null).apply {
```
Egorand: nit: looks like a newline before this declaration is missing

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -452,12 +457,13 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addProperty(propertySpec: PropertySpec) = apply {
-      if (kind == Kind.EXPECT_CLASS) {
+      if (kind == Kind.EXPECT_CLASS || kind == Kind.EXTERNAL_CLASS) {
+        val modifier = if (kind == Kind.EXPECT_CLASS) "expect" else "external"
         require(propertySpec.initializer == null) {
-          "properties in expect classes can't have initializers"
+          "properties in $modifier classes can't have initializers"
```
Egorand: I think it's fine to just print `$kind` here

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -499,6 +505,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         Kind.EXPECT_CLASS -> require(funSpec.body.isEmpty()) {
           "functions in expect classes can't have bodies"
         }
+        Kind.EXTERNAL_CLASS -> require(funSpec.body.isEmpty()) {
```
Egorand: Same here, you can merge this case with `Kind.EXPECT_CLASS` and print `$kind` instead of `expect` and `external`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
cypressious: Thanks, fixes

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
cypressious: done

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
cypressious: done

swankjesse:
I’m still not 100% onboard with being able to specify this at both the function level and the class level. I’d like for KotlinPoet to automatically promote `external` in all members to `external` at the top level when emitting.

But in the KotlinPoet model I think it’s better to be able to specify `external` in one place only. That way any code that operates on KotlinPoet models doesn’t need to look in two places and do precedence.

cypressious:
@swankjesse Not sure if this is covered by your statement but there can be top-level `external` properties and functions.

Egorand:
Looks like it's even possible to have `external interface`s: https://kotlinlang.org/docs/reference/js-interop.html I wonder if `external object` would work?

cypressious:
How should we proceed with this issue?

Egorand:
How complete is our support for `external` at the moment? Any use-cases that we're not covering yet?
It looks good to me. There's no way to achieve what @swankjesse suggests since `external` is so flexible. @JakeWharton what'd you say?

cypressious:
Last time I tested it, it was completely broken. I wasn't able to declare
an external top-level val or an external class with members.

Kirill Rakhman

Am 19.01.2018 20:02 schrieb "Egor Andreevici" <notifications@github.com>:

swankjesse:
@Egorand & I have discussed offline and we have some goals!

1. Do require the `external` modifier on all members of an external type.
2. Don’t emit the `external` modifier on all members of an external type.
3. Support `external` on classes, interfaces, properties, functions, and possibly objects.
4. Support `external` on members of companion objects.

cypressious:
@swankjesse Good to hear. Do you need me to change this PR accordingly or do you plan to implement it yourself?

Egorand:
@cypressious, This PR is a good starting point, if you're interested to continue that would be awesome!

cypressious:
@swankjesse can you give me an example for a valid `external` member of a companion objects?

cypressious:
I rebased on master and redid the changes. Let me know if something is missing.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2908,48 @@ class TypeSpecTest {
...
+      |package com.squareup.tacos
+      |
+      |external class Foo {
+      |    fun bar()
+      |}
+      |""".trimMargin())
+  }
+
+  @Test
+  fun externalInterfaceFunctionHasNoBody() {
```
Egorand: The name of the test doesn't correspond to the contents. Would actually be great to have properties and/or functions inside to verify that they're printed properly.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2908,48 @@ class TypeSpecTest {
...
+    assertThat(toString(typeSpec)).isEqualTo("""
+      |package com.squareup.tacos
+      |
+      |external interface Foo
+      |""".trimMargin())
+  }
+
+
+  @Test
+  fun externalObjectFunctionHasNoBody() {
```
Egorand: Same as above

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -608,6 +608,12 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
     @JvmStatic fun expectClassBuilder(className: ClassName) = expectClassBuilder(className.simpleName())
 
+    @JvmStatic fun externalClassBuilder(name: String) = Builder(Kind.CLASS, name, null).apply {
```
Egorand: I wonder if factory methods like these are really needed...  We allow `external` on interfaces and objects, but we don't have specific factory methods for those, and it would've been too much anyway. `external class` is undoubtedly a more popular use-case than `external interface`, but does it warrant having such a specific factory method? We might think about removing `expectClassBuilder()` as well...

JakeWharton: I would very much prefer not having specific factory methods!

Egorand:
@cypressious 

> can you give me an example for a valid external member of a companion objects?

It's this: [Declaring static members of a class](https://kotlinlang.org/docs/reference/js-interop.html#declaring-static-members-of-a-class)

We want to be able to output:

```kotlin
external class MyClass {
    companion object {
        fun hasNoBodyAndItsFine()
    }
}
```

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
cypressious: Should I remove all of them or just don't add the `external` ones?

Egorand: In scope of this PR, just don't add new ones

cypressious:
@Egorand thx! It seems that you can nest all kinds of types in `external` classes, not only `companion objects`.

Egorand:
Interesting, how deep can you nest? What does it look like in JS?

cypressious:
Indefinitely. 

```kotlin
external class Foo {
  class Nested1 {
    object Nested2 {
      fun bar()
    }
  }
}

Foo.Nested1.Nested2.bar()
```

is compiled to

```js
var Foo$Nested1$Nested2 = Foo.Nested1.Nested2;
Foo$Nested1$Nested2.bar();
```

Egorand:
Oh wow! Any idea how we can model this in Kotlinpoet?

cypressious:
Yes, I actually have it working. I require that nested types have the `EXTERNAL` modifier but simply don't emit it.

cypressious:
PR is updated

src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -84,4 +84,15 @@ class PropertySpecTest {
     assertThat(a == b).isTrue()
     assertThat(a.hashCode()).isEqualTo(b.hashCode())
   }
+
+  @Test
+  fun externalTopLevel() {
+    val prop = PropertySpec.builder("foo", String::class)
+            .addModifiers(KModifier.EXTERNAL)
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2917,101 @@ class TypeSpecTest {
           |""".trimMargin())
   }
 
+  @Test
+  fun externalClassFunctionHasNoBody() {
+    val typeSpec = TypeSpec.classBuilder("Foo")
+            .addModifiers(KModifier.EXTERNAL)
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2917,101 @@ class TypeSpecTest {
...
+      |external class Foo {
+      |    fun bar()
+      |}
+      |""".trimMargin())
+  }
+
+  @Test
+  fun externalInterfaceWithMembers() {
+    val typeSpec = TypeSpec.interfaceBuilder("Foo")
+            .addModifiers(KModifier.EXTERNAL)
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2917,101 @@ class TypeSpecTest {
...
+      |    fun bar()
+      |}
+      |""".trimMargin())
+  }
+
+
+  @Test
+  fun externalObjectWithMembers() {
+    val typeSpec = TypeSpec.objectBuilder("Foo")
+            .addModifiers(KModifier.EXTERNAL)
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -2917,6 +2917,101 @@ class TypeSpecTest {
...
+      |
+      |    fun bar()
+      |}
+      |""".trimMargin())
+  }
+
+  @Test
+  fun externalClassWithNestedTypes() {
+    val typeSpec = TypeSpec.classBuilder("Foo")
+            .addModifiers(KModifier.EXTERNAL)
```
Egorand: nit: 4-space indent

cypressious:
fixed the indents

src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -84,4 +84,15 @@ class PropertySpecTest {
     assertThat(a == b).isTrue()
     assertThat(a.hashCode()).isEqualTo(b.hashCode())
   }
+
+  @Test
+  fun externalTopLevel() {
```
swankjesse: nice

Egorand:
Thanks @cypressious !

PR #288, opened 2017-11-28
Kotlin 1.2

Egorand:


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -310,7 +310,7 @@ class KotlinPoetTest {
         .addFunction(FunSpec.builder("shrink")
             .returns(String::class)
             .receiver(LambdaTypeName.get(
-                parameters = String::class.asClassName(),
+                parameters = *arrayOf(String::class.asClassName()),
```
JakeWharton: Well this is gross...

Egorand: Passing varargs as named parameters is deprecated now, not sure why, I'll try to find more info. Unfortunately `[]` don't work here.

PR #295, opened 2017-12-14
Move all dependency versions under properties

Egorand:
Should make it easier to upgrade without having to scroll through the entire file

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,15 @@
     <junit.version>4.12</junit.version>
     <truth.version>0.36</truth.version>
     <compile-testing.version>0.12</compile-testing.version>
+    <jimfs.version>1.1</jimfs.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -27,6 +27,15 @@
     <junit.version>4.12</junit.version>
     <truth.version>0.36</truth.version>
     <compile-testing.version>0.12</compile-testing.version>
+    <jimfs.version>1.1</jimfs.version>
+    <mockito.version>2.11.0</mockito.version>
+    <ecj.version>4.6.1</ecj.version>
```
JakeWharton: Must. Kill. ECJ.

Egorand: That stuff is only used in a couple of `@Ignore`d test classes, I doubt they're still relevant

PR #303, opened 2017-12-15
Set com.squareup.kotlinpoet as automatic module name

sormuras:
Claim `com.squareup.kotlinpoet` as Automatic-Module-Name in the JAR manifest.

See http://branchandbound.net/blog/java/2017/12/automatic-module-name/ for details and background.

Analog to https://github.com/square/javapoet/commit/03f3a276b2f330c59e549e73542217c6028caed6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,6 +203,17 @@
           </descriptors>
         </configuration>
       </plugin>
+      <plugin>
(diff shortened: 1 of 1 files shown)


JakeWharton:
Thanks!

PR #313, opened 2018-01-10
Escape names in statements

helpermethod:
Related to #298.

While parameters are correctly escaped now, they won't be escaped when referred to in a statement, like

    // param is `do` but addStatement outputs do
    addStatement("%N", escapedParam)

This PR fixes this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
 Change Log
 ==========
 
+## Version 0.6.0
+
+_2017-11-03_
+
+ * New: Support lambda extensions.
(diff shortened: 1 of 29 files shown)


Egorand:
Looks good! I think we can have more tests on this one though. Also, I figured all of this works:

```kotlin
class `if` {

  val `if` = 1

  fun `if`() {
  }
}
```

Should we escape names in all of those?

helpermethod:
Sorry, I'll add some tests today.

For my current use case I don't need escaping on class name and variable declarations but I guess it makes sense overall for consistency reasons.

Egorand:
Cool, thank you!

PR #316, opened 2018-01-23
Correcting handling of super-classes/interfaces on anonymous classes

### LINKED ISSUE #315: Anonymous classes don't seem to properly handle interfaces
```kotlin
interface Blah {
    fun method(): String
}

fun main(args: Array<String>) {
    println(TypeSpec.anonymousClassBuilder("")
            .addSuperinterface(Blah::class)
            .addFunction(FunSpec.builder("method").returns(String::class).addCode("return %S", "hello").build())
            .build().toString())

}
```

This prints:
```kotlin
object : Blah() {
  fun method(): kotlin.String = "hello"
}
```
Which does not compile because `Blah` does not have a constructor. The parenthesis should not be there. Looks like the problem is here : https://github.com/square/kotlinpoet/blob/master/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt#L83

kiiadi:
Fixes #315

Also added helper method for constructing anonymous classes that have no arguments passed to a superclass  : `TypeSpect.anonymousClassBuilder()`

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -79,10 +79,25 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
         codeWriter.emit(" {\n")
       } else if (anonymousTypeArguments != null) {
-        val supertype = if (superinterfaces.isNotEmpty()) superinterfaces[0] else superclass
-        codeWriter.emitCode("object : %T(", supertype)
-        codeWriter.emitCode(anonymousTypeArguments)
-        codeWriter.emit(") {\n")
+
+        codeWriter.emitCode("object")
+        val supertype = if (superclass != ANY) {
+          listOf(CodeBlock.of(" %T(%L)", superclass, anonymousTypeArguments))
+        } else {
+          listOf()
+        }
+
+        val allSuperTypes = supertype + superinterfaces.map { CodeBlock.of(" %T", it) }
+
+        if (allSuperTypes.isNotEmpty()) {
+          codeWriter.emitCode(" :")
+          codeWriter.emitCode(allSuperTypes.joinToCode(","))
+        }
+        if (hasNoBody) {
+          codeWriter.emit(" { }")
+          return
+        }
+        codeWriter.emit(" {\n")
       } else {
         codeWriter.emitKdoc(kdoc)
         codeWriter.emitAnnotations(annotations, false)
@@ -524,12 +539,6 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
       }
 
-      val superclassIsAny = superclass == ANY
-      val interestingSupertypeCount = (if (superclassIsAny) 0 else 1) + superinterfaces.size
-      require(anonymousTypeArguments == null || interestingSupertypeCount <= 1) {
-        "anonymous type has too many supertypes"
-      }
-
       if (primaryConstructor == null) {
         require(funSpecs.none { it.isConstructor } || superclassConstructorParameters.isEmpty()) {
           "types without a primary constructor cannot specify secondary constructors and superclass constructor parameters"
@@ -565,6 +574,10 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     @JvmStatic fun enumBuilder(name: String) = Builder(Kind.ENUM, name, null)
 
     @JvmStatic fun enumBuilder(className: ClassName) = enumBuilder(className.simpleName())
+
+    @JvmStatic fun anonymousClassBuilder(): Builder {
+      return anonymousClassBuilder("")
+    }
 
     @JvmStatic fun anonymousClassBuilder(typeArgumentsFormat: String, vararg args: Any): Builder {
       return Builder(Kind.CLASS, null, CodeBlock.builder()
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -79,10 +79,22 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          listOf()
+        }
+
+        val allSuperTypes = supertype + superinterfaces.map { CodeBlock.of(" %T", it) }
+        if (allSuperTypes.size > 1) {
+          allSuperTypes.dropLast(1).forEach { codeWriter.emitCode(it).emit(",") }
+        }
+        if (allSuperTypes.isNotEmpty()) {
+          codeWriter.emitCode(allSuperTypes.last())
+        }
```
JakeWharton: These `if`s should be able to be replaced by the `joinToCode` helper.

kiiadi: 👍 thanks - I knew there must have been a better way!

kiiadi: Much neater - thanks!

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -79,10 +79,18 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
         codeWriter.emit(" {\n")
       } else if (anonymousTypeArguments != null) {
-        val supertype = if (superinterfaces.isNotEmpty()) superinterfaces[0] else superclass
-        codeWriter.emitCode("object : %T(", supertype)
-        codeWriter.emitCode(anonymousTypeArguments)
-        codeWriter.emit(") {\n")
+
+        codeWriter.emitCode("object :")
+        val supertype = if (superclass != ANY) {
```
NightlyNexus: won't this break if i want
```kotlin
val FOO: Any = object : Any() {
}
```
for some reason?

how about `if (superclass != ANY || superinterfaces.isEmpty())`?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -154,6 +154,39 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  // https://github.com/square/kotlinpoet/issues/315
+  @Test fun anonymousClassWithMultipleSuperTypes() {
+    val superclass = ClassName("com.squareup.wire", "Message")
+    val anonymousClass = TypeSpec.anonymousClassBuilder()
+            .superclass(superclass)
```
Egorand: nit: continuation indent should be 4 spaces

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: That's equivalent to
```kotlin
val FOO: Any = object {
}
```
which is what will be emitted

NightlyNexus: oh, cool, I didn't know that was valid Kotlin. Thanks!

NightlyNexus: actually, won't this emit `...object : {...` (extra colon), or am i missing something again?

Egorand: Oh, right, `:` would be redundant

kiiadi:
@Egorand have included a test case (and logic) to omit the extra `:` in the case that no super-types / super-interfaces are set.

kiiadi:
Changed the logic so that the empty block emits as 
```kotlin
object { }
```
rather than 
```kotlin
object {
}
```

As this seems to better adhere to the style guides I've seen. Let me know if you agree.

Do I need to squash these down into one commit?

Egorand:
@kiiadi agree with the style for brackets. Yes, please squash. Thanks!

kiiadi:
Squashed.

Egorand:
@JakeWharton can you take another look?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -79,10 +79,25 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+        }
+
+        val allSuperTypes = supertype + superinterfaces.map { CodeBlock.of(" %T", it) }
+
+        if (allSuperTypes.isNotEmpty()) {
+          codeWriter.emitCode(" :")
+          codeWriter.emitCode(allSuperTypes.joinToCode(","))
+        }
+        if (hasNoBody) {
+          codeWriter.emit(" { }")
```
swankjesse: is this to style? Typically in Java we always line break for simple consistency

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -154,6 +154,58 @@ class TypeSpecTest {
...
+        ).build()
+
+    assertThat(toString(taco)).isEqualTo("""
+        |package com.squareup.tacos
+        |
+        |import com.squareup.wire.Message
+        |import java.lang.Runnable
+        |
+        |class Taco {
+        |    val NAME: Runnable = object : Message(), Runnable {
```
swankjesse: cool!

PR #317, opened 2018-02-02
Support PropertySpecs as CodeBlock literals

hzsweers:
This adds support for PropertySpecs as CodeBlock literals. I wasn't totally sure where to best test this, so let me know if you're prefer it go somewhere else!


hzsweers:
\o/

swankjesse:
One thing to keep in mind: although their declarations are similar, locals and properties are different things. We wanna avoid encouraging ourselves and others to use `PropertySpec` as a `LocalSpec`!

PR #318, opened 2018-02-09
Fix boundary filtering to Any?

hzsweers:


src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -56,12 +56,14 @@ class TypeVariableName private constructor(
   override fun emit(out: CodeWriter) = out.emit(name)
 
   companion object {
+    private val NULLABLE_ANY = ANY.asNullable()
```
Egorand: Should we add that to the other `ClassName`s we define in `TypeName`? Also, `ANY?` is an option for the field name, `NULLABLE_ANY` is totally fine though.

hzsweers: Up to y'all!

hzsweers:
I can add a test

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
Egorand: Sorry, wanted to say `` `ANY?` ``

hzsweers: I'll leave this along for now and allow y'all to do a follow up if people want

PR #323, opened 2018-02-15
Support dynamic type

### LINKED ISSUE #321: Support `dynamic` for Kotlin JS
There doesn't seem to be any direct support for the `dyanmic` type for Kotlin JS. You can hack it together using something like:

```Kotlin
val DYNAMIC = ClassName("", "dynamic")
```

It would be nice if kotlinpoet provided support for it out of the box though.

Egorand:
Closes #321

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -217,6 +217,25 @@ abstract class TypeName internal constructor(
 @JvmField val FLOAT = ClassName("kotlin", "Float")
 @JvmField val DOUBLE = ClassName("kotlin", "Double")
 
+@JvmField val DYNAMIC = object : TypeName(false, emptyList()) {
+
+  override fun asNullable() =
+      throw UnsupportedOperationException("dynamic can't be nullable")
+
+  override fun asNonNullable() =
+      throw UnsupportedOperationException("dynamic can't be non-nullable")
+
+  override fun annotated(annotations: List<AnnotationSpec>) =
+      throw UnsupportedOperationException("dynamic can't have annotations")
+
+  override fun withoutAnnotations() =
+      throw UnsupportedOperationException("dynamic can't have annotations")
+
+  override fun emit(out: CodeWriter) = out.apply {
+    emit("dynamic")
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -216,6 +216,7 @@ abstract class TypeName internal constructor(
 @JvmField val CHAR = ClassName("kotlin", "Char")
 @JvmField val FLOAT = ClassName("kotlin", "Float")
 @JvmField val DOUBLE = ClassName("kotlin", "Double")
+@JvmField val DYNAMIC = ClassName("", "dynamic")
```
JakeWharton: I don't think it makes sense to model this as a `ClassName`. None of the methods on `ClassName` are relevant for it.

Egorand: Agree. What about `object Dynamic : TypeName()`?

JakeWharton: Works for me.

NightlyNexus:
`Dynamic.INSTANCE` is a little bit unfortunate for Java consumers, but I don't have a better idea.

Egorand:
You're right. I also thought about making `Dynamic` an anonymous class inside `TypeName` with `@JvmField`, would that be a better idea in terms of Java interop?

NightlyNexus:
I would lean towards not sacrificing idiomatic Kotlin for better Java interop.
+1 as is.

src/main/java/com/squareup/kotlinpoet/Dynamic.kt
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+/**
+ * JavaScript-specific `dynamic` type.
+ * See [https://kotlinlang.org/docs/reference/dynamic-type.html].
+ */
+object Dynamic : TypeName(false, emptyList()) {
```
swankjesse: Move to TypeName.kt

PR #324, opened 2018-02-17
Remove superfluous Kind definitions

Egorand:


swankjesse:
Yah I like this.

Egorand:
@JakeWharton @swankjesse do you think `ENUM` and `ANNOTATION` should go as well?

swankjesse:
Unsure. I didn’t like `expect` because it seemed like that might stack with other modifiers like `external`.

For `enum` and `annotation`, it mostly comes down to whatever is easier to understand and maintain.

JakeWharton:
I think so, since that's how the language models it.

Egorand:
Have been playing around with it, removing `ENUM` and `ANNOTATION` adds a ton of clutter: every time we `require(kind == ANNOTATION)` we'll now have to `require(kind == CLASS && ANNOTATION !in modifers)` and so on. While `EXPECT` and `EXTERNAL` can attach to any type, `ENUM` and `ANNOTATION` only make sense with a `class`, hence it's probably fine to model them as separate kinds. Maybe there's a cleaner solution that I'm missing.

JakeWharton:
That seems like something that can be hidden behind a helper function, no?

Egorand:
I've introduced a couple of helper properties:

1) You'll need helpers in both `TypeSpec` and `Builder`
2) Validations become complicated, e.g.:

```kotlin
val isAbstract = KModifier.ABSTRACT in modifiers || kind != Kind.CLASS || isEnum || isAnnotation
```

I'll give it more shots though, maybe I just need better helpers

JakeWharton:
Yeah I'm not surprised at the internal complexity growing, but I think it
should be worth it to have the external model match that of the language
itself. Especially since we're going to get more things to model like
inline classes in the future.

PR #329, opened 2018-02-25
Redesign TypeSpec.Kind

Egorand:


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -314,35 +307,51 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+    internal val modifiers: MutableSet<KModifier> = mutableSetOf()
   ) {
-    CLASS(
+
+    val isEnum get() = this is Class && ENUM in modifiers
+
+    val isAnnotation get() = this is Class && ANNOTATION in modifiers
+
+    val isSimpleClass get() = this is Class && !isEnum && !isAnnotation
+
```
JakeWharton: Private?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -28,7 +33,6 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val anonymousTypeArguments = builder.anonymousTypeArguments
   val kdoc = builder.kdoc.build()
   val annotations = builder.annotations.toImmutableList()
-  val modifiers = builder.modifiers.toImmutableSet()
```
JakeWharton: We still need to expose this (as Set). Delegate?

JakeWharton: Actually we need a defensive copy of the kind or to make it immutable.

Egorand: It shouldn't be possible to modify the instance as all fields are internal, but I can look into improving encapsulation. `sealed` doesn't work with `data`, right?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Yes

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: One builder creating two classes where the second adds a modifier will see that reflected on both.

Sealed works with data.

Egorand:
@swankjesse yeah, I was expecting that, will rebase tomorrow
@JakeWharton `Kind` is now immutable (`modifiers` is a `Set`) and all fields are `internal`, making them `private` won't work since they're referenced in `Builder`

PR #333, opened 2018-02-27
add FileSpec.addWildcardImport method

evgeny-pasynkov:
Add ability to import the whole package into file

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -253,6 +253,10 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
       }
     }
 
+    fun addWildcardImport(packageName: String) = apply {
+      memberImports += Import(packageName + ".*")
+    }
+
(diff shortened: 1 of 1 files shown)


JakeWharton:
We explicitly disallow wildcard imports.

evgeny-pasynkov:
What is the reason for such restriction?
It is up to the user of the library how he would like to see generated code.
I agree that wildcard shouldn't be default - it has be stated directly.

My case: generated code could contain invocation of hundred extension methods - listing all of them is ridiculous

evgeny-pasynkov:
With what I agree - adding wildcard import could be a little bit smarter: all individual imports should be removed when wildcard is present

JakeWharton:
Wildcard imports are people writing code without an IDE. There's no reason to ever see them when using an IDE or when doing code generation since you explicitly know everything that's used.

You don't need to list them all. You only need to add one when you use it. KotlinPoet will de-dupe the list for you.

Egorand:
Closing this. If there's an issue you're not able to overcome without wildcard imports we can look at it.

Issue #332, opened 2018-02-27
Incorrect super constructor call for anonymous object declaration


sandwwraith:
Given `open class Foo(val name: String)`, this code:

```kotlin
val ts = TypeSpec.anonymousClassBuilder().apply {
    superclass(Foo::class)
    addSuperclassConstructorParameter("%S", "name")
}.build()

FileSpec.builder("", "a").addType(ts).build().writeTo(System.out)
```
 produces `object : Foo() { }` instead of expected `object : Foo("name") {}`

But if I call `anonymousClassBuilder("%S", "name")`, result will be `object : Foo("name") {}`. Since method parameter called `typeArgumentsFormat` I suppose this also a wrong behaviour and output should be `object : Foo<name>() {}` (and probably method parameter type should be `TypeName` instead of `CodeBlock`).

Note if the described behaviour is correct, it's impossible to produce construction like `object : Bar<String>("some string") {}` given `open class Bar<T: Any>(val content: T)`

NightlyNexus:
>it's impossible to produce construction like object : Bar<String>("some string") {}

How about this?
```kotlin
val anonymousClass = TypeSpec.anonymousClassBuilder()
    .superclass(ParameterizedTypeName.get(ClassName("com.example", "Bar"), String::class.asClassName()))
    .addSuperclassConstructorParameter("%S", "some string")
    .build()
```

That said, I _think_ the `anonymousClassBuilder(typeArgumentsFormat: String, vararg args: Any)` overload is redundant given `addSuperclassConstructorParameter`. If I'm correct, the overload should be removed, and `addSuperclassConstructorParameter` should be made to work with anonymous classes.

I'll send a PR tonight for discussion.

PR #334, opened 2018-02-28
Remove anonymousTypeArguments. Enums use superclassConstructorParameters.

NightlyNexus:
~A test to demonstrate #332.~
~I'll update with the fix...~

Addresses #332 but also ended up refactoring and changing behavior. Happy to break into separate PRs.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -31,7 +31,6 @@ import kotlin.reflect.KClass
 class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val kind = builder.kind
   val name = builder.name
-  val anonymousTypeArguments = builder.anonymousTypeArguments
   val kdoc = builder.kdoc.build()
   val annotations = builder.annotations.toImmutableList()
   val modifiers = kind.modifiers.toImmutableSet()
@@ -40,6 +39,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val primaryConstructor = builder.primaryConstructor
   val superclass = builder.superclass
   val superclassConstructorParameters = builder.superclassConstructorParameters.toImmutableList()
+  private val isAnonymousClass = name == null && kind is Kind.Class
 
   /**
    * Map of superinterfaces - entries with a null value represent a regular superinterface (with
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -95,7 +95,14 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
         codeWriter.emitCode("object")
         val supertype = if (superclass != ANY) {
-          listOf(CodeBlock.of(" %T(%L)", superclass, anonymousTypeArguments))
+          var superclassConstructorParameters = superclassConstructorParameters
+          if (anonymousTypeArguments.isNotEmpty()) {
+            superclassConstructorParameters = superclassConstructorParameters
+                .toMutableList().apply {
+              add(anonymousTypeArguments)
```
NightlyNexus: made this append, but I'd also like to get rid of `anonymousTypeArguments` completely, I think?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -465,7 +462,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     private fun ensureCanHaveSuperclass() {
-      check(kind.isSimpleClass || kind is Object) {
+      check(kind.isSimpleClass || kind.isEnum || kind is Object) {
         "only classes can have super classes, not $kind"
```
JakeWharton: This message is starting to lose its accuracy.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -422,17 +422,14 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     fun addAnnotation(annotation: KClass<*>) = addAnnotation(annotation.asClassName())
 
     fun addModifiers(vararg modifiers: KModifier) = apply {
-      check(anonymousTypeArguments == null) { "forbidden on anonymous types." }
```
Egorand: Is there anything replacing these checks? I'm fine with just removing them, but it'd be nice to keep them

NightlyNexus: yes, i think we can do the same `isAnonymousClass` check from above. good idea!

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
NightlyNexus: whoops, saw that earlier and forgot to fix. brb.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: Would be nice to have tests for this in the future

NightlyNexus: #335

PR #335, opened 2018-02-28
Add tests for confines on anonymous classes.

NightlyNexus:
And restore stricter ensureCanHaveSuperclass check.


JakeWharton:
Thanks!

PR #338, opened 2018-03-03
Remove Mockito.

NightlyNexus:


src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -113,24 +115,16 @@ class FunSpecTest {
...
-    }.hasMessageThat().isEqualTo("cannot override method with modifiers: [private]")
+      FunSpec.overriding(findFirst(methods, "finalMethod"))
+    }.hasMessageThat().isEqualTo("cannot override method with modifiers: [public, final]")
 
-    whenMock(method.modifiers).thenReturn(ImmutableSet.of(Modifier.STATIC))
     assertThrows<IllegalArgumentException> {
-      FunSpec.overriding(method)
-    }.hasMessageThat().isEqualTo("cannot override method with modifiers: [static]")
+      FunSpec.overriding(findFirst(methods, "privateMethod"))
+    }.hasMessageThat().isEqualTo("cannot override method with modifiers: [private, final]")
```
NightlyNexus: redundant test. (public final, private final) can remove if you prefer.

JakeWharton: Does `private open` work? Also what about `@JvmStatic` to get a static method?

NightlyNexus: `private open` does not compile.
good idea on static. will update restoring that.

PR #348, opened 2018-03-12
Fix importing annotated TypeName

### LINKED ISSUE #346: Annotated TypeName not imported when used as type argument in ParameterizedTypeName
```kotlin
@Test fun typeArgumentNotImported() {
  val file = FileSpec.builder("com.squareup.tacos", "Taco")
      .addFunction(FunSpec.builder("foo")
          .addParameter("a", ParameterizedTypeName.get(
              rawType = List::class.asTypeName(),
              typeArguments = *arrayOf(Int::class.asTypeName()
                  .annotated(AnnotationSpec.builder(JvmSuppressWildcards::class)
                      .build()))
          ))
          .build())
      .build()
  assertThat(file.toString()).isEqualTo("""
    |package com.squareup.tacos
    |
    |import kotlin.Int
    |import kotlin.collections.List
    |import kotlin.jvm.JvmSuppressWildcards
    |
    |fun foo(a: List<@JvmSuppressWildcards kotlin.Int>) {
    |}
    |""".trimMargin())
}
```

Without the annotation that would emit `List<Int>`

Egorand:
Fixes #346

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -325,8 +325,8 @@ internal class CodeWriter constructor(
       val resolved = resolve(simpleName)
       nameResolved = resolved != null
 
-      // We don't care about nullability here, as it's irrelevant for imports.
-      if (resolved == c.asNonNullable()) {
+      // We don't care about nullability and type annotations here, as it's irrelevant for imports.
+      if (resolved == c.asNonNullable().withoutAnnotations()) {
         if (alias != null) return alias
         val suffixOffset = c.simpleNames().size - 1
         return className.simpleNames().subList(suffixOffset,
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -650,4 +650,28 @@ class KotlinPoetTest {
       |class Taco(@JvmField val foo: String)
       |""".trimMargin())
   }
+
+  // https://github.com/square/kotlinpoet/issues/346
+  @Test fun typeArgumentNotImported() {
```
JakeWharton: Name seems backwards

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -326,7 +326,7 @@ internal class CodeWriter constructor(
       nameResolved = resolved != null
 
       // We don't care about nullability here, as it's irrelevant for imports.
```
JakeWharton: Update comment to match

PR #355, opened 2018-04-07
Add type inference for PropertySpec

kaonash:
Enable type inference at property declaration.
```
val foo = "bar"
```

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
@@ -95,4 +95,28 @@ class PropertySpecTest {
       |external val foo: kotlin.String
       |""".trimMargin())
   }
+
+  @Test fun initializer() {
+    val prop = PropertySpec.builder("foo", String::class)
+            .initializer("\"bar\"")
+            .build()
+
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -95,4 +95,28 @@ class PropertySpecTest {
...
+    val prop = PropertySpec.builder("foo", String::class)
+        .initializerWithTypeInference("\"bar\"")
+        .build()
+
+    assertThat(prop.toString()).isEqualTo("val foo = \"bar\"\n")
+  }
+
+  @Test fun delegateWithTypeInferrence() {
+    val prop = PropertySpec.builder("foo", String::class)
+            .delegateWithTypeInference("Delegates.notNull<String>()")
```
JakeWharton: This violates the guarantees otherwise provided by the library. A sibling or nested class with the name `Delegates` or `String` will break this.

kaonash:
Thank you for your reply.
As you say, I always know the types, but sometimes I want to generate more concise code at property declaration.
What do you think?

kaonash:
Thanks a lot.
I would like to close this PR.

PR #367, opened 2018-05-13
Made packageName and name properties public

### LINKED ISSUE #366: Make packageName in FileSpec.Builder public
The [`packageName` member in the `FileSpec.Builder` class](https://github.com/square/kotlinpoet/blob/master/src/main/java/com/squareup/kotlinpoet/FileSpec.kt#L172) is currently `internal` and thus not accessible from the outside. This is pretty annoying because one needs to declare a constant for the package name or repeat the same string over and over again. Having access to this member would remove this necessity.

```kotlin
FileSpec.build("com.fleshgrinder", "KotlinPoet").apply {
    val someClass = ClassName(packageName, "SomeClass")
    val anotherClass = ClassName(packageName, "AnotherClass")
    // ...
}.build().writeTo(Sytem.out)
```

It might be a good idea to directly open up the `name` as well since this might also help to reduce the amount of repetition in certain situations, e.g.:

```kotlin
FileSpec.build("com.fleshgrinder", "KotlinPoet").apply {
    val `class` = ClassName(packageName, name)
    // ...
}.build().writeTo(Sytem.out)
```

There is no harm in making them public since both properties are `val` anyways.

Fleshgrinder:
Closes #366

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -169,8 +169,8 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
   }
 
   class Builder internal constructor(
-    internal val packageName: String,
-    internal val name: String
+    val packageName: String,
+    val name: String
   ) {
     internal val annotations = mutableListOf<AnnotationSpec>()
     internal val comment = CodeBlock.builder()


Egorand:
Thank you!

Egorand:
Almost forgot to ask you to [sign our CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1), could you do that please?

PR #372, opened 2018-05-21
Escape (class/property/funtion/variable) names automatically if they contain space, hyphen, or other symbols

alphaho:
Changes for issue #369

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -350,7 +350,7 @@ class FunSpecTest {
 
   @Test fun escapePunctuationInFunctionName() {
     val funSpec = FunSpec.builder("with-hyphen")
-            .build()
+        .build()
 
     assertThat(funSpec.toString()).isEqualTo("""
       |fun `with-hyphen`() {
--- a/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -347,4 +347,14 @@ class FunSpecTest {
       |}
       |""".trimMargin())
   }
+
+  @Test fun escapePunctuationInFunctionName() {
+    val funSpec = FunSpec.builder("with-hyphen")
+            .build()
+
+    assertThat(funSpec.toString()).isEqualTo("""
+      |fun `with-hyphen`() {
+      |}
+      |""".trimMargin())
+  }
 }
--- a/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
@@ -88,6 +88,13 @@ class UtilTest {
     assertThat("".isIdentifier).isFalse()
   }
 
+  @Test fun escapeNonJavaIdentifiers() {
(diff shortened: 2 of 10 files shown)


src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -347,4 +347,14 @@ class FunSpecTest {
       |}
       |""".trimMargin())
   }
+
+  @Test fun escapePunctuationInFunctionName() {
+    val funSpec = FunSpec.builder("with-hyphen")
+            .build()
```
Egorand: nit: please use 4 spaces for continuation indents

src/test/java/com/squareup/kotlinpoet/UtilTest.kt
```diff
@@ -88,6 +88,13 @@ class UtilTest {
     assertThat("".isIdentifier).isFalse()
   }
 
+  @Test fun escapeNonJavaIdentifiers() {
+    assertThat(escapeIfNotJavaIdentifier("8startWithNumber")).isEqualTo("`8startWithNumber`")
```
Egorand: I like the tests!

hzsweers:
Can imports be covered in this too? Just ran into that case in some internal model generation work where we have a package name like this: `import com.uber.model.core.generated.rtapi.models.object.Meta`

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
alphaho: sure. let me update it accordingly

src/test/java/com/squareup/kotlinpoet/UtilTest.kt
alphaho: thanks very much =D

alphaho:
hi @hzsweers ,  it seems the example `import com.uber.model.core.generated.rtapi.models.object.Meta` works with the original implementation, which would call escapeIfKeyword for each part of the package name.

On escaping package names with hyphens or other symbols, it should be straightforward to make the change, I'm just not quite sure whether it's a good idea or not.

Egorand:
@alphaho there are multiple places where the continuation indent is off, I just marked the first one - could you please fix all of them?

hzsweers:
Just tried building against master and it looks like it didn't work with my case. Will dig in more

hzsweers:
Nevermind, the failure is due to moshi kotlin codegen using a shaded kotlinpoet. All good :+1:

alphaho:
@Egorand oops, let me fix them all. BTW, is the checkstyle not working properly? Thought it would help to flag all the inconsistent indents.

alphaho:
@Egorand about `TypeAliasSpec` name, `ClassName` name, and potentially something else, I would take a look at it. I initially thought all the ones that call `escapeIfKeyword` would be the places needed to be updated.

Egorand:
@alphaho Thanks! There's no checkstyle setup for this project, we should probably fix this, or change CONTRIBUTING.md, as it's misleading at the moment

Egorand:
> @Egorand about TypeAliasSpec name, ClassName name, and potentially something else, I would take a look at it. I initially thought all the ones that call escapeIfKeyword would be the places needed to be updated.

Sounds good! As I said, it's fine if those are in small follow up PRs

Egorand:
Thank you!

alphaho:
You are welcome :-)

PR #373, opened 2018-05-23
Make subtype checking vals inside Kind public

Egorand:


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -325,11 +325,11 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
   ) {
 
-    internal val isEnum get() = this is Class && ENUM in modifiers
+    val isEnum get() = this is Class && ENUM in modifiers
 
-    internal val isAnnotation get() = this is Class && ANNOTATION in modifiers
+    val isAnnotation get() = this is Class && ANNOTATION in modifiers
 
-    internal val isSimpleClass get() = this is Class && !isEnum && !isAnnotation
+    val isSimpleClass get() = this is Class && !isEnum && !isAnnotation
```
swankjesse: Not sure _simple_ is appropriate for our public API. It doesn’t have a formal definition, and users might be surprised that this counts both data classes and not.

I’d prefer to either omit this method or find a name that fits better?

swankjesse: (is _simple_ the word the Kotlin docs use?)

Egorand: No, I actually thought that it's a really bad name, but can't think of a better alternative... Probably hiding it is the best solution

Egorand: This val just aggregates the values returned by other public vals anyway, so I'll revert the change and make it `internal`

PR #376, opened 2018-05-25
Demo test and (naive) fix for suggested imports on matching type name collisions

hzsweers:
This attempts to fix a bug I came across when managing imports and type name collisions. I've been talking with @Egorand and not sure this is the right solution, but it does make the test pass at least.

Currently master produces this code for the test snippet, which isn't correct because the `Taco` import overrides the `anotherTaco` constructor param type and doesn't match the spec

```kotlin
package com.squareup.tacos

import java.sql.Taco
import kotlin.Int

data class Taco(val madeFreshDatabaseDate: Taco, fooNt: Int) {
  constructor(anotherTaco: Taco) : this(java.sql.Taco.defaultInstance(), 0)
}
```

Which doesn't compile because the `

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -55,7 +55,9 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     // First pass: emit the entire class, just to collect the types we'll need to import.
     val importsCollector = CodeWriter(NullAppendable, indent, memberImports)
     emit(importsCollector)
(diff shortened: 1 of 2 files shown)


Egorand:
Discussed offline with @swankjesse, `CodeWriter` uses `pushType()` to navigate nested namespaces, which ensures dependencies are fully qualified in right places if there's a name clash. Looks we're not doing `pushType()` in the right place when emitting the `TypeSpec`.

hzsweers:
Do you want me to switch this to just an issue for now then? I probably won't be able to go look until this weekend

Egorand:
Yep, that'd be great, and your test case is super helpful. We'll try to look into that hopefully this week.

PR #377, opened 2018-05-25
Remove companionObject API and funnel everything through types

hzsweers:
This bit me in some prototyping I was doing. Currently nothing prevents users from adding companion objects via `addType`/`addTypes`, which leads to somewhat undefined behavior at emission-time. I suspect this isn't something that should be allowed, so putting this up as a proposal PR

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -603,10 +603,18 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun addTypes(typeSpecs: Iterable<TypeSpec>) = apply {
+      typeSpecs.forEachIndexed { index, typeSpec ->
+        require(!(typeSpec.kind is Object && COMPANION in typeSpec.kind.modifiers)) {
+          "provided TypeSpec at index $index is a companion object, use the companionObject() function to set it"
+        }
+      }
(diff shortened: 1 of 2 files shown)


Egorand:
Agree with the solution. I thought about an alternative one to just redirect `addType()` to `companionObject()` if the argument is a companion object, but I guess it's better to have just a single way to pass in a companion object.

Egorand:
Discussed this offline with @swankjesse, we probably shouldn't be handling companion objects differently from other subtypes. How about merging `TypeSpec.companionObject` into `TypeSpec.typeSpecs` and making `companionObject()` just call into `addType()`?

hzsweers:
That works for me, but should there be a check to make sure that multiple companion objects aren't added?

Egorand:
Definitely, I'd say we should throw if someone's trying to add a second companion.

hzsweers:
done, removed the `companionObject` API all together and put validation for companion object checks in `build()`

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -628,6 +612,22 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
       }
 
+      val companionObjects = typeSpecs.filter {
```
swankjesse: I think you meant `typeSpecs.count()`

swankjesse:
Rebased, squashed, and merged manually!

PR #378, opened 2018-05-25
Proposal: Add isCompanion to Object kind to differentiate

hzsweers:
Ran into this case in internal prototyping and wanted to propose it as a PR. Currently there's no way to tell if a typespec of Object kind is a companion object or not


Egorand:
Agree with the point that the val should be inside `Object` vs `Kind`, we should probably move `isEnum` and `isAnnotation` inside `Class` too

hzsweers:
Cool, do you want me to move those as part of rebasing?

Egorand:
@hzsweers would be great, thank you!

hzsweers:
Rebased, looks like moving the others is addressed be #385

Issue #383, opened 2018-05-31
Move isEnum from Kind to Type


swankjesse:

mc-lovin:
taking this one :)

PR #385, opened 2018-06-02
Move isKind and isAnnotation to Type

mc-lovin:
https://github.com/square/kotlinpoet/issues/383

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -39,6 +39,8 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
   val primaryConstructor = builder.primaryConstructor
   val superclass = builder.superclass
   val superclassConstructorParameters = builder.superclassConstructorParameters.toImmutableList()
+  val isEnum = builder.isEnum
+  val isAnnotation = builder.isAnnotation
   private val isAnonymousClass = builder.isAnonymousClass
 
   /**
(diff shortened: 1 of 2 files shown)


mc-lovin:
@Egorand @swankjesse PTAL

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -400,6 +395,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     internal val funSpecs = mutableListOf<FunSpec>()
     internal val typeSpecs = mutableListOf<TypeSpec>()
     internal val isAnonymousClass get() = name == null && kind is Kind.Class
+    internal val isEnum get() = this.kind is Kind.Class && ENUM in kind.modifiers
+    internal val isAnnotation get() = this.kind is Kind.Class && ANNOTATION in kind.modifiers
+    internal val isSimpleClass = this.kind is Kind.Class && !isEnum && !isAnnotation
```
JakeWharton: super nit: `this` isn't needed in these

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -326,15 +328,8 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
-
-    val isAnnotation get() = this is Class && ANNOTATION in modifiers
-
-    internal val isSimpleClass get() = this is Class && !isEnum && !isAnnotation
-
-    internal val implicitPropertyModifiers get() =
-      if (isAnnotation) emptySet()
-      else defaultImplicitPropertyModifiers + when {
+    internal val implicitPropertyModifiers get() = defaultImplicitPropertyModifiers + when {
+        ANNOTATION in modifiers -> emptySet()
```
JakeWharton: I feel like this isn't correct, but it's unrelated to your PR. An expect annotation should get the implicits.

Egorand: Guess you're right. I'll try to write a failing test and will file an issue.

PR #388, opened 2018-06-03
Use isCompanion for companion object counting

hzsweers:
Following up on my note :)

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -612,7 +612,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
       }
 
-      val companionObjectsCount = typeSpecs.count { it.kind.let { it is Object && it.isCompanion } }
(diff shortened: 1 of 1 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -612,10 +612,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
       }
 
-      val companionObjectsCount = typeSpecs.count {
-        // Replace this with isCompanion check after merged
-        it.kind is Object && COMPANION in it.kind.modifiers
-      }
+      val companionObjectsCount = typeSpecs.count { it.kind.let { it is Object && it.isCompanion } }
```
JakeWharton: Is the nested `let` needed? I would rather it omitted. We should ban nested use of `it`s in the style guide.

hzsweers: Technically, no. I can remove it here but I personally don't think it should be blanket banned in a style guide :|

hzsweers: 7fce736

PR #390, opened 2018-06-05
Avoid type name collisions in primary constructor

### LINKED ISSUE #387: Type names collide when nested typename references have the same name
Carrying over from #376

The following test code

```kotlin
        .addType(TypeSpec.classBuilder("Taco")
            .addModifiers(KModifier.DATA)
            .addProperty(PropertySpec.builder("madeFreshDatabaseDate", sqlTaco)
                .initializer("madeFreshDatabaseDate")
                .build())
                .addParameter("madeFreshDatabaseDate", sqlTaco)
                .addParameter("fooNt", INT)
                .build())
            .addFunction(FunSpec.constructorBuilder()
                .addParameter("anotherTaco", ClassName("com.squareup.tacos", "Taco"))
                .callThisConstructor(CodeBlock.of("%T.defaultInstance(), 0", sqlTaco))
               .build())
            .build())
        .build()
```

Currently produces invalid code like this:

```kotlin
package com.squareup.tacos

import java.sql.Taco
import kotlin.Int

data class Taco(val madeFreshDatabaseDate: Taco, fooNt: Int) {
  constructor(anotherTaco: Taco) : this(java.sql.Taco.defaultInstance(), 0)
}
```

which isn't correct because the Taco import overrides the anotherTaco constructor param type and doesn't match the spec. It should look like this:

```kotlin
package com.squareup.tacos

import kotlin.Int

data class Taco(val madeFreshDatabaseDate: java.sql.Taco, fooNt: Int) {
    constructor(anotherTaco: Taco) : this(java.sql.Taco.defaultInstance(), 0)
}
```

Egorand:
Fixes #387 

Not sure this is a complete solution, but it seems to address the specific issue

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -129,6 +129,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         codeWriter.emitWhereBlock(typeVariables)
 
         primaryConstructor?.let {
+          codeWriter.pushType(this) // avoid name collisions when emitting primary constructor
           var useKeyword = false
           var emittedAnnotations = false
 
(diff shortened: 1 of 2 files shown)


hzsweers:
Nice

PR #399, opened 2018-06-10
Expose mutable builder properties and move their validations to build-time

hzsweers:
This is the analogue to https://github.com/square/javapoet/pull/650

Each commit should be digestible for review and isolated. The general approach was to make the properties public, add tests, and then move any of their validations to `build()`

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -273,7 +273,7 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     }
 
     fun build(): FileSpec {
-      annotations.forEach { annotationSpec ->
+      for (annotationSpec in annotations) {
         if (annotationSpec.useSiteTarget != FILE) {
           error(
               "Use-site target ${annotationSpec.useSiteTarget} not supported for file annotations.")
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -173,11 +173,12 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     val packageName: String,
     val name: String
   ) {
-    internal val annotations = mutableListOf<AnnotationSpec>()
     internal val comment = CodeBlock.builder()
     internal val memberImports = sortedSetOf<Import>()
     internal var indent = DEFAULT_INDENT
     internal val members = mutableListOf<Any>()
+
+    val annotations = mutableListOf<AnnotationSpec>()
 
     init {
       require(name.isName) { "not a valid file name: $name" }
@@ -271,7 +272,15 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
       this.indent = indent
     }
 
-    fun build() = FileSpec(this)
+    fun build(): FileSpec {
+      annotations.forEach { annotationSpec ->
+        if (annotationSpec.useSiteTarget != FILE) {
+          error(
+              "Use-site target ${annotationSpec.useSiteTarget} not supported for file annotations.")
+        }
+      }
+      return FileSpec(this)
+    }
   }
 
   companion object {
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -163,15 +163,16 @@ class FunSpec private constructor(builder: Builder) {
 
   class Builder internal constructor(internal val name: String) {
     internal val kdoc = CodeBlock.builder()
-    internal val annotations = mutableListOf<AnnotationSpec>()
-    internal val modifiers = mutableListOf<KModifier>()
-    internal val typeVariables = mutableListOf<TypeVariableName>()
     internal var receiverType: TypeName? = null
     internal var returnType: TypeName? = null
-    internal val parameters = mutableListOf<ParameterSpec>()
     internal var delegateConstructor: String? = null
     internal var delegateConstructorArguments = listOf<CodeBlock>()
     internal val body = CodeBlock.builder()
+
+    val annotations = mutableListOf<AnnotationSpec>()
+    val modifiers = mutableListOf<KModifier>()
+    val typeVariables = mutableListOf<TypeVariableName>()
+    val parameters = mutableListOf<ParameterSpec>()
 
     fun addKdoc(format: String, vararg args: Any) = apply {
       kdoc.add(format, *args)
@@ -226,12 +227,10 @@ class FunSpec private constructor(builder: Builder) {
     }
 
     fun addTypeVariables(typeVariables: Iterable<TypeVariableName>) = apply {
-      check(!name.isAccessor) { "$name cannot have type variables" }
       this.typeVariables += typeVariables
     }
 
     fun addTypeVariable(typeVariable: TypeVariableName) = apply {
-      check(!name.isAccessor) { "$name cannot have type variables" }
       typeVariables += typeVariable
     }
 
@@ -260,8 +259,6 @@ class FunSpec private constructor(builder: Builder) {
     }
 
     fun addParameter(parameterSpec: ParameterSpec) = apply {
-      check(name != GETTER) { "$name cannot have parameters" }
-      check(name != SETTER || parameters.size == 0) { "$name can have only one parameter" }
       parameters += parameterSpec
     }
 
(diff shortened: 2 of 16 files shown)


src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -271,7 +272,15 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
       this.indent = indent
     }
 
-    fun build() = FileSpec(this)
+    fun build(): FileSpec {
+      annotations.forEach { annotationSpec ->
```
swankjesse: Please use a regular for loop wherever possible.

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -271,7 +272,15 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
...
+        if (annotationSpec.useSiteTarget != FILE) {
+          error(
+              "Use-site target ${annotationSpec.useSiteTarget} not supported for file annotations.")
```
swankjesse: I don't follow this. Is this a new constraint?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -336,7 +333,14 @@ class FunSpec private constructor(builder: Builder) {
       body.addStatement(format, *args)
     }
 
-    fun build() = FunSpec(this)
+    fun build(): FunSpec {
+      if (typeVariables.isNotEmpty()) {
+        check(!name.isAccessor) { "$name cannot have type variables" }
```
swankjesse: Alternately,

check(typeVariables.isEmpty() || !name.isAccessor) { ...}

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -187,7 +185,10 @@ class PropertySpec private constructor(builder: Builder) {
 
     fun receiver(receiverType: KClass<*>) = receiver(receiverType.asTypeName())
 
-    fun build() = PropertySpec(this)
+    fun build(): PropertySpec {
+      modifiers.forEach { it.checkTarget(KModifier.Target.PROPERTY) }
```
swankjesse: Regular for loop please

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -97,7 +95,18 @@ class TypeAliasSpec private constructor(builder: TypeAliasSpec.Builder) {
       kdoc.add(block)
     }
 
-    fun build() = TypeAliasSpec(this)
+    fun build(): TypeAliasSpec {
+      modifiers.forEach {
```
swankjesse: Regular for loop please

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -598,6 +583,21 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
     fun build(): TypeSpec {
+      if (enumConstants.isNotEmpty()) {
+        check(isEnum) { "${this.name} is not enum and cannot have enum constants" }
+        enumConstants.keys.forEach {
+          require(it.isName) { "not a valid enum constant: $name" }
+        }
+      }
+
+      if (superclassConstructorParameters.isNotEmpty()) {
+        ensureCanHaveSuperclass()
```
swankjesse: Off topic: we should use the word check, not ensure, right?

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -97,7 +95,18 @@ class TypeAliasSpec private constructor(builder: TypeAliasSpec.Builder) {
...
+        require(it in ALLOWABLE_MODIFIERS) {
+          "unexpected typealias modifier $it"
+        }
+      }
+      return TypeAliasSpec(this)
+    }
+
+    private companion object {
+      private val ALLOWABLE_MODIFIERS = setOf(PUBLIC, INTERNAL, PRIVATE, ACTUAL)
```
Egorand: nit: shouldn't it be ALLOWED? non-native speaker here, ignore me if that's the correct wording

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Agree, let's be consistent

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -598,6 +583,21 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
     }
 
     fun build(): TypeSpec {
+      if (enumConstants.isNotEmpty()) {
+        check(isEnum) { "${this.name} is not enum and cannot have enum constants" }
+        enumConstants.keys.forEach {
```
Egorand: For-loop here too

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
hzsweers: Technically it's not. In `addAnnotation`, if there is no site target, it synthesizes this file site target onto it. If there is one present and it's not file, it calls `error`. The final result is the same though, and the idea if that this check enforces that same requirement

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
hzsweers: I think either work :)

hzsweers:
Yeah I much prefer small bite-sized commits for readability :)

> I thought about other ways we can make it safer and still allow more control over the builder model, e.g. making Builders data class and leveraging copy(), but it might not be worth it.

Agreed on this bit. @swankjesse and I talked about it a fair bit and while copy() if convenient, you'd need to be able to access the current values somehow too. That leads to a question of maybe making a `toBuilder()` overload instead with default parameter values, but that limits it to just toBuilder() time and not just any builder. You could also have `replace`-ers, but again you'd need a way to expose the current value. That might not be terrible though if we used a higher order function to turn over a copy though?

```kotlin
fun replaceFunctions(body: (List<FunSpec>) -> List<FunSpec>) = apply {
  val newSpecs = body(funSpecs.toList()).toList()
  funSpecs.clear()
  addFunctions(newSpecs) // Get all the usual validations
}

// In use

builder.replaceFunSpecs { currentSpecs ->
  currentSpecs.filter { it.name == "functionToRemove" }
}
```

What do y'all think? Similar could be done with the JavaPoet PR (CC @ronshapiro)

> I noticed you didn't expose all mutable fields, e.g. CodeBlocker.Builders, which can also be modified, so how do we decide which fields to expose and which ones to leave private?

True. I kept the scope around collection types, but could see this being a useful one too.

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
hzsweers: 6755c8d

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
hzsweers: 0f36a8d

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
hzsweers: c1b5f71

src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
hzsweers: 9f851f0

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
hzsweers: 840f2cd

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
hzsweers: 8b3be8c

ronshapiro:
Could the visitor pattern be helpful here? A TypeVisitor could visit all of the aspects of the type, allowing for easy filtering (e.g. one method would accept the list of MethodSpecs and return a list of MethodSpecs, modified to their liking). For more complicated filtering, users could use a MethodVisitor for each method in the list, allowing further modification. Or something to that effect whereby there's lots of power in the visitor without sacrificing anything about the model's API.

ASM has something similar in their ClassVisitor/MethodVisitor API. And it seems to follow the javax.lang.model API style too.

hzsweers:
A visitor seems kind of overkill, and not everything is necessarily tied to a type (each builder is sort of standalone). I feel like the replace API hits a nice sweetspot of being simple without adding too much surface area and keeping the eager validation of before. Since the user provides the function, they can make that however complex or simple as they want and the library's none-the-wiser

swankjesse:
I don't like it!

hzsweers:
Reverted the PoC and rebased

PR #405, opened 2018-06-18
Move variance declaration back as it was before.

swankjesse:
We'd discussed both ways and decided that it was better as-is.


Egorand:
Yeah, let's keep it similar to `KTypeParameter`. @JakeWharton got any opinion on this?

PR #411, opened 2018-06-22
Add JavaPoet's docs on supported specs to README

Egorand:


README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
     .build()
 ```
 
+### Functions
+
+All of the above functions have a code body. Use `KModifier.ABSTRACT` to get a function without any
+body. This is only legal if the enclosing class is either abstract or an interface.
```
swankjesse: only legal if it is enclosed by an abstract class or interface?

otherwise we imply top-level functions don’t exist

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+    .build()
+```
+
+Now we're getting the following output:
+
+```kotlin
+class HelloWorld(private val greeting: kotlin.String)
+```
+
+Notice that KotlinPoet omits `{}` for classes with no bodies.
```
swankjesse: no bodies –> empty bodies

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+    .addProperty(android)
+    .addProperty("robot", String::class, KModifier.PRIVATE)
+    .build()
+```
+
+Which generates:
+
+```kotlin
+class HelloWorld {
+    private val android: kotlin.String
```
swankjesse: don’t we import kotlin.String ?

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+
+    PAPER
+}
+```
+
+Fancy enums are supported, where the enum values override methods or call a superclass constructor.
+Here's a comprehensive example:
+
+```kotlin
+val helloWorld = TypeSpec.enumBuilder("Roshambo")
```
swankjesse: nice

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+        .addStatement("%N.sortedWith(%L)", "strings", comparator)
+        .build())
+    .build()
+```
+
+This generates a method that contains a class that contains a method:
+
+```kotlin
+class HelloWorld {
+    fun sortByLength(strings: kotlin.collections.List<kotlin.String>) {
```
swankjesse: are our imports broken?

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+}
+```
+
+Use `AnnotationSpec.builder()` to set properties on annotations:
+
+```kotlin
+val logRecord = FunSpec.builder("recordEvent")
+    .addModifiers(KModifier.ABSTRACT)
+    .addAnnotation(AnnotationSpec.builder(Headers::class)
+        .addMember("accept = %S", "application/json; charset=utf-8")
```
swankjesse: should we change the API so that "accept" is one argument, and its value is another?

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+val headerList = ClassName("", "HeaderList")
+val header = ClassName("", "Header")
+val logRecord = FunSpec.builder("recordEvent")
+    .addModifiers(KModifier.ABSTRACT)
+    .addAnnotation(AnnotationSpec.builder(headerList)
+        .addMember("[\n%>%L,\n%L%<\n]", AnnotationSpec.builder(header)
+            .addMember("name = %S", "Accept")
+            .addMember("value = %S", "application/json; charset=utf-8")
+            .build(), AnnotationSpec.builder(header)
+            .addMember("name = %S", "User-Agent")
```
swankjesse: indentation feels off here

README.md
```diff
@@ -306,6 +306,510 @@ val byteToHex = FunSpec.builder("byteToHex")
...
+])
+abstract fun recordEvent(logRecord: LogRecord): LogReceipt
+```
+
+KotlinPoet supports use-site targets for annotations:
+
+```kotlin
+val utils = FileSpec.builder("com.example", "Utils")
+    .addAnnotation(AnnotationSpec.builder(JvmName::class)
+        .useSiteTarget(UseSiteTarget.FILE)
```
swankjesse: nice

README.md
Egorand: No, this output is produced by `TypeSpec.toString()`, rather than `FileSpec.toString()`. I can make a pass and strip all imports for better readability.

README.md
Egorand: That's what we had initially and decided to simplify: 455c3a67a8cba64ed7ad1f3296384e1ff593f187

PR #418, opened 2018-07-08
Add API for creating lambda expressions

### LINKED ISSUE #416: Generating multiline lambda function
``` 
val unknownFields = reader.decodeMessage { tag ->
                when (tag) {
                    1 -> name = ProtoAdapter.STRING.decode(reader)
                    2 -> id = ProtoAdapter.INT32.decode(reader)
                    3 -> email = ProtoAdapter.STRING.decode(reader)
                    4 -> phone.add(PhoneNumber.ADAPTER.decode(reader))
                    else -> UnkownFieldsBuilder.UNKNOWN_FIELD
                }
            }
```

What would be the best way to generate this ? Can we use a combination of begin / end control flow statements ?

Egorand:
Fixes #416 

cc @mc-lovin

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -354,8 +354,32 @@ class CodeBlock private constructor(
      *     Shouldn't contain braces or newline characters.
      */
     fun beginControlFlow(controlFlow: String, vararg args: Any?) = apply {
-      add(controlFlow + " {\n", *args)
+      add("$controlFlow {\n", *args)
       indent()
+    }
+
+    /**
+     * @param statement statement that precedes the lambda expression, e.g. a function call that
+     * takes the expression as an argument.
+     * @param parameters optional lambda expression parameter names.
+     */
+    fun beginLambdaExpression(statement: String, vararg parameters: String) =
+        beginLambdaExpression(CodeBlock.of(statement), *parameters)
+
+    /**
+     * @param statement statement that precedes the lambda expression, e.g. a function call that
+     * takes the expression as an argument.
+     * @param parameters optional lambda expression parameter names.
+     */
+    fun beginLambdaExpression(
+      statement: CodeBlock,
+      vararg parameters: String
+    ): Builder {
+      return apply {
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -354,10 +354,44 @@ class CodeBlock private constructor(
...
+     * @param controlFlow the control flow construct and its code, such as "if (foo == 5)".
+     *     Shouldn't contain braces or newline characters.
+     * @param parameterNames optional control flow parameter names. Allow producing expressions such
+     * as:
+     *
+     * ```
+     * list.forEachIndexed { index, value ->
+     * ```
+     */
+    fun beginParameterizedControlFlow(controlFlow: String, vararg parameterNames: String) =
```
JakeWharton: This seems like a weird name to describe what always seems to be a lambda. Is it ever not a lambda? I guess you could argue that it's control flow-like, but it seems like a stretch of the term. And then my other concern is that "parameterized" is a term that's usually associated with generics.

Egorand: Good point, the naming is off here. I can try changing it to `beginLambdaExpression` and tinker with param names to see what works better.

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -354,10 +354,34 @@ class CodeBlock private constructor(
...
+    fun beginLambdaExpression(statement: String, vararg parameters: String) =
+        beginLambdaExpression(CodeBlock.of(statement), *parameters)
+
+    /**
+     * @param statement statement that precedes the lambda expression, e.g. a function call that
+     * takes the expression as an argument.
+     * @param parameters optional lambda expression parameter names.
+     */
+    fun beginLambdaExpression(
+      statement: CodeBlock,
```
swankjesse: sort of a weird word because it’s not a statement

swankjesse: codeBlock

PR #428, opened 2018-07-20
Escape dollar signs.

### LINKED ISSUE #427: Dollar sign is not escaped in CodeBlock.
https://github.com/square/kotlinpoet/blob/3ba154fbcde528ddfc35bdf3f5658003318c3671/src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt#L198

I think this should generate an extra escape (`val b = \"string with a \\\$ dollar sign\"\n"`) for the dollar sign.
Use case in https://github.com/square/moshi/issues/602

NightlyNexus:
Escaping is needed to prevent making string templates ("\$stringLiteral") and is safe in other usages ("\$" is the same string as "$").

Closes #427.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,8 @@ class HelloWorld {
 }
 ```
 
+Note that "$" is also escaped. To make string templates, use literals with [%L](#%L-for-Literals).
+
 ### %T for Types
 
 KotlinPoet has rich built-in support for types, including automatic generation of `import`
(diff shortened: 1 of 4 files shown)


NightlyNexus:
use literal values?

Egorand:
@NightlyNexus can you please add a test for string interpolation? It's also worth mentioning this behavior in README.

NightlyNexus:
✔️
not sure where it'd be best in the README. Appended to `%S for Strings`, perhaps?

Egorand:
Yeah, the documentation should be in `%S for Strings`.

I'm a little unsure about using `%L` for string templates, it doesn't seem intuitive to me and will be hard to discover. It almost feels like a workaround for the fact that we escape `$` automatically.

NightlyNexus:
Yeah, probably better to have to escape manually here. Thanks!

PR #429, opened 2018-07-20
building TypeName for Array<primitive-type>

evgeny-pasynkov:
fix exception on building TypeName for arrays of primitive types, such as `Array<Int>`
Correctly distinguish two types: `IntArray` and `Array<Int>`

exception being fixed:
java.lang.IllegalArgumentException: no type arguments: kotlin.IntArray
	at com.squareup.kotlinpoet.ParameterizedTypeName.<init>(ParameterizedTypeName.kt:38)
	at com.squareup.kotlinpoet.ParameterizedTypeName$Companion.get$kotlinpoet(ParameterizedTypeName.kt:141)
	at com.squareup.kotlinpoet.ParameterizedTypeNames.asTypeName(ParameterizedTypeName.kt:170)

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
@@ -19,6 +19,11 @@ package com.squareup.kotlinpoet
 import com.google.common.truth.Truth.assertThat
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.plusParameter
 import org.junit.Test
+import java.io.Closeable
+import java.io.InputStream
+import kotlin.reflect.KTypeProjection
+import kotlin.reflect.KVariance
+import kotlin.reflect.full.createType
 
 class ParameterizedTypeNameTest {
   @Test fun classNamePlusParameter() {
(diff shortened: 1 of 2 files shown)


Egorand:
Trying to wrap my head around this: could you please share a failing test case for this change? Might make sense to open an issue that describes the problem.

evgeny-pasynkov:
I have provided the test case: just revert my fix and run the case.

The roots of the problem is that kotlin reflection counts Array<Int> as
IntArray, though underlying java types are different, and, which more
important, different tyoe parameters count

evgeny-pasynkov:
This statement is right for all primitive arrays - ByteArray, CharArray, LongArray, etc

evgeny-pasynkov:
Any chance to merge this?

src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
```diff
@@ -57,6 +62,26 @@ class ParameterizedTypeNameTest {
     assertThat(typeName.toString()).isEqualTo("java.util.List<java.lang.String>")
   }
 
+  @Test fun primitiveArray() {
+    assertThat(ByteArray::class.asTypeName().toString()).isEqualTo("kotlin.ByteArray")
```
swankjesse: BooleanArray?

PR #432, opened 2018-07-31
Add KDoc to ParameterSpec; collapse into parent KDoc

### LINKED ISSUE #431: Function and constructor params should have KDoc
KDoc comments added via a new `ParameterSpec.addKdoc()` function would be collapsed into the containing `FunSpec` KDoc, or all the way to the containing `TypeSpec` KDoc if the function is a primary constructor.

For example:

```kotlin
FunSpec.builder("foo")
    .addParameter(ParameterSpec.builder("string", String::class.asTypeName())
        .addKdoc("A string parameter.\n")
        .build())
    .build()
```

produces:

```kotlin
/**
 * @param string A string parameter.
 */
fun foo(string: String) {
}
```

and:

```kotlin
TypeSpec.classBuilder("Taco")
  .addKdoc("A taco.\n")
  .primaryConstructor(FunSpec.builder("foo")
    .addParameter(ParameterSpec.builder("string", String::class.asTypeName())
        .addKdoc("A string parameter.\n")
        .build())
    .build())
  .build()
```

produces:

```kotlin
/**
 * A taco.
 * @param string A string parameter.
 */
class Taco(string: String)
```

In the case of constructor properties, parameter KDoc should probably override property KDoc as `@param` comments are treated as property comments by Dokka.

tadfisher:
Fixes #431 

Adds `ParameterSpec.kdoc` and `ParameterSpec.Builder.addKdoc()`.

Parameter KDoc is appended to function KDoc as `@param` blocks when `emit(includeParametersKdoc = true)`.

If the primary constructor of a `TypeSpec` contains parameters or properties with non-empty `KDoc`, the `TypeSpec` emits those comments as `@param` blocks. If a parameter with KDoc corresponds to a property with KDoc, the parameter wins. `@param` is preferable as Dokka treats it as documentation for both the constructor parameter and type property.

When a `FunSpec` is a primary constructor, it is emitted with `includeParametersKdoc = false`.


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -133,6 +138,19 @@ class FunSpec private constructor(builder: Builder) {
...
   val isAccessor get() = name.isAccessor
 
+  private fun kdocWithParameters(): CodeBlock {
+    if (parameters.all { it.kdoc.isEmpty() }) {
+      return kdoc
+    }
+
+    return with (kdoc.toBuilder()) {
+      for (parameterSpec in parameters) {
+        add("@param %L %L", parameterSpec.name, parameterSpec.kdoc)
```
Egorand: Will this add `@param` lines even for parameters that don't have KDoc? I'd actually remove the `all` check at the top of the function body and do an `isEmpty` check for each parameter to skip those that don't have the KDoc.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -143,7 +161,7 @@ class FunSpec private constructor(builder: Builder) {
   override fun hashCode() = toString().hashCode()
 
   override fun toString() = buildString {
-    emit(CodeWriter(this), "Constructor", TypeSpec.Kind.Class().implicitFunctionModifiers)
+    emit(CodeWriter(this), "Constructor", TypeSpec.Kind.Class().implicitFunctionModifiers, true)
```
Egorand: Would be great to use named args for boolean values

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -292,6 +292,31 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
 
+  /** Returns KDoc comments including those of primary constructor parameters. */
+  private fun kdocWithConstructorParameters(): CodeBlock {
+    if (primaryConstructor == null || primaryConstructor.parameters.isEmpty()) {
+      return kdoc
+    }
+
+    val constructorProperties = constructorProperties()
+
+    if (primaryConstructor.parameters.all { it.kdoc.isEmpty() } &&
```
Egorand: I think this check is redundant, given that we check for empty KDoc in the snippet below. Is this to avoid doing the `toBuilder()`/`build()` on the KDoc in case there's nothing to add?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -158,6 +158,21 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionParamWithKdoc() {
```
Egorand: Can you add a test that has a param with KDoc and a param without the KDoc?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1122,6 +1122,57 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun kdocWithParameters() {
+    val taco = TypeSpec.classBuilder("Taco")
```
Egorand: I believe the indentation here is incorrect, we use 2 spaces

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1122,6 +1122,57 @@ class TypeSpecTest {
...
+            .addKdoc("A hard or soft tortilla, loosely folded and filled with whatever\n")
+            .addKdoc("[random][%T] tex-mex stuff we could find in the pantry\n", Random::class)
+            .addKdoc(CodeBlock.of("and some [%T] cheese.\n", String::class))
+            .primaryConstructor(FunSpec.constructorBuilder()
+                    .addParameter(ParameterSpec.builder("temperature", Double::class)
+                            .addKdoc(CodeBlock.of("%L", """
+                                |Taco temperature. Can be as cold as the famous ice tacos from
```
Egorand: Love this!

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -43,7 +43,9 @@ class PropertySpec private constructor(builder: Builder) {
     val isInlineProperty = getter?.modifiers?.contains(KModifier.INLINE) ?: false &&
         setter?.modifiers?.contains(KModifier.INLINE) ?: false
     val propertyModifiers = if (isInlineProperty) modifiers + KModifier.INLINE else modifiers
-    codeWriter.emitKdoc(kdoc)
+    if (!inline) {
```
swankjesse: Curious: why the inline check?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -292,6 +292,31 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+            constructorProperties.values.all { it.kdoc.isEmpty() }) {
+      return kdoc
+    }
+
+    return with (kdoc.toBuilder()) {
```
swankjesse: Nit: space after with.

src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
Egorand: `inline` will only be true when properties are used as primary constructor params, in that case there should be no new line after the property code and property annotations should be printed on a single line. The name is confusing though.

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
tadfisher: Correct, but it's probably premature optimization. I'll pull it out.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
tadfisher: Fixed, thanks.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
tadfisher: Added arg names, thanks.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
tadfisher: Fixed, thanks.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
tadfisher: I added a couple of additional parameters to this test, I hope that's OK.

PR #442, opened 2018-08-11
Add parens for annotated types

jbolinger:
Using a `LambdaTypeName` with a receiver type that has an annotation currently produces the following:

```
fun foo(init: @Marker MyType.() -> Unit)
```

I think it's useful to wrap it in parens so that the generated code is:

```
fun foo(init: (@Marker MyType).() -> Unit)
```

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/LambdaTypeName.kt
@@ -62,7 +62,11 @@ class LambdaTypeName internal constructor(
     }
 
     receiver?.let {
-      out.emitCode("%T.", it)
+      if (it.isAnnotated) {
+        out.emitCode("(%T).", it)
+      } else {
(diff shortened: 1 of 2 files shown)


Egorand:
Does this change have any practical implications, i.e. are there any scenarios where `@Marker MyType.() -> Unit` would not compile and `(@Marker MyType).() -> Unit` would? Or is this style recommended by official/Android Kotlin style guide? Otherwise, the fix feels subjective and one may argue that it doesn't improve code readability.

jbolinger:
Yes, for example if the `@Marker` is an annotation for a DslMarker the behavior is different with and without the parens. For a little more detail take a look at this thread: https://discuss.kotlinlang.org/t/scope-control-for-builder-like-dsls/2289/2

Egorand:
Gotcha, so it's basically a case of annotating the receiver vs annotating the lambda type, makes sense!

PR #445, opened 2018-08-21
Keep type-parameter variance when constructing TypeName from KType

alex2069:
Updates `ParameterizedTypeName` to preserve type-parameter variance when constructed via `KType.asTypeName()`.

Issue encountered when attempting to generate a property from another property's return type:
```kotlin
interface MyInterface {
    val field: KClass<out Annotation>
}
val typeName = MyInterface::field.returnType.asTypeName()
```

Expected: `kotlin.reflect.KClass<out kotlin.Annotation>`
But was getting without variance: `kotlin.reflect.KClass<kotlin.Annotation>`

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
@@ -24,6 +24,7 @@ import kotlin.reflect.KClass
 import kotlin.reflect.KType
 import kotlin.reflect.KTypeParameter
 import kotlin.reflect.KTypeProjection
+import kotlin.reflect.KVariance
 
 class ParameterizedTypeName internal constructor(
   private val enclosingType: TypeName?,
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
```diff
@@ -20,10 +20,7 @@ package com.squareup.kotlinpoet
 import java.lang.reflect.Modifier
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
-import kotlin.reflect.KClass
-import kotlin.reflect.KType
-import kotlin.reflect.KTypeParameter
-import kotlin.reflect.KTypeProjection
+import kotlin.reflect.*
```
JakeWharton: Please revert. We do not allow star imports.

src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
```diff
@@ -88,4 +89,24 @@ class ParameterizedTypeNameTest {
...
+    val outVariance: KClass<out Annotation>
+    val inVariance: KClass<in Test>
+    val invariantNullable: KClass<Test>?
+    val star: KClass<*>
+    val multiVariant: Map<in String, List<Map<KClass<out Number>, *>?>>
+  }
+
+  private fun assertKTypeProjections(kType: KType) = assertThat(kType.asTypeName().toString()).isEqualTo(kType.toString())
+
+  @Test fun kTypeOutProjection() = assertKTypeProjections(Projections::outVariance.returnType)
```
Egorand: Nice tests!

src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
alex2069: Fixed

Issue #439, opened 2018-08-08
%S doesn't escape string template characters and/or unable to escape template characters


alex2069:
Not sure if this could be considered a bug or not.

For the given function:
```kotlin
val stringFromUnknownSource = "Some string\n" + "$" + "annoyingUser" + " is annoying."
val getStringFunc = FunSpec.builder("getString").addStatement("return %S", stringFromUnknownSource).build()
```

Results in:
```kotlin
fun getString() = """
  |Some string
  |$annoyingUser is annoying.
  """.trimMargin()
```

Which fails to compile since the user or whatever was trying to be a smart ass (or perhaps they really did just make a mistake heh).

I would kind of expect the output of `return %S` to be a pure string and nothing more. Whereas if I wanted a string with side-effects (template string) I would use `return "%L"`? Not sure TBH as I suspect it might be intentional?

So I suppose the question is - where should the responsibility of escaping come into it? The documentation kind of implies it handles the escaping for you:
> `%S` escapes the value as a *string*, wraps it with double quotes, and emits that. For example, `6" sandwich` is emitted `"6\" sandwich"`.

Ironically however, when I tried to escape the `$` by adding a `\` in front of it, the `%S` then escapes that, thwarting my escaping attempts.

As a workaround I've had to replace all `%S` usages with `%L` and a pre-escaped string.
So the above turns into:
```kotlin
fun String.toKString() = "\"\"\"${replace("\$", "\${'\$'}")}\"\"\""
FunSpec.builder("getString").addStatement("return %L", stringFromUnknownSource.toKString()).build()
```

And Generates:
```kotlin
fun getString() = """Some string
${'$'}annoyingUser is annoying."""
```

-----

So just before submitting this I figured I should check for related issues and found #427 and #428 - so not sure about this now. However they are both closed and the consensus was that it didn't seem intuitive to use `%L` for templates (which I disagree with - I expected `%S` to be escaping). 

At the moment it's doing something in between (which is counter intuitive). This also means there are two ways to emit a templated string `%S` and `"%L"`, but no way to have a "pure" string - which I would argue is desired most of the time (also it appears `%L` with a multi-line string is altered by the code indentation/formatting).

If the behaviour remains unchanged then we can never reliably use `%S` from unknown sources (which I'd imagine is frequently the case; from user's annotation declarations or what have you).

swankjesse:
Should we have different escape sequences for strings with interpolation (`%P` for pattern ) vs. those without (`%S` for string) ?

[...]

PR #447, opened 2018-08-24
Add support for more literal types

alex2069:
This PR adds support for additional types when used with `%L` formatting:
1. If the type is a `Char`, `Float`, or `Long` types, the emitted code uses Kotlin literal notations
2. If the type is a primitive array type (not `Array<*>` but `PrimitiveTypeArray` such as `BooleanArray`), the emitted code uses the corresponding array type construction functions
3. If the type is a `List`, `Map`, or `Array`, then the emitted code uses `listOf()`, `mapOf()`, or `arrayOf()` respectively

----
### Commit 1 Reasoning
Special handling is needed for these primitive types otherwise - currently we need to explicitly check the type when we `addStatement` or whatever:
```kotlin
when (valueType.kind) {
    TypeKind.BOOLEAN,
    TypeKind.BYTE,
    TypeKind.SHORT,
    TypeKind.INT,
    TypeKind.DOUBLE -> func.addStatement("return %L", value)
    TypeKind.CHAR -> func.addStatement("return '%L'", value)
    TypeKind.LONG -> func.addStatement("return %LL", value)
    TypeKind.FLOAT -> func.addStatement("return %Lf", value)
}
```
`Short` and `Byte` type information is still not preserved however as there is no literal notation for them (potentially they would need `.toShort()` and `.toByte()` for them in emitted code).

### Commits 2 & 3 Reasoning
These are more for convenience. Example use-case:

For the given code:
```kotlin
interface SomeType
interface Interface {
    val list: List<SomeType>
}

// some iterable of unknown length/types
val iterable: Collection<TypeSpec> = listOf(
    TypeSpec.anonymousClassBuilder().addSuperinterface(SomeType::class).build(),
    TypeSpec.anonymousClassBuilder().addSuperinterface(SomeType::class).build(),
    TypeSpec.anonymousClassBuilder().addSuperinterface(SomeType::class).build()
)
val p = PropertySpec.builder("list", List::class.parameterizedBy(SomeType::class), KModifier.OVERRIDE)
    .initializer("listOf<%T>(%L)", SomeType::class, iterable)
    .build()
```

Generates:
```kotlin
override val list: kotlin.collections.List<SomeType> = listOf<SomeType>([object : SomeType {
}, object : SomeType {
}, object : SomeType {
}])
```
Which obviously fails since list/array notation `[...]` are not supported outside of annotations.

My first thought was to use `iterable.joinToString()`, but then I realised that stops KotlinPoet from correctly identifying type usage and thus it's unable to generate the correct imports etc (i.e. it works fine except that everything is fully qualified and I suspect it also interferes with subsequent indenting/formatting or whatever).

At the moment the closest/simplest I can get is something like:
```kotlin
inline fun <reified T : Any> Collection<*>.toListOf() =
    CodeBlock.of("listOf<%T>(${",%L".repeat(size).drop(1)})", T::class, *toTypedArray())

val p = PropertySpec.builder("list", List::class.parameterizedBy(SomeType::class), KModifier.OVERRIDE)
    .initializer(iterable.toListOf<SomeType>())
    .build()
```

Incidentally however, while writing tests I discovered an alternative `CodeBlock.joinToCode()` - however this still feels rather clunky requiring you `.map` the entire list to `CodeBlock.of("%L", it)` everywhere you want to use a list/map/array.


swankjesse:
I'm reluctant to have `%L` do anything other than toString() its input. It would be a big behavior change from what we do today.

The word _literal_ might have been misleading; we don’t emit a string literal for strings; we emit the value of the string.

swankjesse:
It might be a mistake for KotlinPoet to attempt to cover all sorts of types here. It’s simpler for you to write straightforward extensions that you need, and we don’t have to document a bunch of special behavior.

Some examples I worry about:
 * Should a list of strings emit the contents of those strings (which might be variable names) or quoted string literals?
 * What if we want to support multimaps in the future. Does that change behavior of existing code?

Egorand:
Should we just limit this to primitives?

swankjesse:
Even with primitives we’d be losing our ability to emit a Long without a trailing `L`.

I think maybe what we want to recommend is users just create extension functions to turn values into source code:
```
val String?.kt = if (this != null) CodeBlock.of("%S", this) else CodeBlock.of("null")
val Float?.kt = if (this != null) CodeBlock.of("%Lf", this) else CodeBlock.of("null")
val Long?.kt = if (this != null) CodeBlock.of("%LL", this) else CodeBlock.of("null")
```

This is pretty simple & should compose pretty well.

alex2069:
(sorry for the long comment - been thinking about resolutions to this for a while now)

> - Should a list of strings emit the contents of those strings (which might be variable names) or quoted string literals?

I would be extremely surprised if it didn't just output a list of normal strings. I can't imagine a scenario where you'd want to output a list of executable strings - if that was desired it'd be the significant minority for sure.


> - What if we want to support multimaps in the future. Does that change behaviour of existing code?

It doesn't really matter as you *can't* at the moment. If you try to `%L` anything but a `Boolean`, `Int`, `Double`, or KotlinPoet type, that output is effectively useless or unsafe and isn't guaranteed to compile - unless you're using it within a string (and hope it doesn't contain a `$word`) or it has a custom `.toString()` intended for `%L`. Other exceptions are context sensitive (which is just a nightmare); passing in a `Float` *might* work if the language supports omitting the `f` in that context, and similarly for the other primitives. So adding support for future types is unlikely to affect many as you can't really pass them in reliably at the moment anyway.


> I'm reluctant to have `%L` do anything other than toString() its input. It would be a big behavior change from what we do today.
>
> The word _literal_ might have been misleading; we don’t emit a string literal for strings; we emit the value of the string.

But that's kind of the source of the problem though - `%L` is relying on some underlying behaviour of a type to represent itself in a meaningful way. Sometimes it'll compile and sometimes it wont - at the moment the exact same `%L` statement may compile in one place but fail in another depending on the types declared and context of use (an experienced source of frustration).

I would almost argue it should reject anything but supported types and require a different formatting char for "default" `.toString()` (or force you to call `.toString()` manually) - emitting its `.toString()` and hoping it works feels like magic;
- Some weird custom type may just coincidentally work one day, but fail another when a library updates
- Perhaps the end-user encounters some edge case on a types `.toString()` representation or whatever
- If you're dealing with interfaces or open classes it'll likely break when the user overrides the `.toString()`
- What happens when inline classes become a thing? I believe they support overriding the `.toString()`?

Granted some of these are contrived/naive implementations - but at the moment that's enabled by the current `%L` behaviour.


> I think maybe what we want to recommend is users just create extension functions to turn values into source code:
>
> ```kotlin
> val String?.kt = if (this != null) CodeBlock.of("%S", this) else CodeBlock.of("null")
> val Float?.kt = if (this != null) CodeBlock.of("%Lf", this) else CodeBlock.of("null")
> val Long?.kt = if (this != null) CodeBlock.of("%LL", this) else CodeBlock.of("null")
> ```
> This is pretty simple & should compose pretty well.

I have found myself adding in primitive/string type checks all over the place to ensure the correct output is produced, as sometimes you just don't know the type at run-time so you can't use typed extension functions like this (e.g. the likely very frequently used `AnnotationValue.getValue()` is an `Object`). There really needs to be a simple user-friendly way to do this. Expecting every user to add extension functions for something so fundamental seems crazy to me.


> Even with primitives we’d be losing our ability to emit a Long without a trailing `L`.

I'd imagine the majority of the time you'd want the type information kept. If you really wanted the `.toString()` of a value then include it as part of the template and/or `.toString()` it and (in the current implementation) pass it in as `%L`. To me `%L` should be the *literal* value - not the `.toString()` of a value - it wasn't until I looked into the code (due to this whole issue) that I discovered otherwise. Prior to that I had been using a mix of string templates and formatting chars; `CodeBlock.of("val foo = arrayOfNulls<%T>($size)", someType)` - it came as a surprise that it just "strings" things and hopes for the best.

In fact the more I think about it `%L` is anything but literal lol. The only types that are supported are `Boolean`, `Int`, `Double`, and KotlinPoet types. Everything else is an approximation at best, and a compile error at worst.

If I were to do it all from scratch I would have it as something like:
- `%L` for literal types;
    - KotlinPoet types
    - enums
    - classes `Type::class`
    - strings (fully-escaped)
    - primitives (using mix of literal notation/`.toShort()`/`.toByte()`/etc.)
    - primitive arrays (there's no type ambiguity here and they'd be useful from an annotation's values)
    - arrays (I believe these can be safely implemented as there's little type ambiguity here)
    - potentially a subset of common iterables (this could be iffy though as mentioned above - though most use-cases would just want `listOf()` and `mapOf()` - if more advanced types were desired they would expectedly require special handling by the user)
    - unsupported types throwing an exception
    - potentially a pluggable system to allow the user to add their own special/reusable handling across projects
- `%S` (or `%V` or something) for the `.toString()` value of things
- `%E` (or whatever) for executable strings

So, having said all that lol - unfortunately given it is effectively 1.0.0 already I suspect there is a desire to avoid breaking changes (though this really needs to be resolved sooner rather than later given there's no "typing" with format strings making it hard to change later, and that `%L` for "literal types" is a misnomer at the moment).

As a potential non-breaking change; `%V` for the "Value" of a type (performing `%L` listed above) - though as mentioned above this feels a little like the `.toString()` value though, but can't be helped if `%L` is to remain unchanged.

Since annotations support a small set of well defined types, at the very least these should be supported natively (i.e. anything that can be returned from `AnnotationValue.getValue()`).

alex2069:
So I had a very brief look as to whether it could even be considered to alter `%L` to such a degree and suffice to say I don't think it'll be a trivial change.
Instead I've added the proposed `%V` formatting char and restored `%L` to its original behaviour. Incidentally this also helps with #439 in that strings are fully escaped, which would also allow `%S` to remain unchanged.

PR #449, opened 2018-08-30
Default TypeVariable bounds lists to nullable any on empty inputs

### LINKED ISSUE #420: Allow creating TypeVariableName with no bounds and default to Any?
Currently 

```TypeVariableName("T", bounds = *emptyArray<TypeName>())```

 fails with 

```java.lang.IllegalArgumentException: T has no bounds``` 

An option would be to not crash but to default bounds to `Any?`.

### PR DESCRIPTION
**Pull request by ZacSweers**
Resolves #420

Wasn't quite sure how to best design this, hope the property extraction looks alright

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -771,8 +771,10 @@ class TypeVariableName private constructor(
           bounds += TypeName.get(typeMirror, typeVariables)
         }
         bounds.remove(ANY)
-        bounds.remove(NULLABLE_ANY)
         bounds.remove(JAVA_OBJECT)
+        if (bounds.isEmpty()) {
+          bounds.add(NULLABLE_ANY)
+        }
       }
       return typeVariableName
     }
@@ -792,8 +794,10 @@ class TypeVariableName private constructor(
           bounds += TypeName.get(bound, map)
         }
         bounds.remove(ANY)
-        bounds.remove(NULLABLE_ANY)
         bounds.remove(JAVA_OBJECT)
+        if (bounds.isEmpty()) {
+          bounds.add(NULLABLE_ANY)
+        }
       }
       return result
     }
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -713,22 +713,38 @@ class TypeVariableName private constructor(
     /** Returns type variable named `name` with `variance` and without bounds.  */
     @JvmStatic @JvmName("get") @JvmOverloads
     operator fun invoke(name: String, variance: KModifier? = null) =
-        TypeVariableName.of(name, listOf(NULLABLE_ANY), variance)
+        TypeVariableName.of(
+            name = name,
+            bounds = NULLABLE_ANY_LIST,
+            variance = variance
+        )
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt [hunk 1]
```diff
@@ -27,13 +27,15 @@ import kotlin.reflect.KVariance
...
   val name: String,
-  val bounds: List<TypeName>,
+  inputBounds: List<TypeName>,
   val variance: KModifier? = null,
   val reified: Boolean = false,
   nullable: Boolean = false,
   annotations: List<AnnotationSpec> = emptyList()
 ) : TypeName(nullable, annotations) {
 
+  val bounds = if (inputBounds.isEmpty()) listOf(NULLABLE_ANY) else inputBounds
```
swankjesse: can you push this out to the calling code? As far as I know only `TypeVariableName.Companion#of` actually benefits from this behavior.

And if you push it into that method, you can just make `listOf(NULLABLE_ANY)` the default value of the parameter. While you’re there you could make variance default to null so calling this thing is super easy.

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -66,10 +68,6 @@ class TypeVariableName private constructor(
       require(variance == null || variance.isOneOf(KModifier.IN, KModifier.OUT)) {
         "$variance is an invalid variance modifier, the only allowed values are in and out!"
       }
-      require(bounds.isNotEmpty()) {
```
swankjesse: ... and if you push it out, you can restore this check.

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
hzsweers: can do. Originally did this because I saw that this consutructor was used in a few methods in TypeVariableName, but after looking again there's no danger of them passing empty lists

hzsweers: Couple other questions

- should `variance` be moved to before the `bounds` arg? Feels weird to have a vararg in the middle of the params, and now they'd both have default values? Going to do that in my update but can change if needed
- what about the overloads that accept `Type` and `KClass`? Don't specify that arg if they're empty?

src/main/java/com/squareup/kotlinpoet/TypeVariableName.kt
hzsweers: I don't think so. The awkward part in moshi was that we were always passing in a bounds list, it just might sometimes be empty. In this case, that would still fail this check every time, whereas I thought the goal was to check for empty inputs and coerce them to a `listOf(NULLABLE_ANY)` 🤔

Egorand:
@hzsweers @swankjesse let's decide on how we proceed with this

ZacSweers:
Rebased and rehashed this. This moves the bounds handling back out to the `of()` or otherwise `internal` creators and restores the empty bounds check. 

This also adds a test for `TypeVariableName.get()`, which would accept an empty type variable before and convert it to the not-so-idiomatic `<T : java.lang.Object>` along the way. This addresses that in 39551ef, but happy to revert if we don't want that. CC @JakeWharton

ZacSweers:
@swankjesse want to take another look? I think I've covered your comments

ZacSweers:
I've rebased this, still think it's worth getting in

Egorand:
@swankjesse can you PTAL?

PR #453, opened 2018-09-01
Add note on default value to readme

### LINKED ISSUE #451: How to create a default value for fun parameter
How would I generate something like this?

```
fun someFunction(someValue:String?="defaultValue"){

}
```

shaishavgandhi05:
Resolves #451

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,33 @@ class HelloWorld {
 Note that due to a [bug](https://youtrack.jetbrains.com/issue/KT-15286), the IDE will not autocomplete the `parameterizedBy` or `plusParameter` extensions
 and you'll have to add the import statement manually to get those extensions.
 
+#### Nullable Types
+
+KotlinPoet also supports Kotlin's nullable types. Simply add the extension `asNullable()` to any type to make it nullable. For example:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -499,7 +499,8 @@ Declare parameters on methods and constructors with either `ParameterSpec.builde
 `FunSpec`'s convenient `addParameter()` API:
 
 ```kotlin
-val android = ParameterSpec.builder("android", String::class)
+val android = ParameterSpec.builder("android", String::class.asTypeName().asNullable())
```
JakeWharton: Why make it nullable? For the readme I'd prefer to avoid the noise considering it doesn't change the behavior.

shaishavgandhi05: That was in response to an issue asking about how to add nullable parameters. Since there isn't an example in the readme for nullable parameters, I thought this might be a good place to exhibit both the default value and nullable. 

I'm open to adding it elsewhere.

swankjesse: Please do. Just a short description & a fun example!

shaishavgandhi05:
Anything holding this up?

PR #456, opened 2018-09-01
allow sealed classes to declare abstract functions

### LINKED ISSUE #455: Sealed classes should be allowed to declare abstract functions
e.g., `TypeSpec.Kind` and `plusModifiers()`

and yet `TypeSpec.Builder#build()` restricts it to only classes marked as `abstract`.

evanchooly:
fixes #455

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -23,6 +23,7 @@ import com.squareup.kotlinpoet.KModifier.EXPECT
 import com.squareup.kotlinpoet.KModifier.EXTERNAL
 import com.squareup.kotlinpoet.KModifier.PUBLIC
 import com.squareup.kotlinpoet.KModifier.SEALED
+import com.squareup.kotlinpoet.KModifier.Target.PROPERTY
 import com.squareup.kotlinpoet.TypeSpec.Kind.Interface
 import com.squareup.kotlinpoet.TypeSpec.Kind.Object
 import java.lang.reflect.Type
@@ -287,7 +288,15 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
       if (CodeBlock.of("%N", parameter) != property.initializer) continue
       result[property.name] = property.toBuilder()
           .addAnnotations(parameter.annotations)
-          .addModifiers(*parameter.modifiers.toTypedArray())
+          .addModifiers(*parameter.modifiers
+                  .filter { try {
+                      it.checkTarget(PROPERTY)
+                      true
+                    } catch(e: IllegalArgumentException) {
+                    false
+                    }
+                  }
+                  .toTypedArray())
           .build()
     }
     return result
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -22,6 +22,7 @@ import com.squareup.kotlinpoet.KModifier.ENUM
 import com.squareup.kotlinpoet.KModifier.EXPECT
 import com.squareup.kotlinpoet.KModifier.EXTERNAL
 import com.squareup.kotlinpoet.KModifier.PUBLIC
+import com.squareup.kotlinpoet.KModifier.SEALED
 import com.squareup.kotlinpoet.TypeSpec.Kind.Interface
 import com.squareup.kotlinpoet.TypeSpec.Kind.Object
 import java.lang.reflect.Type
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -627,7 +628,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
       val isAbstract = ABSTRACT in kind.modifiers || kind !is Kind.Class || !isSimpleClass
       for (funSpec in funSpecs) {
-        require(isAbstract || ABSTRACT !in funSpec.modifiers) {
+        require(isAbstract || ABSTRACT !in funSpec.modifiers || SEALED !in funSpec.modifiers) {
```
Egorand: This is incorrect, you're basically checking that none of the functions inside this `TypeSpec` have `sealed` modifier. Since `sealed` doesn't make sense for functions - the check will always succeed. Instead, you should be treating a `sealed` class as abstract by modifying `val isAbstract` above:

```
val isAbstract = ABSTRACT in kind.modifiers || SEALED in kind.modifiers || kind !is Kind.Class || !isSimpleClass
```

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -510,6 +510,22 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun sealedClassesMayDefineAbstractFunctions() {
+    val sld = TypeSpec.classBuilder("Sealed")
```
Egorand: `sld` is cryptic, how about `sealedClass`?

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
evanchooly: doh, you're right.  will correct that.

evanchooly: It's odd that the test passed with that wonky check, though.  I'd looked at isAbstract earlier but i clearly got some wires crossed because i was looking at a field on the class instead.

JakeWharton:
Does this work for vals too?

evanchooly:
I did *not* address abstract properties.  I didn't see where those were handled, specifically, but I can take another look for it.

evanchooly:
I just updated the test to add an abstract property to the test to ensure that works.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -510,6 +510,27 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun sealedClassesMayDefineAbstractMembers() {
+    val sealedClass = TypeSpec.classBuilder("Sealed")
+        .addModifiers(KModifier.SEALED)
+            .addProperty(PropertySpec.builder("name", ClassName("", "String"))
```
Egorand: Please fix the indentation in this part. Also, `ClassName("", "String")` -> `String::class`

evanchooly:
Actually, if I use String::class, i get an import for `kotlin.String` which is not what I want really.  I `ClassName("", "String") results in a cleaner emitted code.  So now that test fails because it doesn't expect the import.  Which way do you want it?

JakeWharton:
`String::class`

Egorand:
I'd still prefer `String::class` and the import in the expected code

evanchooly:
Fair enough.  Will correct that.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -510,6 +509,26 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun sealedClassesMayDefineAbstractMembers() {
+    val sealedClass = TypeSpec.classBuilder("Sealed")
+        .addModifiers(KModifier.SEALED)
+        .addProperty(PropertySpec.builder("name", String::class)
+                             .addModifiers(ABSTRACT).build())
```
Egorand: The indentation here is broken again..

evanchooly: I meant to ask about that.  Is there a `kotlinpoet` configuration I can import?  that formatting is done by the (iirc) default formatter in IDEA.

Egorand: Yes, https://github.com/square/java-code-styles. It's been updated to format Kotlin too.

PR #461, opened 2018-09-04
Replace TypeName with ClassName in AnnotationSpec

### LINKED ISSUE #460: AnnotationSpec.Builder with TypeName
I noticed that in it's private constructor, `AnnotationSpec` accepts a `TypeName`. However, it's builder doesn't accept a `TypeName` but instead needs a `ClassName` which extends `TypeName`. I believe JavaPoet also has something similar. Why not just expose `TypeName` in the builder? 

My use-case:
I have a project which already generates Java code via JavaPoet. I'm trying to make the API kotlin friendly by generating extension functions, which means I'm usually interoping between JavaPoet and KotlinPoet. So when I'm converting `JavaAnnotationSpec` to `KotlinAnnotationSpec`(renamed via typealias), I could have done 
```kotlin
javaAnnotation.map { KotlinAnnotationSpec.builder(it.type) }
``` 
instead of having to construct the `KotlinClassName` again and feed it to the `KotlinAnnotationSpec.builder`

shaishavgandhi05:
Following up on discussion in #460 , I've moved AnnotationSpec to internally also use `ClassName` instead of `TypeName`. 

Not sure if we still want to keep the variable name as `type`, but I've changed it to `className`. 

Closes #460

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -27,7 +27,7 @@ import kotlin.reflect.KClass
 
 /** A generated annotation on a declaration.  */
 class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
-  val type: TypeName = builder.type
+  val className: ClassName = builder.className
   val members = builder.members.toImmutableList()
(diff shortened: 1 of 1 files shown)


JakeWharton:
Thanks!

PR #470, opened 2018-09-11
Allow using CodeBlocks with statements as property initializers and default parameter values

### LINKED ISSUE #462: Cannot set a property initializer to a CodeBlock with a (lambda) statement
Trying to generate a property with a lambda initializer:
```kotlin
val property: SomeTypeAlias = { arg ->
    println("arg=$arg")
}
```

Using:
```kotlin
val param = ParameterSpec.builder("arg", ANY).build()
val initializer = CodeBlock.builder()
    .beginControlFlow("{ %L ->", param)
    .addStatement("println(\"arg=\$%N\")", param)
    .endControlFlow()
    .build()
println("initializer=\n$initializer")

val lambdaTypeName = ClassName.bestGuess("com.example.SomeTypeAlias")
val property = PropertySpec.builder("property", lambdaTypeName)
    .initializer(initializer)
    .build()
println("property=\n$property")
```

The code block output is correct:
```kotlin
{ arg: kotlin.Any ->
    println("arg=$arg")
}
```

But when it comes time to output the property I get: `Unexpected error: java.lang.IllegalStateException: statement enter %[ followed by statement enter %[`

I can change the `addStatement` to just `add` and it works, but it results in the closing bracket being on the same line as the last line `println("arg=$arg")}`

Tried removing the lambda part with just the `println`:
```kotlin
CodeBlock.builder()
    .addStatement("println(\"hello world\")", param)
    .build()
```
Which also fails with the same error (though not sure if the above is "legal" - in fact not sure if the lambda declaration is the correct way to do it either TBH).

Egorand:
Fixes #462

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -142,6 +142,8 @@ class CodeBlock private constructor(
   internal fun replaceAll(oldValue: String, newValue: String) =
       CodeBlock(formatParts.map { it.replace(oldValue, newValue) }, args)
 
+  internal fun hasStatements() = formatParts.any { "%[" in it }
+
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
     if (other == null) return false
--- a/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -41,7 +41,7 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
 
   internal fun emitDefaultValue(codeWriter: CodeWriter) {
     if (defaultValue != null) {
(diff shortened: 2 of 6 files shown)


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -142,6 +142,13 @@ class CodeBlock private constructor(
   internal fun replaceAll(oldValue: String, newValue: String) =
       CodeBlock(formatParts.map { it.replace(oldValue, newValue) }, args)
 
+  internal fun hasStatements(): Boolean {
+    for (formatPart in formatParts) {
```
JakeWharton: `any`

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -142,6 +142,13 @@ class CodeBlock private constructor(
...
+      if (formatPart.contains("%[")) return true
```
JakeWharton: `in`

Issue #476, opened 2018-09-14
Escape parameter name in PropertySpec which is also a Kotlin keyword


emrul:
If I have a piece of code like this:
```
PropertySpec.builder("object", ...)
```
I get a "not a valid name: object" error.

This is similar to #263 #293 - could we have KotlinPoet automatically escape these so the name of the generated property becomes, for example \`object\`?

Egorand:
Sure! Wanna submit a PR?

emrul:
Of course, was quite simple since the escaping was already being done in code emitter so just removed the checks and added a couple of tests to verify.

PR #477, opened 2018-09-15
Escape property and variable names if keywords

emrul:
PR for #476

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -237,7 +237,6 @@ class PropertySpec private constructor(builder: Builder) {
 
   companion object {
     @JvmStatic fun builder(name: String, type: TypeName, vararg modifiers: KModifier): Builder {
-      require(name.isName) { "not a valid name: $name" }
       return Builder(name, type)
           .addModifiers(*modifiers)
     }
@@ -249,7 +248,6 @@ class PropertySpec private constructor(builder: Builder) {
         = builder(name, type.asTypeName(), *modifiers)
 
     @JvmStatic fun varBuilder(name: String, type: TypeName, vararg modifiers: KModifier): Builder {
-      require(name.isName) { "not a valid name: $name" }
       return Builder(name, type)
           .mutable(true)
           .addModifiers(*modifiers)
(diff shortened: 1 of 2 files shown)


cypressious:
Kinda blocked by this

emrul:
@Egorand done

Egorand:
Thanks!

cypressious:
@Egorand Would it be possible to merge this so that I can depend on the snapshot?

cypressious:
Oh, I just noticed a bug:

I generate a type like so:

```kotlin
val typeBuilder = TypeSpec.classBuilder(name)
    .addProperties(properties.map {
        PropertySpec
            .varBuilder(it.key, parameterType(it.key, it.value))
            .initializer(it.key)
            .build()
    })
        .addParameters(
            properties.map {
                ParameterSpec
                    .builder(it.key, parameterType(it.key, it.value))
                    .apply { if (it.value.optional) defaultValue("null") }
                    .build()
            }
        )
        .build())
```

and get the following result

```kotlin
class AlarmInfo(
    `when`: Float? = null,
    var delayInMinutes: Float? = null,
    var periodInMinutes: Float? = null
) {
    var `when`: Float? = when
}
```

cypressious:
Maybe not a bug after all. If I manually escape the `.initializer(it.key)` part, it works. Still, not optimal.

emrul:
@cypressious I had a look at this - I agree it is not optimal.  However, 'initialiser' passes off to CodeBlock and it seems automatic escaping is not carried out there.

I'm fairly new to the KotlinPoet internals but I can't see an effective way to auto-escape keywords within codeblocks; I think I'd have to leave that to the more experienced contributors.

Egorand:
@cypressious can you file that please? Should be fixed in a follow-up PR

cypressious:
Filed #483

Egorand:
@emrul Sorry for keeping this open for a while. Can you please rebase and fix conflicts (we removed `varBuilder()`)?

emrul:
@Egorand @shaishavgandhi05 apologies for the silence - I'm on the road for the next few weeks with very limited capability.  Hoping someone else can pick this up since it should remain a very simple change to rebase.

shaishavgandhi05:
No problem! I can pick this up.

Egorand:
Thanks @emrul!

PR #478, opened 2018-09-16
Add more tests for LambdaTypeName

### LINKED ISSUE #475: Add tests for LambdaTypeNames that return LambdaTypeNames
E.g. we want to generate this: `(Int) -> ((Int) -> Unit)`

Egorand:
Closes #475

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/LambdaTypeNameTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/LambdaTypeNameTest.kt
@@ -85,7 +85,7 @@ class LambdaTypeNameTest {
         parameters = *arrayOf(Int::class.asTypeName()),
         returnType = returnTypeName)
     assertThat(typeName.toString())
-        .isEqualTo("(kotlin.Int) -> (kotlin.Int) -> kotlin.Unit")
+        .isEqualTo("(kotlin.Int) -> ((kotlin.Int) -> kotlin.Unit)")
   }
 
   @Test fun lambdaParameterType() {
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/LambdaTypeNameTest.kt
```diff
@@ -76,4 +76,26 @@ class LambdaTypeNameTest {
...
+
+  @Test fun lambdaReturnType() {
+    val returnTypeName = LambdaTypeName.get(
+        parameters = *arrayOf(Int::class.asTypeName()),
+        returnType = Unit::class.asTypeName())
+    val typeName = LambdaTypeName.get(
+        parameters = *arrayOf(Int::class.asTypeName()),
+        returnType = returnTypeName)
+    assertThat(typeName.toString())
+        .isEqualTo("(kotlin.Int) -> (kotlin.Int) -> kotlin.Unit")
```
Egorand: Actually, since "the arrow notation is right-associative", `(Int) -> (Int) -> Unit` is equivalent to `(Int) -> ((Int) -> Unit)`

JakeWharton: I'd feel more comfortable if we emit parenthesis although I can't imagine this occurs anywhere in the wild...

Egorand: I'll add this

PR #479, opened 2018-09-16
Add support for anonymous functions

### LINKED ISSUE #474: Add support for anonymous functions
https://kotlinlang.org/docs/reference/lambdas.html#anonymous-functions

Egorand:
Closes #474

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -71,7 +71,9 @@ class FunSpec private constructor(builder: Builder) {
     codeWriter.emitAnnotations(annotations, false)
     codeWriter.emitModifiers(modifiers, implicitModifiers)
 
-    if (!isConstructor && !name.isAccessor) {
+    if (name == null) {
+      codeWriter.emit("fun")
+    } else if (!isConstructor && !name.isAccessor) {
       codeWriter.emit("fun ")
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -152,6 +153,8 @@ class ParameterSpec private constructor(builder: ParameterSpec.Builder) {
     @JvmStatic fun unnamed(type: Type) = unnamed(type.asTypeName())
 
     @JvmStatic fun unnamed(type: TypeName) = Builder("", type).build()
+
+    @JvmStatic fun untyped(name: String) = Builder(name, null).build()
```
JakeWharton: I'm a bit skeptical about needing to support this. Can we leave it out and see if anyone ever complains?

Egorand: Sure

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -366,8 +370,8 @@ class FunSpec private constructor(builder: Builder) {
     internal const val GETTER = "get()"
     internal const val SETTER = "set()"
 
-    internal val String.isConstructor get() = this == CONSTRUCTOR
-    internal val String.isAccessor get() = this.isOneOf(GETTER, SETTER)
+    internal val String?.isConstructor get() = this == CONSTRUCTOR
```
swankjesse: `===` ?

Egorand: Good call!

Egorand: ..although technically this is what `String.equals()` already does, but semantically I like `===` better

PR #489, opened 2018-10-02
Rename NameAllocator.clone() to NameAllocator.copy()

swankjesse:
Clone is not very Kotlin


src/main/java/com/squareup/kotlinpoet/NameAllocator.kt
```diff
@@ -92,15 +92,15 @@ class NameAllocator private constructor(
...
+      tagToName[tag] = replaced // Put things back as they were!
       throw IllegalArgumentException("tag $tag cannot be used for both '$replaced' and '$result'")
     }
 
     return result
   }
 
   /** Retrieve a name created with [NameAllocator.newName]. */
-  fun get(tag: Any): String {
+  operator fun get(tag: Any): String {
```
Egorand: Sweet

PR #491, opened 2018-10-03
Remove PropertySpec.varBuilder(). It is asymmetric.

swankjesse:


src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -148,6 +148,7 @@ class PropertySpec private constructor(builder: Builder) {
     val modifiers = mutableListOf<KModifier>()
     val typeVariables = mutableListOf<TypeVariableName>()
 
+    /** True to create a `var` instead of a `val`. */
     fun mutable(mutable: Boolean) = apply {
```
Egorand: Btw should we make `true` the default arg? So you would just say `mutable()` instead of `mutable(true)`.

swankjesse: Great idea. Fixed.

PR #495, opened 2018-10-15
Add overloads to add kdoc to return type

### LINKED ISSUE #471: addKDoc overload for `returns`
I saw this suggested on JavaPoet. This would be a nice addition. Something like `returns(String::class, kdoc)`

shaishavgandhi05:
Resolves #471

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -62,9 +62,9 @@ class FunSpec private constructor(builder: Builder) {
     codeWriter: CodeWriter,
     enclosingName: String?,
     implicitModifiers: Set<KModifier>,
-    includeParametersKdoc: Boolean
+    includeExtrasKdoc: Boolean = false
   ) {
-    if (includeParametersKdoc) {
+    if (includeExtrasKdoc) {
       codeWriter.emitKdoc(kdocWithExtras())
     } else {
       codeWriter.emitKdoc(kdoc)
@@ -165,7 +165,7 @@ class FunSpec private constructor(builder: Builder) {
         codeWriter = CodeWriter(this),
         enclosingName = "Constructor",
         implicitModifiers = TypeSpec.Kind.Class().implicitFunctionModifiers,
-        includeParametersKdoc = true)
+        includeExtrasKdoc = true)
   }
 
   fun toBuilder(): Builder {
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -33,6 +33,7 @@ import kotlin.reflect.KClass
 class FunSpec private constructor(builder: Builder) {
   val name = builder.name
   val kdoc = builder.kdoc.build()
+  val returnKdoc = builder.returnKdoc.build()
   val annotations = builder.annotations.toImmutableList()
   val modifiers = builder.modifiers.toImmutableSet()
   val typeVariables = builder.typeVariables.toImmutableList()
@@ -64,7 +65,7 @@ class FunSpec private constructor(builder: Builder) {
     includeParametersKdoc: Boolean
   ) {
     if (includeParametersKdoc) {
-      codeWriter.emitKdoc(kdocWithParameters())
+      codeWriter.emitKdoc(kdocWithExtras())
     } else {
       codeWriter.emitKdoc(kdoc)
     }
@@ -138,13 +139,14 @@ class FunSpec private constructor(builder: Builder) {
 
   val isAccessor get() = name.isAccessor
 
-  private fun kdocWithParameters(): CodeBlock {
+  private fun kdocWithExtras(): CodeBlock {
     return with(kdoc.toBuilder()) {
       for (parameterSpec in parameters) {
         if (parameterSpec.kdoc.isNotEmpty()) {
           add("@param %L %L", parameterSpec.name, parameterSpec.kdoc)
         }
       }
+      add(returnKdoc)
       build()
     }
   }
@@ -183,6 +185,7 @@ class FunSpec private constructor(builder: Builder) {
 
   class Builder internal constructor(internal val name: String) {
     internal val kdoc = CodeBlock.builder()
+    internal val returnKdoc = CodeBlock.builder()
     internal var receiverType: TypeName? = null
     internal var returnType: TypeName? = null
     internal var delegateConstructor: String? = null
@@ -263,14 +266,17 @@ class FunSpec private constructor(builder: Builder) {
 
     fun receiver(receiverType: KClass<*>) = receiver(receiverType.asTypeName())
 
-    fun returns(returnType: TypeName) = apply {
+    @JvmOverloads fun returns(returnType: TypeName, kdoc: CodeBlock? = null) = apply {
       check(!name.isConstructor && !name.isAccessor) { "$name cannot have a return type" }
       this.returnType = returnType
+      kdoc?.let {
+        this.returnKdoc.add(it)
+      }
     }
 
-    fun returns(returnType: Type) = returns(returnType.asTypeName())
+    @JvmOverloads fun returns(returnType: Type, kdoc: CodeBlock? = null) = returns(returnType.asTypeName(), kdoc)
 
-    fun returns(returnType: KClass<*>) = returns(returnType.asTypeName())
+    @JvmOverloads fun returns(returnType: KClass<*>, kdoc: CodeBlock? = null) = returns(returnType.asTypeName(), kdoc)
 
     fun addParameters(parameterSpecs: Iterable<ParameterSpec>) = apply {
       for (parameterSpec in parameterSpecs) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -263,14 +266,17 @@ class FunSpec private constructor(builder: Builder) {
 
     fun receiver(receiverType: KClass<*>) = receiver(receiverType.asTypeName())
 
-    fun returns(returnType: TypeName) = apply {
+    @JvmOverloads fun returns(returnType: TypeName, kdoc: CodeBlock? = null) = apply {
```
Egorand: Nice

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -186,6 +186,23 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionParamWithKdocAndReturnKdoc() {
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("string", String::class.asTypeName())
+            .addKdoc("A string parameter.\n")
+            .build())
+        .addParameter(ParameterSpec.builder("nodoc", Boolean::class.asTypeName()).build())
+        .returns(String::class.asTypeName(), CodeBlock.of("@return the foo.\n"))
```
Egorand: I think we should be adding `@return`ourselves, similar to what we do for the params.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -199,6 +216,22 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionWithReturnKDoc() {
+    val unitType = UNIT
```
Egorand: nit: can be inlined

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -199,6 +216,22 @@ class FunSpecTest {
...
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("f", LambdaTypeName.get(returnType = unitType)).build())
+        .returns(String::class, CodeBlock.builder().add("@return the foo.\n").build())
```
Egorand: Also, can we handle the `\n`? Ideally `"the foo"` should be a valid KDoc string, which we'll print out as `@return a foo\n`

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -61,10 +62,10 @@ class FunSpec private constructor(builder: Builder) {
     codeWriter: CodeWriter,
     enclosingName: String?,
     implicitModifiers: Set<KModifier>,
-    includeParametersKdoc: Boolean
+    includeExtrasKdoc: Boolean = false
```
Egorand: Not sold on "extras", any other ideas regarding the naming?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
shaishavgandhi05: Oops that's right!

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
shaishavgandhi05: Sure I don't mind doing that. However, when we print out `@param` comments, we don't add the new line by default and that's why I didn't add the line either. I like the idea of not having to manually append the `\n`, but for the sake of consistency we'd want to add it to the `@param` comments also, which might be a breaking change. What do you think?

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
Egorand: These are called "tags" in Javadoc, so I suggest `includeKdocTags`. The function below should be renamed from `kdocWithParameters()` to `kdocWithTags()`.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -263,14 +266,17 @@ class FunSpec private constructor(builder: Builder) {
...
+      kdoc?.let {
+        this.returnKdoc.add("@return %L", it)
+      }
     }
 
-    fun returns(returnType: Type) = returns(returnType.asTypeName())
+    @JvmOverloads fun returns(returnType: Type, kdoc: CodeBlock? = null) = returns(returnType.asTypeName(), kdoc)
 
-    fun returns(returnType: KClass<*>) = returns(returnType.asTypeName())
+    @JvmOverloads fun returns(returnType: KClass<*>, kdoc: CodeBlock? = null) = returns(returnType.asTypeName(), kdoc)
```
Egorand: How about providing overloads that accept `kdoc` as a `String`? This way you won't need to create a `CodeBlock` for simple cases.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -186,6 +186,23 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionParamWithKdocAndReturnKdoc() {
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("string", String::class.asTypeName())
```
Egorand: nit: `asTypeName()` is redundant, here and in the following lines, there are overloads that accept `KClass`.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -186,6 +186,23 @@ class FunSpecTest {
...
+        .addParameter(ParameterSpec.builder("nodoc", Boolean::class.asTypeName()).build())
+        .returns(String::class.asTypeName(), CodeBlock.of("the foo.\n"))
+        .addCode(CodeBlock.of("return \"foo\""))
+        .build()
+    assertThat(funSpec.toString()).isEqualTo("""
+      |/**
+      | * @param string A string parameter.
+      | * @return the foo.
+      | */
+      |fun foo(string: kotlin.String, nodoc: kotlin.Boolean): kotlin.String = "foo"""".trimMargin())
```
Egorand: Hmm, is this code compilable?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -186,6 +186,23 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionParamWithKdocAndReturnKdoc() {
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("string", String::class.asTypeName())
+            .addKdoc("A string parameter.\n")
+            .build())
+        .addParameter(ParameterSpec.builder("nodoc", Boolean::class.asTypeName()).build())
+        .returns(String::class.asTypeName(), CodeBlock.of("the foo.\n"))
```
Egorand: Yeah, woulda been a lot nicer to write `returns(String::class, kdoc = "the foo")`

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -263,14 +266,17 @@ class FunSpec private constructor(builder: Builder) {
 
     fun receiver(receiverType: KClass<*>) = receiver(receiverType.asTypeName())
 
-    fun returns(returnType: TypeName) = apply {
+    @JvmOverloads fun returns(returnType: TypeName, kdoc: CodeBlock? = null) = apply {
       check(!name.isConstructor && !name.isAccessor) { "$name cannot have a return type" }
       this.returnType = returnType
+      kdoc?.let {
+        this.returnKdoc.add("@return %L", it)
```
Egorand: I think we should also be emitting the `\n` at the end of the KDoc. How will the output look if the user doesn't supply the `\n`?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
Egorand: Sorry, missed this comment before my recent review... I'm not sure what would the result be if `\n` is not passed in as a part of the string, but I guess it's not something somebody would ever want, hence library users will pass in a string without the new line, notice that it looks ugly and will hack the library by appending the `\n` - this sucks. Let's add `\n` ourselves, and I'll open an issue to fix it for params.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
shaishavgandhi05: Yeah I think it makes sense. It looks pretty weird if the user doesn't add the `\n`. I'll make the changes soon!

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
shaishavgandhi05: Done!

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
shaishavgandhi05: Yup, compiles fine.

PR #496, opened 2018-10-15
Shortcut functions

friedrich-goetze:
For my own Project, I created these shortcuts for many of *kotlinpoet*s builders and I like to share them.

pros for usercode:
- will be less cluttered
- less lines
- less indentations, therefore less space taken by indentation
- correct builders are selected automatically (i.e. `FunSpec.constructorBuilder(...)` for `TypeSpec.primaryConstructor(...)`, `FunSpec.getterBuilder(...)` for `PropertySpec.getter(...)` etc...

I also added a Test which is also a small demonstration.

Shortcuts I didn't use are missing, i.e. for `TypeAliasSpec`.

Here's a preview snipped taken from the testclass:

```kotlin
fileSpec("com.squareup.kotlinpoet.test", "FooBar") {
  propertySpec("foo", String::class, INTERNAL)

  classSpec("FooClass") {
    primaryConstructorSpec {
      addCode("println(\"Hello World\")")
    }

    varPropertySpec("foo", String::class, INTERNAL) {
      initializer("\"the foo\"")
    }

    propertySpec("foobar", String::class) {
      getterSpec { addCode("return \"\${foo}bar\"") }
    }

    funSpec("dooTheFooBar") {
      returns(Int::class.asTypeName().asNullable())
      addParameter("howLong", Period::class, INTERNAL)
      addCode("println(\"\$foobar\")")
    }
  }

  interfaceSpec("FooInterface") {
    funSpec("calcMagicNumber") {
      returns(Long::class)
      addCode("return 42L")
    }
  }

  objectSpec("FooObject") {
  }
}
```

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/Shortcuts.kt
+++ b/src/main/java/com/squareup/kotlinpoet/Shortcuts.kt
@@ -1,0 +1,343 @@
+package com.squareup.kotlinpoet
+
+import java.lang.reflect.Type
+import javax.lang.model.element.ExecutableElement
+import javax.lang.model.type.DeclaredType
+import javax.lang.model.util.Types
+import kotlin.reflect.KClass
+
+/*
+ * Shortcuts for FileSpec
+ */
+
+fun fileSpec(
+  packageName: String,
+  fileName: String,
+  block: FileSpec.Builder.() -> Unit
+) = FileSpec.builder(packageName, fileName).apply(block).build()
+
+/*
+ * Shortcuts for adding TypeSpec to FileSpec
+ */
+
+fun FileSpec.Builder.classSpec(
+  name: String,
+  block: TypeSpec.Builder.() -> Unit
+): FileSpec.Builder =
+  addType(TypeSpec.classBuilder(name).apply(block).build())
+
+fun FileSpec.Builder.classSpec(
+  className: ClassName,
+  block: TypeSpec.Builder.() -> Unit
+): FileSpec.Builder =
+  addType(TypeSpec.classBuilder(className).apply(block).build())
+
+fun FileSpec.Builder.expectClassSpec(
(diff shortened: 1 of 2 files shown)


JakeWharton:
Without the use of `@DslContext` on the builders you can create weird representations in the DSL that don't map to what you expect. Otherwise these seem like mostly okay as far as DSLs go. Perhaps you want to illicit discussion on #62 first.

friedrich-goetze:
Using  `@DslMarker` should really help, but this annotation has to be applied to another annotation (you have to define yourself), which then has to be applied all Foo.Builder types. Therefore, after creating an annotation like `@KotlinPoetDsl`, one of the following has to be done:
- Annotate every builder-class with it
- Make the builder *open*, so an inherited class can be annotated
- Add annotated wrapperclasses for each builder

tieskedh:
there is a fourth option, which seems the right one:

```kotlin
class A{
    val aval : String = ""
}
class B {
    val bval :String = ""
}
fun a( block: (@KotlinPoetDSL A).()->Unit): Unit = TODO()
fun A.b(block: (@KotlinPoetDSL B).()->Unit): Unit = TODO()
fun main(args: Array<String>) {
    a { 
        aval
        b { 
            aval
            bval
        }
    }
}
```

JakeWharton:
We control the types, though. Why wouldn't we just annotate them directly?

tieskedh:
don't know ;-)
I placed this in the discussion as well.
The only reason I can come up to don't do it is that in this way you can add it as a plugin.
I added it because I misread the `.` in your answer for a `,` :-).

swankjesse:
I don't think we want this in the core library. It's neat and nice but it means we offer two competing APIs. This one is nice if you have full context, but the Builder API is most flexible.

Issue #499, opened 2018-10-19
FileSpec.Builder with blank package name fails


michaelfoidl:
I am currently trying to generate a simple Kotlin class without a specified package (this has internal reasons). However, when I pass an empty string as an argument for the `FileSpec.Builder`, the file never is created. Any other package name works just fine.

`FileSpec.builder("", configurationName)`

Am I missing something or is this a bug?

As far as I understand, this has something to do with the `writeTo` function: If the package name is left blank, the (potentially missing) directories are never created (see linked code below).

[https://github.com/square/kotlinpoet/blob/6921368f2c810a1dc8b5b835542ef52aa2c49f35/src/main/java/com/squareup/kotlinpoet/FileSpec.kt#L67](url)

In my case, the target location does not exist at build time which results in

`An exception occurred: java.nio.file.NoSuchFileException:
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:79)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:230)
	at java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:434)
	at java.nio.file.Files.newOutputStream(Files.java:216)
	at com.squareup.kotlinpoet.FileSpec.writeTo(FileSpec.kt:80)
	at com.squareup.kotlinpoet.FileSpec.writeTo(FileSpec.kt:85)`

PR #500, opened 2018-10-19
FileSpec.Builder blank package and subfolder fix

michaelfoidl:
When creating files in subfolders, these are not created if no package is specified. This problem also described in #499 was fixed by changing the code so that missing directories are always created.
I also added a test for this kind of situation.


michaelfoidl:
done

PR #505, opened 2018-10-22
Return KDoc follow up

shaishavgandhi05:
There were some subtle bugs with the initial implementation of adding KDoc to `returns()`. 
1. This fixes the bug where `toBuilder()` wouldn't have the Kdoc associated with the return type.
2. Have moved the `returnKdoc` to only have the description and not the `@return` keyword as well. We only add the `@return` on emission. 
3. To accurately model `returns`, we now just replace the Kdoc given in the `returns()` method instead of appending to the `returnKdoc` CodeBlock. This makes more sense since if you do `toBuilder()` and then change the return type, the newer Kdoc should replace the older one and not append to it.

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -221,6 +221,22 @@ class FunSpecTest {
       |fun foo(string: kotlin.String, nodoc: kotlin.Boolean): kotlin.String = "foo"""".trimMargin())
   }
 
+  @Test fun functionWithModifiedReturnKdoc() {
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("nodoc", Boolean::class).build())
+        .returns(String::class, "the foo.")
+        .addCode(CodeBlock.of("return \"foo\""))
+        .build()
+        .toBuilder()
+        .returns(String::class, "the modified foo.")
+        .build()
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -221,6 +221,22 @@ class FunSpecTest {
       |fun foo(string: kotlin.String, nodoc: kotlin.Boolean): kotlin.String = "foo"""".trimMargin())
   }
 
+  @Test fun functionWithModifiedReturnKdoc() {
+    val funSpec = FunSpec.builder("foo")
+        .addParameter(ParameterSpec.builder("nodoc", Boolean::class).build())
```
Egorand: supernit: use the overload that accepts `name` and `type`, no need to create `ParameterSpec` manually here

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -221,6 +221,22 @@ class FunSpecTest {
...
+        .returns(String::class, "the foo.")
+        .addCode(CodeBlock.of("return \"foo\""))
```
Egorand: supernit: `.addCode("return %S", "foo")` is the pattern we generally adhere to

shaishavgandhi05: TIL. The one I did always felt awkward. Have fixed it in a test above this one too which had it.

PR #511, opened 2018-10-23
Add support for receiver kdoc

### LINKED ISSUE #509: @receiver Kdoc overloads
In the spirit of covering the major [tags](https://kotlinlang.org/docs/reference/kotlin-doc.html#block-tags) in KDoc, this would be a good addition. Extension functions are pretty widely used and we should provide overloads similar to `returns` that adds documentation to it. 

Additionally, if I understand correctly, the project aims to emulate the Android Kotlin style guide. That guide mentions the following
```
Block tags
Any of the standard “block tags” that are used appear in the order @constructor, @receiver, 
@param, @property, @return, @throws, @see, and these never appear with an empty description. 
When a block tag doesn’t fit on a single line, continuation lines are indented 8 spaces from 
the position of the @.
```

In that spirit, we should add the `@receiver` tag before `@param` or `@return`. 

Happy to send a PR if this is something that you think is a good addition to KotlinPoet.

shaishavgandhi05:
Adds support for adding a kdoc to the receiver. The order of the Kdoc tags are according to the [Android Kotlin Style Guide](https://developer.android.com/kotlin/style-guide).

Resolves #509

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -34,6 +34,7 @@ class FunSpec private constructor(builder: Builder) {
   val name = builder.name
   val kdoc = builder.kdoc.build()
   val returnKdoc = builder.returnKdoc
+  val receiverKdoc = builder.receiverKdoc
   val annotations = builder.annotations.toImmutableList()
   val modifiers = builder.modifiers.toImmutableSet()
   val typeVariables = builder.typeVariables.toImmutableList()
@@ -141,6 +142,9 @@ class FunSpec private constructor(builder: Builder) {
 
   private fun kdocWithTags(): CodeBlock {
     return with(kdoc.toBuilder()) {
+      if (receiverKdoc.isNotEmpty()) {
+        add("@receiver %L\n", receiverKdoc)
+      }
       for (parameterSpec in parameters) {
         if (parameterSpec.kdoc.isNotEmpty()) {
           add("@param %L %L\n", parameterSpec.name, parameterSpec.kdoc)
@@ -174,6 +178,7 @@ class FunSpec private constructor(builder: Builder) {
     val builder = Builder(name)
     builder.kdoc.add(kdoc)
     builder.returnKdoc = returnKdoc
+    builder.receiverKdoc = receiverKdoc
     builder.annotations += annotations
     builder.modifiers += modifiers
     builder.typeVariables += typeVariables
@@ -189,6 +194,7 @@ class FunSpec private constructor(builder: Builder) {
   class Builder internal constructor(internal val name: String) {
     internal val kdoc = CodeBlock.builder()
     internal var returnKdoc = CodeBlock.EMPTY
+    internal var receiverKdoc = CodeBlock.EMPTY
     internal var receiverType: TypeName? = null
     internal var returnType: TypeName? = null
     internal var delegateConstructor: String? = null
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -260,14 +266,21 @@ class FunSpec private constructor(builder: Builder) {
       typeVariables += typeVariable
     }
 
-    fun receiver(receiverType: TypeName) = apply {
+    @JvmOverloads fun receiver(receiverType: TypeName, kdoc: CodeBlock? = null) = apply {
```
Egorand: Since the default field value is `CodeBlock.EMPTY`, would it make sense to change the `kdoc` type to non-nullable and use `CodeBlock.EMPTY` as default? That would prevent people from explicitly passing `null` and we won't need to null-check in the method body.

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -476,6 +476,43 @@ class FunSpecTest {
     assertThat(funSpec.toBuilder().build()).isEqualTo(funSpec);
   }
 
+  @Test fun receiverWithKdoc() {
+    val funSpec = FunSpec.builder("toBar")
+        .receiver(String::class, "the string to transform.")
```
Egorand: I think we want to promote using named args for this parameter, can you please change it to `kdoc = "the string to transform."`)

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -476,6 +476,43 @@ class FunSpecTest {
...
+    assertThat(funSpec.toString()).isEqualTo("""
+      |/**
+      | * @receiver the string to transform.
+      | */
+      |fun kotlin.String.toBar(): kotlin.String = "bar"""".trimMargin())
+  }
+
+  @Test fun withAllKdocTags() {
+    val funSpec = FunSpec.builder("charAt")
+        .receiver(String::class, "the string you want the char from.")
```
Egorand: Same here, and for `returns`

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -476,6 +476,43 @@ class FunSpecTest {
...
+        .receiver(String::class, "the string to transform.")
+        .returns(String::class)
+        .addCode(CodeBlock.of("return %S", "bar"))
+        .build()
+
+    assertThat(funSpec.toString()).isEqualTo("""
+      |/**
+      | * @receiver the string to transform.
+      | */
+      |fun kotlin.String.toBar(): kotlin.String = "bar"""".trimMargin())
```
Egorand: nit: Having `""".trimMargin()` is really confusing, can you move it to the new line?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -476,6 +476,43 @@ class FunSpecTest {
...
+    assertThat(funSpec.toString()).isEqualTo("""
+      |/**
+      | * Returns the character at the given [position].
+      | *
+      | * @receiver the string you want the char from.
+      | * @param position the index of the character that is returned.
+      | * @return The char at the given [position].
+      | */
+      |fun kotlin.String.charAt(position: kotlin.Int): kotlin.Char = -1
+    """.trimMargin())
```
Egorand: nit: indent

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
shaishavgandhi05: Good point! The same thing makes sense for `returns` also. I can address that in a follow up.

PR #512, opened 2018-10-23
Avoid emitting "Unit" as return type

shaishavgandhi05:
In accordance with the Kotlin [coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#avoiding-redundant-constructs), we can safely omit emitting the "Unit" class when it's used as a return type. 

I'm not aware of any other wildcards like "Unit", hence just the single exclusion. If you know of any, I can make the check more general.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -125,7 +125,8 @@ class FunSpec private constructor(builder: Builder) {
       param.emit(codeWriter, includeType = name != SETTER)
     }
 
-    if (returnType != null) {
+    // Omit emitting "Unit" for redundancy.
+    if (returnType != null && returnType != Unit::class.asTypeName()) {
       codeWriter.emitCode(": %T", returnType)
     }
 
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -125,8 +125,7 @@ class FunSpec private constructor(builder: Builder) {
       param.emit(codeWriter, includeType = name != SETTER)
     }
 
-    // Omit emitting "Unit" for redundancy.
-    if (returnType != null && returnType != Unit::class.asTypeName()) {
+    if (emitReturnType(returnType)) {
       codeWriter.emitCode(": %T", returnType)
     }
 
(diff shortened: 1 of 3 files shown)


JakeWharton:
I'm hesitant about the behavior of this combined with expression bodies. If the method to which you are delegating changes its return type we would propagate that type instead of causing a compiler error.

shaishavgandhi05:
Agreed that `returns` is optional. Was mainly thinking of a case where the return type is inferred from an annotation in Java/Kotlin and not explicitly added as `Unit` by the consumer. Since IntelliJ would warn about the redundancy,I thought of removing it. 

@JakeWharton I'm not sure I completely understand. Can you give an example where this would break?  

It's a nice-to-have and I'm not married to it too much. If you feel it's a good addition and we can find a way around the expression bodies thing, then that's great. Otherwise I can close this.

JakeWharton:
If you want a method `fun foo(): Unit` that delegates to `fun bar(): Unit` you might write generation for
```kotlin
fun foo(): Unit {
  return bar()
}
```
In order to get KotlinPoet to collapse this to
```kotlin
fun foo(): Unit = bar()
```
This is fine.

Currently, if `bar()` changes its return type from `Unit` to `String` you would get a compilation error in your generated code.

If we omit `: Unit`, though, your code would have compiled but the return type of the function would no longer be `Unit` it would be `String` and would likely cause a compilation error for the callers of `foo()`. Or worse, it might go unnoticed resulting in a binary-incompatible change to your code.

Granted, this is a fairly contrived case as I don't expect many (if anyone) to be doing this. And it can be worked around by not omitting the return type for expression body functions.

shaishavgandhi05:
Thanks for clarifying! Yeah it's true that that would break functionality in various ways. 

As you mentioned however, it's a rather non-probable edge case. My reasoning is that KotlinPoet code would be extremely close to what you would write yourself. In that case, you would probably be removing the `Unit` especially since IntelliJ tells you to and in that case you would be exposed to the same problem. 

Still, I think it's better to err on the side of caution. I can update it to exclude removing the `Unit` for expression bodies and add a test case for it!

JakeWharton:
In library mode, if they ever release it, visibility modifiers and return types must always be specified. I would prefer we don't do this for expression bodies to be safe. Normal function bodies are fine.

Egorand:
That's a good compromise, let's remove `Unit` only if we emit function body and keep it for single-expression functions.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -153,6 +153,26 @@ class FunSpec private constructor(builder: Builder) {
...
+   * delegating function.
+   */
+  private fun emitReturnType(returnType: TypeName?): Boolean {
+    if (returnType != null) {
+      return returnType != Unit::class.asTypeName() || isExpressionBody()
+    }
+    return false
+  }
+
+  private fun isExpressionBody(): Boolean {
```
JakeWharton: Shouldn't this be used elsewhere to determine behavior as well?

shaishavgandhi05: The only other place it's required is where you actually want to emit the expressionBody returned from `withoutPrefix()` so no point using it there. https://github.com/square/kotlinpoet/blob/16c8e1e5c6bd2d3cbe513fa60ee114a7c4be4c81/src/main/java/com/squareup/kotlinpoet/FunSpec.kt#L94

We are doing a double check of sorts which feels weird but I couldn't find an easy way to consolidate both.

PR #514, opened 2018-10-26
Add support for multi-line file comments

### LINKED ISSUE #358: Support multiline comments in FileSpec
Add `FileSpec.addKDoc()`

shaishavgandhi05:
This changes behavior to emit in format:
```kotlin
/**
 * MULTI LINE
 * COMMENT
 */
```

instead of
```kotlin
// MULTI LINE
// COMMENT
```
when the comments are more than one line. 

Resolves #358

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -87,7 +87,7 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
 
   private fun emit(codeWriter: CodeWriter) {
     if (comment.isNotEmpty()) {
-      if (isMultiLine(comment)) {
+      if (comment.isMultiLine()) {
         codeWriter.emitKdoc(comment)
       } else {
         codeWriter.emitComment(comment)
@@ -135,8 +135,8 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     codeWriter.popPackage()
   }
 
-  private fun isMultiLine(comment: CodeBlock): Boolean {
-    return comment.toString().contains("\n")
+  private fun CodeBlock.isMultiLine(): Boolean {
+    return this.toString().contains("\n")
   }
 
   override fun equals(other: Any?): Boolean {
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -87,7 +87,11 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
 
   private fun emit(codeWriter: CodeWriter) {
     if (comment.isNotEmpty()) {
-      codeWriter.emitComment(comment)
+      if (isMultiLine(comment)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -131,6 +135,10 @@ class FileSpec private constructor(builder: FileSpec.Builder) {
     codeWriter.popPackage()
   }
 
+  private fun isMultiLine(comment: CodeBlock): Boolean {
```
shaishavgandhi05: Wasn't sure whether to pollute `CodeBlock` with this utility method. Does this seem general enough that it could be used somewhere else? We could add it to `CodeBlock` in that case.

src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -571,17 +571,60 @@ class FileSpecTest {
...
+            .addStatement("distributed under the License is distributed on an \"AS IS\" BASIS,")
+            .addStatement("WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.")
+            .addStatement("See the License for the specific language governing permissions and")
+            .addStatement("limitations under the License.")
+            .build())
+        .build()
+
+    assertThat(source.toString()).isEqualTo("""
+      |/**
+      | * Copyright (C) 2015 Square, Inc.
```
Egorand: Looks awesome!

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
Egorand: It's fine having it here for now, we'll have an option of moving it to `CodeBlock` when we have more than one usage.

nit: It would read better if it was an extension function.

shaishavgandhi05: Cool! 
Re: extension function. It is my understanding that since we already own `CodeBlock`, adding an extension function would be equivalent to adding the function in `CodeBlock` class itself, since for all purposes, it becomes part of `CodeBlock`'s API. Is there any advantage to adding it as extension function?

Egorand: It'll be an extension function private to `FileSpec`, similar to creating a `private static isMultiline(CodeBlock comment)`. Readability in itself is a big advantage IMO.

shaishavgandhi05: Ahh interesting! Yep, making it private to `FileSpec` makes a lot of sense. Thanks!

swankjesse:
Just reading this now. I think this is a mistake. Comments may be prefixed with `/**`, `/*`, or `//`.

The first is only for KDoc comments. I don't believe these are KDoc comments, are they? If they are then they're intended to be included in the Dokka HTML report.

I think only the other two forms are correct. All else being equal I prefer // because it requires fewer lines of code.

ZacSweers:
Compose uses this but just needs the middle format. Would the right fix just be to use `/*` and never `/**`?

PR #517, opened 2018-10-30
Add checks for inline classes

### LINKED ISSUE #504: Add support for inline classes
https://kotlinlang.org/docs/reference/inline-classes.html

shaishavgandhi05:
Adds support for doing checks around `inline` classes. Seems like a good time now that 1.3 is official! 

I've used this [KEEP](https://github.com/Kotlin/KEEP/blob/master/proposals/inline-classes.md) as the main reference. It covers most checks like
* Single parameter constructor
* Public constructor
* No super classes
* Can implement interfaces
* Only single backing property
* No `init` blocks
* Read only (`val`) properies.
and also has tests for each of these cases. 

Have also added 2 additional tests to cover valid inline classes as well as inline enum classes. 

Resolves #504

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -476,7 +476,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
         if (isInlineClass) {
           check(primaryConstructor.parameters.size == 1) {
-            "Inline class can only have 1 parameter in constructor"
+            "Inline class must have 1 parameter in constructor"
           }
         }
       }
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -24,6 +24,8 @@ import com.squareup.kotlinpoet.KModifier.EXTERNAL
 import com.squareup.kotlinpoet.KModifier.PUBLIC
 import com.squareup.kotlinpoet.KModifier.SEALED
 import com.squareup.kotlinpoet.KModifier.INLINE
+import com.squareup.kotlinpoet.KModifier.PRIVATE
+import com.squareup.kotlinpoet.KModifier.INTERNAL
 import com.squareup.kotlinpoet.TypeSpec.Kind.Interface
 import com.squareup.kotlinpoet.TypeSpec.Kind.Object
 import java.lang.reflect.Type
@@ -669,16 +671,30 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
 
       if (isInlineClass) {
         primaryConstructor?.let {
-          // Inline classes only have 1 parameter
+          // Inline classes only have 1 constructor parameter
           check(it.parameters.size == 1) {
-            "Inline class can only have 1 parameter in constructor"
+            "Inline class must have 1 parameter in constructor"
+          }
+          // Inline class must have public primary constructor
+          check(PRIVATE !in it.modifiers && INTERNAL !in it.modifiers) {
+            "Inline class must have a public primary constructor"
           }
         }
+        // Inline classes can only have 1 backing property
         check(propertySpecs.size == 1) {
-          "Inline class can only have 1 property"
+          "Inline class must have a single read-only (val) property."
         }
+        // Inline classes property must be immutable.
         check(!propertySpecs.first().mutable) {
           "Inline class must have a single read-only (val) property."
+        }
+        // Inline class cannot have init blocks
+        check(initializerBlock.isEmpty()) {
+          "Inline class can't have initializer blocks"
+        }
+        // Inline class cannot have super class
+        check(superclass == Any::class.asTypeName()) {
+          "Inline classes cannot have super classes"
         }
       }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -575,12 +600,7 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         = addProperty(name, type.asTypeName(), *modifiers)
 
     fun addInitializerBlock(block: CodeBlock) = apply {
-      check(isSimpleClass || isEnum || kind is Object) {
-        "$kind can't have initializer blocks"
-      }
-      check(EXPECT !in kind.modifiers) {
-        "expect $kind can't have initializer blocks"
-      }
+      checkCanHaveInitializerBlocks()
```
Egorand: Nice

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -649,6 +669,35 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
         }
       }
 
+      if (isInlineClass) {
+        primaryConstructor?.let {
+          // Inline classes only have 1 constructor parameter
```
Egorand: I find these comments redundant, the code is self-explanatory

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -649,6 +669,35 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          check(it.parameters.size == 1) {
+            "Inline class must have 1 parameter in constructor"
```
Egorand: nit: Either "An inline class" or "Inline classes"

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -649,6 +669,35 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          }
+          // Inline class must have public primary constructor
+          check(PRIVATE !in it.modifiers && INTERNAL !in it.modifiers) {
+            "Inline class must have a public primary constructor"
+          }
+        }
+        // Inline classes can only have 1 backing property
+        check(propertySpecs.size == 1) {
```
Egorand: I believe inline classes can declare additional properties, as long as they don't have backing fields

src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -649,6 +669,35 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
...
+          "Inline class must have a single read-only (val) property."
+        }
+        // Inline classes property must be immutable.
+        check(!propertySpecs.first().mutable) {
```
Egorand: We should allow declaring multiple properties, and then there will be no guarantee that `first()` returns the primary constructor property, so we'll have to query the property by the corresponding constructor param name

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1755,6 +1756,151 @@ class TypeSpecTest {
         |""".trimMargin())
   }
 
+  @Test fun validInlineClass() {
+    val guacamole = TypeSpec.classBuilder("Guacamole")
+        .primaryConstructor(FunSpec.constructorBuilder()
+            .addParameter("avacado", String::class)
```
Egorand: supernit: it's "avOcado" :)

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1755,6 +1756,151 @@ class TypeSpecTest {
...
+
+  @Test fun inlineClassWithMutableProperties() {
+    assertThrows<IllegalStateException> {
+      TypeSpec.classBuilder("Guacamole")
+          .primaryConstructor(FunSpec.constructorBuilder()
+              .addParameter("avacado", String::class)
+              .build())
+          .addProperty(PropertySpec.builder("avacado", String::class)
+              .initializer("avacado")
+              .mutable(true)
```
Egorand: nit: `true` can be omitted

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1755,6 +1756,151 @@ class TypeSpecTest {
...
+    }.hasMessageThat().isEqualTo("Inline class must have a public primary constructor")
+  }
+
+  @Test fun inlineEnumClass() {
+    val guacamole = TypeSpec.enumBuilder("Foo")
+        .primaryConstructor(FunSpec.constructorBuilder()
+            .addParameter("x", Int::class)
+            .build())
+        .addEnumConstant("A", TypeSpec.anonymousClassBuilder()
+            .addSuperclassConstructorParameter("%S", "1")
```
Egorand: supernit: Constructor param's type is `Int`, why are we passing a `String`?

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
shaishavgandhi05: I'm horrified. I misspelled my beloved avocado. :cry:

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
shaishavgandhi05: Woops. My bad

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
Egorand: Lol

Egorand:
Thanks!

PR #518, opened 2018-11-01
Add support for empty setters

### LINKED ISSUE #494: lateinit var with private setter doesnt work
Here's my approach:
```kotlin
PropertySpec.varBuilder("foo", String::class, LATEINIT)
    .setter(FunSpec.setterBuilder()
        .addModifiers(PRIVATE)
        .build()
    ).build()
```
I would expect the following kotlin code to be generated:

```kotlin
lateinit var foo: String
    private set
```
But I get:
```java
java.lang.IllegalStateException: set() can have only one parameter
        at com.squareup.kotlinpoet.FunSpec$Builder.build(FunSpec.kt:339)
        ...
```

cwwwu:
Fixes #494

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -44,13 +44,14 @@ class FunSpec private constructor(builder: Builder) {
   val delegateConstructor = builder.delegateConstructor
   val delegateConstructorArguments = builder.delegateConstructorArguments.toImmutableList()
   val body = builder.body.build()
+  private val isEmptySetter = name == SETTER && parameters.isEmpty()
 
   init {
     require(body.isEmpty() || ABSTRACT !in builder.modifiers) {
       "abstract function ${builder.name} cannot have code"
     }
-    require(name != SETTER || parameters.size == 1) {
-      "$name must have exactly one parameter"
+    require(name != SETTER || parameters.size <= 1) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -44,6 +44,7 @@ class FunSpec private constructor(builder: Builder) {
   val delegateConstructor = builder.delegateConstructor
   val delegateConstructorArguments = builder.delegateConstructorArguments.toImmutableList()
   val body = builder.body.build()
+  val isEmptySetter = name == SETTER && parameters.isEmpty()
```
JakeWharton: private

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -44,12 +44,13 @@ class FunSpec private constructor(builder: Builder) {
...
   val body = builder.body.build()
+  private val isEmptySetter = name == SETTER && parameters.isEmpty()
 
   init {
     require(body.isEmpty() || ABSTRACT !in builder.modifiers) {
       "abstract function ${builder.name} cannot have code"
     }
-    require(name != SETTER || parameters.size == 1) {
+    require(name != SETTER || parameters.size <= 1) {
       "$name must have exactly one parameter"
```
shaishavgandhi05: Can we change the message to reflect the newer condition?

PR #522, opened 2018-11-05
Escape property names if Kotlin keywords

### LINKED ISSUE #476: Escape parameter name in PropertySpec which is also a Kotlin keyword

shaishavgandhi05:
I didn't have push access to the previous branch and so I couldn't just rebase the conflicts. I've basically just copied over the changes and tests. 

Closes #476


Egorand:
Neat, thanks!

PR #525, opened 2018-11-06
Fold property with escaped name

### LINKED ISSUE #483: Primary constructor parameter with escaped name doesn't get matched with the corresponding initializer
I generate a type like so:

```kotlin
val typeBuilder = TypeSpec.classBuilder(name)
    .addProperties(properties.map {
        PropertySpec
            .varBuilder(it.key, parameterType(it.key, it.value))
            .initializer(it.key)
            .build()
    })
        .addParameters(
            properties.map {
                ParameterSpec
                    .builder(it.key, parameterType(it.key, it.value))
                    .build()
            }
        )
        .build())
```

and get the following result

```kotlin
class AlarmInfo(
    `when`: Float? = null,
    var delayInMinutes: Float? = null,
    var periodInMinutes: Float? = null
) {
    var `when`: Float? = when
}
```

If I manually escape the .initializer(it.key) part, it works. Still, not optimal.

Follow up to #477

Egorand:
Fixes #483


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -150,14 +150,18 @@ class CodeBlock private constructor(
...
+    return toUnescapedString() == other.toUnescapedString()
   }
 
-  override fun hashCode() = toString().hashCode()
+  override fun hashCode() = toUnescapedString().hashCode()
 
   override fun toString() = buildString { CodeWriter(this).emitCode(this@CodeBlock) }
 
+  private fun toUnescapedString() = toString().run {
+    if (isNotEmpty() && first() == '`' && last() == '`') substring(1, length - 1) else this
```
shaishavgandhi05: The `first()` and `last()` look good in context of the bug. Just feels like `toUnescapedString()` is a misnomer since it only unescapes examples like "`when`" and not examples where the escaping might be somewhere in the middle of the CodeBlock (of course those examples wouldn't need the `equals()` then)

JakeWharton:
I'm not sure this behavior makes sense. If the class was named `Identifier` it might, but doing it for `CodeBlock` is a bit weird. Can this logic live at the collapsing logic site instead?

Egorand:
Agree. What seemed to be the problem to me is that by comparing `CodeBlock`s using `toString()` we're mixing presentation logic with business logic, which doesn't feel right, but it might not be the real cause here. I'll try to approach this from a different angle.

PR #528, opened 2018-11-20
Replace ASCII escape sequences with unicode symbols

### LINKED ISSUE #486: Replace ASCII escapes with fancy Unicode symbols
I find the `%[` things hard to read in source code. Using Unicode characters instead might be more readable. That would bring us here:

 * `⇥` increases the indentation level.
 * `⇤` decreases the indentation level.
 * `«` begins a statement. For multiline statements, every line after the first line is double-indented.
 * `»` ends a statement.

cwwwu:
Closes #486

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -27,8 +27,8 @@ import kotlin.reflect.KClass
  * Code blocks are not necessarily well-formed Kotlin code, and are not validated. This class
  * assumes kotlinc will check correctness later!
  *
- * Code blocks support placeholders like [java.text.Format]. This class uses a percent sign
- * `%` but has its own set of permitted placeholders:
+ * Code blocks support placeholders like [java.text.Format]. This class primarily uses a percent
+ * sign `%` but has its own set of permitted placeholders:
  *
  *  * `%L` emits a *literal* value with no escaping. Arguments for literals may be strings,
  *    primitives, [type declarations][TypeSpec], [annotations][AnnotationSpec] and even other code
@@ -47,11 +47,11 @@ import kotlin.reflect.KClass
  *  * `%%` emits a percent sign.
  *  * `%W` emits a space or a newline, depending on its position on the line. This prefers to wrap
  *    lines before 100 columns.
- *  * `%>` increases the indentation level.
- *  * `%<` decreases the indentation level.
- *  * `%[` begins a statement. For multiline statements, every line after the first line is
+ *  * `⇥` increases the indentation level.
+ *  * `⇤` decreases the indentation level.
+ *  * `«` begins a statement. For multiline statements, every line after the first line is
  *    double-indented.
- *  * `%]` ends a statement.
+ *  * `»` ends a statement.
  */
 class CodeBlock private constructor(
   internal val formatParts: List<String>,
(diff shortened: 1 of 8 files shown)


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -194,6 +194,12 @@ class CodeBlock private constructor(
       }
 
       while (p < format.length) {
+        if (format[p] in setOf('⇥', '⇤', '«', '»')) {
```
Egorand: How about checking for these symbols in `isNoArgPlaceholder()`?

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -251,8 +257,14 @@ class CodeBlock private constructor(
...
+          formatParts += format[p].toString()
+          p++
+          continue
+        }
+
         if (format[p] != '%') {
-          var nextP = format.indexOf('%', p + 1)
+          var nextP = format.indexOfAny(listOf("%", "«", "»", "⇥", "⇤"), p + 1)
```
Egorand: nit: Use the version that takes a `CharArray`

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
cwwwu: There are a handful of places that use `isNoArgPlaceholder()` assuming that it's prefixed with a `%`. I'll rework this a bit (maybe a new method to check for single character argument placeholders)

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
cwwwu: 👍Fixed that. I also found a bug in the midst of making that change in handling the new single character placeholders in `addNamed` 😄

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -90,7 +90,7 @@ class CodeBlock private constructor(
       }
 
       // If the matching format part has an argument, check that too.
-      if (formatPart.startsWith("%") && !isNoArgPlaceholder(formatPart[1])) {
+      if (formatPart.startsWith("%") && !isMultiCharNoArgPlaceholder(formatPart[1])) {
```
Egorand: Nice, I like the name.

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -107,8 +83,8 @@ class CodeBlockTest {
     val map = LinkedHashMap<String, Any>()
     map.put("text", "tacos")
     val block = CodeBlock.builder()
-        .addNamed("%>\n%text:L for %%3.50", map).build()
-    assertThat(block.toString()).isEqualTo("\n    tacos for %3.50")
+        .addNamed("⇥\n%text:L for\n⇤%%3.50", map).build()
```
Egorand: Any specific reason for this test change? Does the code still work with the old test case?

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Egorand: Ha, nice. Any new test cases that we should add?

src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
cwwwu: Yes, it still works. I changed this test case as a result of encountering that bug I mentioned above. The key difference here is that there is a single no-arg placeholder in the middle of code block.

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
cwwwu: I ended up modifying the test `namedAndNoArgFormat` for this.

PR #533, opened 2018-11-25
Revert "Add support for multi-line file comments (#514)"

swankjesse:
This reverts commit 3c6174f4adbba39fc4b9bad8f79b30d602bdb334.


swankjesse:
I think we might want to discuss when to use `/*` vs. `//` for comments. But I don’t think we want to use KDoc here because there isn’t a declaration that we’re documenting.

shaishavgandhi05:
That's fair. I missed the fact that a file license header starts with a `/*` and not a `/**`, the latter being the KDoc. 

I think the primary use-case for that issue was the ability to add a license header.

Egorand:
Fair point.

PR #534, opened 2018-11-25
Move the modifiers set out of TypeSpec.Kind

swankjesse:
This was a modeling mistake caused by following Java kinds like '@interface'
and 'enum' rather than Kotlin kinds which are only 'object', 'interface', and
'class'.


Egorand:
Agree, this design is better.

PR #535, opened 2018-11-26
Move WildcardTypeName.Companion.STAR to be beside UNIT and other types

swankjesse:
It was weird to have many interesting types defined together but not this one.

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/WildcardTypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/WildcardTypeName.kt
@@ -46,37 +46,32 @@ class WildcardTypeName private constructor(
   override fun withoutAnnotations() = WildcardTypeName(upperBounds, lowerBounds, nullable)
 
   override fun emit(out: CodeWriter): CodeWriter {
-    if (lowerBounds.size == 1) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/squareup/kotlinpoet/WildcardTypeName.kt
```diff
@@ -46,37 +46,32 @@ class WildcardTypeName private constructor(
...
-     * is `CharSequence.class`, this returns `? extends CharSequence`. If `bound` is `Object.class`,
-     * this returns `?`, which is shorthand for `? extends Object`.
+     * Returns a type that represents an unknown type that extends `upperBound`. For example, if
+     * `upperBound` is `CharSequence`, this returns `out CharSequence`. If `upperBound` is `Any?`,
+     * this returns `*`, which is shorthand for `out Any?`.
      */
-    @JvmStatic fun subtypeOf(upperBound: TypeName): WildcardTypeName {
-      return WildcardTypeName(listOf(upperBound), emptyList())
-    }
+    @JvmStatic fun subtypeOf(upperBound: TypeName) =
```
swankjesse: Should this method be named something with the word `out`? Like outOf?

PR #536, opened 2018-11-26
PR proposal: Make TypeName a sealed class

hzsweers:
Better matches how it's viewed in the compiler, and makes it nice and easy to do things like:

```kotlin
when (typeName) {
  is ClassName -> //
  is TypeVariableName -> //
  else -> //
}
```

in a way that the Kotlin compiler can help us with.

Changes:
- Tried to keep things separated out as extension functions where possible to avoid bloat in the TypeName.kt file
- `Dynamic` is now a top-level `object` type
  - This is potentially awkward

Open questions:
- Keep extension functions? Fold them into the class? Interfaces?
  - Extension functions result in potentially discoverability issues, and requires an import for users.
  - Maybe this could be better as an interface with default impls where we'd have an extension function? In some cases anyway, but comes at the cost of exposing interfaces just for code management.
- What about static factories in companion objects today?
  - Could also be extension functions on the companion objects, but even more awkward for java users
  - Could also use interface approach since companion objects can implement interfaces


hzsweers:
Added examples of the interfaces approach. We can actually avoid exposing the interface per se by marking it internal I think. I need to double check, but my understanding is this makes the methods public while making the interface only implement-able by kotlinpoet.

53cbdf8 demoes it in the normal extension functions. It requires making `names` public and not `internal` though :/. Not sure if this is avoidable
8087a43 demoes it for factories in companion object. Does still require overriding to mark it as `@JvmStatic` though

hzsweers:
Also - the region blocks are so I can do this in the IDE:

image

swankjesse:
Good idea. I think if we’re doing this, we should do it the simplest way possible.

In particular:
 - No extra interfaces.
 - Use regular functions, not extension functions. Extension functions don’t have dynamic dispatch, and dynamic dispatch is good! 
 - No region blocks. This is of limited use when the code structure is already there.

hzsweers:
Cool. I'll rehash this once https://github.com/square/kotlinpoet/pull/535 and related ones are in

PR #545, opened 2018-11-28
Replace TypeName modifier methods with copy()

Egorand:


src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -332,7 +332,7 @@ internal class CodeWriter constructor(
       nameResolved = resolved != null
 
       // We don't care about nullability and type annotations here, as it's irrelevant for imports.
-      if (resolved == c.asNonNull().withoutAnnotations()) {
+      if (resolved == c.copy(nullable = false, annotations = emptyList())) {
```
swankjesse: neat

src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -64,9 +64,12 @@ import kotlin.reflect.KVariance
  * Create new reference types like `com.example.HelloWorld` with [ClassName.bestGuess]. To build composite
  * types like `Set<Long>`, use the factory methods on [ParameterizedTypeName], [TypeVariableName],
  * and [WildcardTypeName].
+ *
+ * @param S the self-type, allowing [copy] to return the subclass type
  */
-sealed class TypeName constructor(
-  val nullable: Boolean, annotations: List<AnnotationSpec>
+sealed class TypeName<out S : TypeName<S>> constructor(
```
swankjesse: Ick. I don’t think the cost is worth the benefit. Would rather we just override all the corresponding methods to narrow the return type

Egorand: Agree, it's super noisy.. Will work on v2

src/test/java/com/squareup/kotlinpoet/AnnotatedTypeNameTest.kt
```diff
@@ -33,52 +33,52 @@ class AnnotatedTypeNameTest {
...
   }
 
   @Test fun annotatedTwice() {
     val expected = "@$NN @java.lang.Override kotlin.String"
     val type = String::class.asTypeName()
-    val actual = type.annotated(NEVER_NULL)
-        .annotated(AnnotationSpec.builder(Override::class).build())
+    val actual = type
+        .copy(annotations = type.annotations + NEVER_NULL +
+            AnnotationSpec.builder(Override::class).build())
```
swankjesse: for later, should we have something like `fun Class<out Annotation>.asAnnotationSpec()` ?

Egorand: Yeah, we should

PR #546, opened 2018-11-28
Rename members of WildcardTypeName to match with the producer/consumer generics model

Egorand:


swankjesse:
inTypes / outTypes makes this really easy to get right. Nice PR.

PR #547, opened 2018-11-28
Rename TypeName.nullable into TypeName.isNullable

Egorand:


src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -412,7 +412,7 @@ class LambdaTypeName private constructor(
   }
 
   fun copy(
-    nullable: Boolean = this.nullable,
+    nullable: Boolean = this.isNullable,
```
shaishavgandhi05: Should the `copy` parameter also be renamed to `isNullable` for consistency?

Egorand: I think `isNullable` reads better when it's an accessor ("is this TypeName nullable?"), and `nullable` reads better when it's a parameter ("I'm creating a nullable TypeName").

shaishavgandhi05: Hmm but then the constructor parameter is `isNullable` 🙂

Egorand: `TypeName` is abstract and its subtypes don't have public constructors

shaishavgandhi05: Oops missed that!

PR #560, opened 2018-12-12
Import known types in KDoc

### LINKED ISSUE #558: Import known types in KDoc
Types referenced in KDoc aren't imported:
```kotlin
@Test fun importTypeInKDoc() {
      .addType(TypeSpec.classBuilder("Taco")
          .addKdoc("Tacos and [%T]s.\n", String::class)
          .build())
      .build()
    |package com.squareup.tacos
    |
    |/**
    | * Tacos and [kotlin.String]s.
    | */
    |class Taco
    |""".trimMargin())
}
```
This is a safe approach that ensures that types that are not on the classpath don't accidentally get imported, triggering compilation failures. However, when a type is already used somewhere else in the file it's safe to rely on the existing import.

shaishavgandhi05:
The nature of the check `if (!kdoc)` makes me feel like this was probably there for a reason so probably worth a discussion. 🙂 

Looks cleaner with the imports but also more lines. 

Resolves #558

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
@@ -745,4 +745,59 @@ class FileSpecTest {
       |
       |""".trimMargin())
   }
+
+  @Test fun importTypeInKDoc() {
+    val source = FileSpec.builder("com.squareup.tacos", "Taco")
+        .addType(TypeSpec.classBuilder("Taco")
+            .addKdoc("Tacos and [%T]s.\n", String::class)
+            .build())
+        .build()
+    assertThat(source.toString()).isEqualTo("""
+    |package com.squareup.tacos
+    |
+    |import kotlin.String
+    |
+    |/**
+    | * Tacos and [String]s.
+    | */
+    |class Taco
+    |""".trimMargin())
+  }
+
+  @Test fun kdocWithoutImportedType() {
+    val source = FileSpec.builder("com.squareup.tacos", "Taco")
+        .addType(TypeSpec.classBuilder("Taco")
+            .addKdoc("Tacos and [kotlin.String]s.\n")
+            .build())
+        .build()
+    assertThat(source.toString()).isEqualTo("""
+    |package com.squareup.tacos
+    |
+    |/**
+    | * Tacos and [kotlin.String]s.
+    | */
+    |class Taco
+    |""".trimMargin())
+  }
+
+  @Test fun importTypeInKDocWithClashingImports() {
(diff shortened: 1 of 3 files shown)


src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -745,4 +745,59 @@ class FileSpecTest {
...
+    |package com.squareup.tacos
+    |
+    |/**
+    | * Tacos and [kotlin.String]s.
+    | */
+    |class Taco
+    |""".trimMargin())
+  }
+
+  @Test fun importTypeInKDocWithClashingImports() {
```
swankjesse: what’s clashing with what?

shaishavgandhi05: Good point. The kdoc (String) import clashes with the property (String) import. Will rename it to be more explicit.

shaishavgandhi05:
Hmm I guess the thought was that if you have reference to a type, then it's safe to assume it will be on the compile classpath.

Egorand:
I believe the right strategy here is to not import any types when processing KDoc, but to use simple names for types imported elsewhere in the code when emitting KDoc.

shaishavgandhi05:
Yeah that makes way more sense. Will update the logic.

shaishavgandhi05:
> I believe the right strategy here is to not import any types when processing KDoc, but to use simple names for types imported elsewhere in the code when emitting KDoc.

Technically, that already works on master, doesn't it?
```kotlin
@Test fun importTypeInKDoc() {
        .addType(TypeSpec.classBuilder("Taco")
            .addKdoc("Tacos and [%T]s.\n", String::class)
            .addFunction(FunSpec.builder("sendInfo")
                .addParameter("age", Int::class)
                .addKdoc("Has [%T] but no [%T]\n", String::class, Boolean::class)
                .build())
            .build())
        .build()
    |package com.squareup.tacos
    |
    |import kotlin.Int
    |import kotlin.String
    |
    |/**
    | * Tacos and [String]s.
    | */
    |class Taco {
    |    /**
    |     * Has [String] but no [kotlin.Boolean]
    |     */
    |    fun sendInfo(name: String, age: Int) {
    |    }
    |}
    |""".trimMargin())
  }
```

shaishavgandhi05:
Closing this as I'm not sure what the next steps are. From what I understand, the intended behavior in `master` is probably correct as it uses simple names when a type is imported.

Egorand:
Hmm, yeah, looks like it works as expected. I'll close the issue.

PR #569, opened 2018-12-20
Support both @property and @param KDoc tags in TypeSpec

### LINKED ISSUE #566: Support @property KDoc tags in TypeSpec
Right now we're able to emit `@param` tags for primary constructor parameters. We should also emit `@property` for primary constructor parameters that are also properties. `@param` should only be used for parameters that aren't backed by properties.

Egorand:
Closes #566

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -307,7 +307,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
                 ?: constructorProperties[parameterSpec.name]?.kdoc?.takeUnless { it.isEmpty() }
         if (kdoc != null) {
           if (isNotEmpty() && index == 0) add("\n")
-          add("@param %L %L", parameterSpec.name, kdoc.ensureEndsWithNewLine())
+          val isProperty = parameterSpec.name in constructorProperties
+          val tag = if (isProperty) "@property" else "@param"
+          add("$tag %L %L", parameterSpec.name, kdoc.ensureEndsWithNewLine())
         }
       }
       build()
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -307,7 +307,9 @@ class TypeSpec private constructor(builder: TypeSpec.Builder) {
                 ?: constructorProperties[parameterSpec.name]?.kdoc?.takeUnless { it.isEmpty() }
         if (kdoc != null) {
           if (isNotEmpty() && index == 0) add("\n")
-          add("@param %L %L", parameterSpec.name, kdoc.ensureEndsWithNewLine())
+          val isProperty = parameterSpec.name in constructorProperties
+          val tag = if (isProperty) "@property" else "@param"
+          add("$tag %L %L", parameterSpec.name, kdoc.ensureEndsWithNewLine())
```
shaishavgandhi05: We might have to figure out the ordering of the tags in this case. According to the [Android Kotlin Style Guide](https://developer.android.com/kotlin/style-guide#documentation) `@param` should come before `@property`. Without ordering we may be mixing it up with `@param` possible coming between two `@property`?

Egorand: Interesting. My personal preference would be to use the ordering of the constructor params for the respective tags, but we should adhere to the Kotlin/Android style guides whenever possible.

Egorand: I'll merge as-is: there's no mention on the order of those tags in the official Kotlin style guide, and hardcoding specific order based on Android's style guide doesn't seem like a good idea. Making the order of the tags reflect the ordering of the things in code feels like the most meaningful approach, plus, you can make `@param`s come before `@property`s by structuring the class header to reflect that ordering. Happy to change this later based no the feedback.

JakeWharton:
I don't think I like this change. Did you check the rendered documentation? This removes the documentation from the constructor page despite being the intended location.

Egorand:
Good point, should've checked it. I wonder what tags we need to document both the constructor params and the properties.

JakeWharton:
I suspect we need to include both?

Also what happens when you add documentation to the getter as well?

PR #568, opened 2018-12-20
Fix formatting of TypeSpec KDoc

Egorand:
Fixes #563

PR #570, opened 2018-12-21
Move to newer CodeBlock API for emitting new lines

shaishavgandhi05:
Builds on the work done by #568 and uses the same API for `FunSpec`. It's a little trickier and tedious here because of the different types of kdoc tags and the need for adding a breakline before them in case that's where the tags begin.

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -172,12 +172,12 @@ class FunSpecTest {
   @Test fun returnsUnitWithExpressionBody() {
     val funSpec = FunSpec.builder("foo")
         .returns(Unit::class)
-        .addCode(CodeBlock.of("return bar()"))
+        .addStatement("return bar()")
         .build()
 
     assertThat(funSpec.toString()).isEqualTo("""
       |fun foo(): kotlin.Unit = bar()
-      """.trimMargin())
+      |""".trimMargin())
   }
 
   @Test fun functionParamWithKdoc() {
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -259,6 +259,22 @@ class FunSpecTest {
...
+        .addCode("return %S", "foo")
+        .addKdoc("Do the foo")
+        .build()
+    assertThat(funSpec.toString()).isEqualTo("""
+      |/**
+      | * Do the foo
+      | *
+      | * @return the foo.
+      | */
+      |fun foo(nodoc: kotlin.Boolean): kotlin.String = "foo"""".trimMargin())
```
oldergod: Don't we want a new line before `"""`?

swankjesse: Yes please

Egorand: Ouch, this snippet actually looks broken, notice the extra closing bracket at the end of the line... Normally you'd do `.addStatement("return %S", "foo")`, which would emit correct code. Not sure this is worth filing.

shaishavgandhi05: Thanks! I went around and fixed that for some other tests also (written by yours truly, guilty as charged). 
0b9f23257bfa0ec49dac663c93b43ab3d128c83b

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -516,6 +532,24 @@ class FunSpecTest {
       """.trimMargin())
   }
 
+  @Test fun receiverWithKdocAndMainKDoc() {
+    val funSpec = FunSpec.builder("toBar")
+        .receiver(String::class, kdoc = "the string to transform.")
+        .returns(String::class)
+        .addKdoc("%L", "Converts to bar")
+        .addCode(CodeBlock.of("return %S", "bar"))
```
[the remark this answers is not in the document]
shaishavgandhi05: 0b9f23257bfa0ec49dac663c93b43ab3d128c83b

Issue #574, opened 2019-01-02
Add WildcardTypeName example


LouizFC:
As obvious as WildcardTypeName class usage is, I wasted a good portion of time trying to find it.

I think the name itself is not intuitive. I am familiarized with Java, but when in Kotlin the first thing that comes to my mind when using "in and out" is "variance" not "wildcard". 

I think it would be good to add some example using WildcardTypeName just so people can find this class more easely

Egorand:
Thanks for the feedback, an extra piece of documentation is definitely helpful. Good point on the type name, indeed it goes back to JavaPoet's `WildcardTypeName` which is used to generate stuff like `? extends Number`. We should look for a better name.

Egorand:
Closed with #575

PR #575, opened 2019-01-02
Added an example for WildcardTypeName usage

LouizFC:
I added an example. The generated code does raise a warning about "type variance redundancy", but I think it is enough show the wildcard usage mentioned in #574

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,18 +272,39 @@ KotlinPoet has classes for building each of these:
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
 
 val hoverboard = ClassName("com.mattel", "Hoverboard")
+val skateboard = ClassName("com.mattel", "Skateboard")
+val board = ClassName("com.mattel", "Board")
+
 val list = ClassName("kotlin.collections", "List")
 val arrayList = ClassName("kotlin.collections", "ArrayList")
 val listOfHoverboards = list.parameterizedBy(hoverboard)
 val arrayListOfHoverboards = arrayList.parameterizedBy(hoverboard)
+val listOfSkateboards = list.parameterizedBy(skateboard)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,38 +272,71 @@ KotlinPoet has classes for building each of these:
...
     .build()
+
+val notFlyingThings = FunSpec.builder("notFlyingThings")
+    .returns(listOfSkateboards)
+    .addStatement("return listOf(%T(), %T())", skateboard, skateboard)
+    .build()
+
+val things = FunSpec.builder("things")
+    .returns(producerListOfBoards)
+    .addParameter("thingFlies", ClassName("kotlin", "Boolean"))
```
Egorand: nit: `Boolean::class` is a better choice

README.md
```diff
@@ -272,38 +272,71 @@ KotlinPoet has classes for building each of these:
...
+    .beginControlFlow("if(thingFlies) {")
```
Egorand: nit: space after `if` + you don't really need the closing bracket, `if (thingFlies)` will produce the same result

README.md
```diff
@@ -272,38 +272,71 @@ KotlinPoet has classes for building each of these:
...
+    .addStatement("return %N()", flyingThings)
+    .endControlFlow()
```
Egorand: nit: replace `end` and `begin` with `nextControlFlow()`

LouizFC:
@Egorand Thank you for your feedback, I simplified the example, but the IDE still has a warning.
We could use an Array to get the warning out.

I will read thru the CLA later and get back to you. In the worst case, if I end up not signing it for whatever reason, feel free to add an similar example

LouizFC:
I have signed the CLA. Feel free to close this PR.

Egorand:
Thank you!

PR #579, opened 2019-01-02
No line wrapping between val and property name

### LINKED ISSUE #576: Don't break line between val/var and property name
Example of generated code:
```kotlin
@WireField(tag = 2, adapter = "ScheduledTransactionPreference.ADAPTER") val
        scheduled_transaction_preference: ScheduledTransactionPreference? = null,
```

Egorand:
Fixes #576

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -59,7 +59,7 @@ class PropertySpec private constructor(builder: Builder) {
     }
     codeWriter.emitAnnotations(annotations, inline)
     codeWriter.emitModifiers(propertyModifiers, implicitModifiers)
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -26,14 +26,14 @@ class PropertySpecTest {
   @Test fun nullable() {
     val type = String::class.asClassName().copy(nullable = true)
     val a = PropertySpec.builder("foo", type).build()
-    assertThat(a.toString()).isEqualTo("val foo: kotlin.String?\n")
+    assertThat(a.toString()).isEqualTo("val·foo: kotlin.String?\n")
```
swankjesse: I think this is a bug! Shouldn’t we vacuum out the · in the toString() ?

PR #584, opened 2019-01-03
Allow passing line prefix into LineWrapper to enable proper line wrapping in KDoc

### LINKED ISSUE #578: Line breaking inside KDoc produces lines not starting with *
Example:
```kotlin
/**
* The description for the choice, e.g. "Currently unavailable due to high demand. Please
        try
* later." May be null.
*/
fun summary_text(summary_text: String?): Builder {
    this.summary_text = summary_text
    return this
}
```

Egorand:
Fixes #578

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/LineWrapperTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/LineWrapperTest.kt
@@ -186,4 +186,21 @@ class LineWrapperTest {
         | -1
         """.trimMargin())
   }
+
+  @Test fun linePrefix() {
+    val out = StringBuffer()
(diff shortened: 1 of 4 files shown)


src/test/java/com/squareup/kotlinpoet/LineWrapperTest.kt
```diff
@@ -186,4 +186,21 @@ class LineWrapperTest {
         | -1
         """.trimMargin())
   }
+
+  @Test fun linePrefix() {
```
swankjesse: yass

Egorand:
@swankjesse I've extended the test in `KotlinPoetTest` to include KDocs for a class and a function inside it.

PR #588, opened 2019-01-05
Support referencing members using %M and MemberName type

### LINKED ISSUE #433: API to reference top-level Kotlin entities, including package name
It would be great to be able to have a reference to a top-level fun or property from another file, and use it like we currently use TypeNames.  I don't know a good name for such a reference - `TopLevelReference` seems excessively vague, whereas `PropertyOrFunReference` is too Springy.  I'm totally open to better ideas!

This hypothetical reference would let us do something like:

```kt
val reference = Reference("com.foo.bar", "doSomething")
val function = FunSpec.builder("somethingElse")
    .returns(String::class)
    .addStatement("return %R()", reference)
    .build()

println(FileSpec.builder("com.example", "Example")
    .addFunction(function)
    .build())
```

and, hopefully, `CodeWriter` would handle the imports just as it does for `%T` placeholders today:

```kt
// Example.kt
package com.example

import com.foo.bar.doSomething

fun somethingElse(): String = doSomething()
```

At first, I thought this might be accomplished by using a `FunSpec` with the name I wanted, but as it happens `FunSpec` doesn't take a package name.  There doesn't seem to be any way in the current API to do this.

Egorand:
Closes #433

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -49,6 +49,7 @@ internal class CodeWriter constructor(
   private val indent: String = DEFAULT_INDENT,
   private val memberImports: Map<String, Import> = emptyMap(),
   val importedTypes: Map<String, ClassName> = emptyMap(),
+  val importedMembers: Map<String, MemberName> = emptyMap(),
   columnLimit: Int = 100
 ) : Closeable {
   private val out = LineWrapper(out, indent, columnLimit)
@@ -58,8 +59,9 @@ internal class CodeWriter constructor(
   private var comment = false
   private var packageName = NO_PACKAGE
   private val typeSpecStack = mutableListOf<TypeSpec>()
-  private val memberImportClassNames = mutableSetOf<String>()
+  private val memberImportNames = mutableSetOf<String>()
   private val importableTypes = mutableMapOf<String, ClassName>()
+  private val importableMembers = mutableMapOf<String, MemberName>()
   private val referencedNames = mutableSetOf<String>()
   private var trailingNewline = false
 
@@ -74,7 +76,7 @@ internal class CodeWriter constructor(
     for ((memberName, _) in memberImports) {
       val lastDotIndex = memberName.lastIndexOf('.')
       if (lastDotIndex >= 0) {
-        memberImportClassNames.add(memberName.substring(0, lastDotIndex))
+        memberImportNames.add(memberName.substring(0, lastDotIndex))
       }
     }
   }
@@ -233,7 +235,7 @@ internal class CodeWriter constructor(
           if (typeName is ClassName && partIterator.hasNext()) {
             if (!codeBlock.formatParts[partIterator.nextIndex()].startsWith("%")) {
               val candidate = typeName
-              if (candidate.canonicalName in memberImportClassNames) {
+              if (candidate.canonicalName in memberImportNames) {
                 check(deferredTypeName == null) { "pending type for static import?!" }
                 deferredTypeName = candidate
                 defer = true
@@ -242,6 +244,11 @@ internal class CodeWriter constructor(
           }
           if (!defer) typeName.emit(this)
           typeName.emitNullable(this)
+        }
+
+        "%M" -> {
+          val memberName = codeBlock.args[a++] as MemberName
+          memberName.emit(this)
         }
 
         "%%" -> emit("%")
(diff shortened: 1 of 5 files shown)


Egorand:
Extra feature to add: resolving name conflicts by importing enclosing types. E.g.:
```kotlin
com.example.HolyGuac.createTaco
com.example.TacoBell.createTaco
```
would currently resolve into:
```kotlin
import com.example.HolyGuac.createTaco

createTaco()
com.example.TacoBell.createTaco()
```
but can be resolved into:
```kotlin
import com.example.HolyGuac.createTaco
import com.example.TacoBell

createTaco()
TacoBell.createTaco()
```

Egorand:
Need reviews!

src/main/java/com/squareup/kotlinpoet/MemberName.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  fun copy(
+    packageName: String = this.packageName,
+    enclosingClassName: ClassName? = this.enclosingClassName,
+    simpleName: String = this.simpleName
+  ): MemberName {
+    return MemberName(packageName, enclosingClassName, simpleName)
+  }
+
+  override fun equals(other: Any?): Boolean {
```
swankjesse: should this be a data class?

src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -0,0 +1,159 @@
+package com.squareup.kotlinpoet
```
swankjesse: nit: copyright 2019

src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -0,0 +1,159 @@
...
+      |import com.twitter.tacos.TwitterTacos
+      |
+      |fun makeTastyTacos() {
+      |    createTaco()
+      |    TwitterTacos.Companion.createTaco()
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun memberNameAliases() {
```
swankjesse: oh hell yes

src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -0,0 +1,159 @@
...
+      |import com.twitter.tacos.createTaco as createTwitterTaco
+      |
+      |fun makeTastyTacos() {
+      |    createSquareTaco()
+      |    createTwitterTaco()
+      |}
+      |""".trimMargin())
+  }
+
+  @Test fun keywordsEscaping() {
```
swankjesse: this is a great example

src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -0,0 +1,159 @@
...
+      |import org.mockito.`when`
+      |
+      |class TacoTest {
+      |    @Before
+      |    fun setUp() {
+      |        `when`(tacoService.createTaco()).thenReturn(tastyTaco())
+      |    }
+      |}
+      |""".trimMargin())
+  }
```
swankjesse: can you add a test for a collision between a member name and a class name? It’s gross but possible

src/main/java/com/squareup/kotlinpoet/MemberName.kt
Egorand: Yep, no reason for it not to

src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
Egorand: Good call!

src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -360,6 +367,31 @@ internal class CodeWriter constructor(
...
 
+  fun lookupName(memberName: MemberName): String {
+    val simpleName = memberImports[memberName.canonicalName]?.alias ?: memberName.simpleName
+    // Match an imported member.
+    val importedMember = importedMembers[simpleName]
+    if (importedMember == memberName) {
+      return simpleName
+    } else if (importedMember != null && memberName.enclosingClassName != null) {
+      val enclosingClassName = lookupName(memberName.enclosingClassName)
+      return "$enclosingClassName.$simpleName"
```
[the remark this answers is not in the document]
Egorand: Yep, escaping happens inside `MemberName.emit()` after the name has been resolved. Test added.

PR #595, opened 2019-01-11
Allow creating PropertySpecs without an explicit type

ansman:


JakeWharton:
This feature is intentionally omitted. Is there a scenario where you don't actually know the type?

ansman:
For me it's mostly about code style and preventing Kotlin from giving a lint warning that you can omit a type for a local variable:
```kotlin
var foo: Boolean = true
```

JakeWharton:
I'm not sure explicitly declaring a type should ever be punished by style or a linter. Kotlin's style actually recommends the type be always specified for library code and it's hard not to see generated code being closer to library code than user code. And the type serves as a safety mechanism if your initializer is incorrect  with the expected type the code won't compile.

ansman:
I agree with that (ironically enough Kotlinpoet doesn't use explicit types in `PropertySpec` itself though) but for local variables I think defining a type should be optional.

My biggest concern is people opening a generated file and seeing lint warnings, of course I cannot guarantee that the code doesn't have any warnings since I don't know what lint the users have configured but assuming they have the default settings I don't want any warnings.

As a side note I don't think it should be Kotlinpoets job to enforce a certain code style but rather model the language and allow what Kotlin allows (to some extent).

JakeWharton:
To be fair, we didn't design PropertySpec for locals and don't see it as a primary use. I think Egor mentioned that in the other issue. Is something pushing you towards using PropertySpec instead of just `addStatement("var foo = true")`?

Egorand:
I don't like the idea of having less strict API to assist a secondary use case. The primary use case for `PropertySpec` is properties, and we'd like to have their type always declared explicitly. You don't have to use `%L` with `PropertySpec` to have a meaningful abstraction for local variables, I'd suggest introducing your own type to carry relevant data, overriding `toString()` and using template strings to emit its instances.

PR #599, opened 2019-01-13
Add a paragraph on non-wrapping spaces to README

Egorand:


Egorand:
Ideally, it shouldn't be used that often, but there are indeed cases where the client code has to use it.

PR #607, opened 2019-01-20
add 'openjdk11' to Travis build matrix

sullis:


Egorand:
Nice! Does it make sense to keep `oraclejdk8` around, or we can remove that one?

PR #614, opened 2019-02-06
Fix & Test #613

### LINKED ISSUE #613: ParameterizedTypeName.get loses nullability information
```kotlin
ParameterizedTypeName.get(Any::class, true, emptyList())
```
 will return the `TypeName` for `Any`, not for `Any?`.

This is because the `nullable` parameter will be discarded if the `typeArguments` are empty in https://github.com/square/kotlinpoet/blob/master/src/main/java/com/squareup/kotlinpoet/TypeName.kt#L583 : 

```kotlin
  /** Returns a type name equivalent to type with given list of type arguments.  */
    internal fun get(
      type: KClass<*>,
      nullable: Boolean,
      typeArguments: List<KTypeProjection>
    ): TypeName {
      if (typeArguments.isEmpty()) {
        return type.asTypeName()
      }

     […]
```

jGleitz:
fixes #613

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -580,7 +580,7 @@ class ParameterizedTypeName internal constructor(
       typeArguments: List<KTypeProjection>
     ): TypeName {
       if (typeArguments.isEmpty()) {
-        return type.asTypeName()
+        return type.asTypeName().run { if (nullable) copy(nullable = true) else this }
       }
 
       val effectiveType = if (type.java.isArray) Array<Unit>::class else type
--- a/src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt
@@ -30,6 +30,7 @@ import kotlin.reflect.KVariance
 import kotlin.reflect.full.createType
 import kotlin.reflect.full.declaredFunctions
 import kotlin.reflect.full.starProjectedType
+import kotlin.reflect.full.withNullability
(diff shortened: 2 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -580,7 +580,7 @@ class ParameterizedTypeName internal constructor(
       typeArguments: List<KTypeProjection>
     ): TypeName {
       if (typeArguments.isEmpty()) {
-        return type.asTypeName()
+        return type.asTypeName().run { if (nullable) copy(nullable = true) else this }
```
Egorand: Sweet!

src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt
```diff
@@ -193,6 +194,9 @@ abstract class AbstractTypesTest {
         .isEqualTo("kotlin.collections.Map<kotlin.String, *>")
     assertThat(Map.Entry::class.createType(listOf(KTypeProjection(KVariance.INVARIANT, String::class.createType(emptyList())), KTypeProjection.STAR)).asTypeName().toString())
         .isEqualTo("kotlin.collections.Map.Entry<kotlin.String, *>")
+    assertThat(Any::class.starProjectedType.withNullability(true).asTypeName().toString())
+        .isEqualTo("kotlin.Any?")
+
```
Egorand: nit: empty line

jGleitz: removed it.

jGleitz:
I have signed the CLA.

PR #617, opened 2019-02-10
Migrate to Gradle Kotlin DSL

Egorand:


settings.gradle.kts
```diff
@@ -0,0 +1,14 @@
+// TODO(egorand): Remove once gradle-maven-publish-plugin:0.8.0 is published
```
vanniktech: There still seems to be an issue with the dokka documentation. I haven't forgotten about this

Egorand: Yeah, too bad `plugins {}` makes it this hard to consume plugins from outside of the plugin portal. Not a big fan.

Issue #621, opened 2019-02-15
Enum classes with companion object and without entries 


Dominaezzz:
Instead of generating,
```
enum class RenderPassCreate(override val value: Int) {
    ;
    companion object
}
```
it generates,
```
enum class RenderPassCreate(override val value: Int) {
    companion object
}
```

It seems Kotlin requires the semi colon when the enum class has members.
`RenderPassCreate.kt: (7, 89): Expecting ';' after the last enum entry or '}' to close enum class body`

Egorand:
Oh, that's an interesting edge case, thanks for filing! 

Looks like the semicolon is required every time the enum has anything besides constants. Would you be interested in submitting a fix? I also feel like this should be handled better by the compiler, will try to find/file the YouTrack issue.

Egorand:
Closed with #622

PR #622, opened 2019-02-15
Fix edge case with empty enum classes (#621)

Dominaezzz:
This should fix it.
Will a unit test be required?


JakeWharton:
Always!

JakeWharton:
Thank you!

PR #626, opened 2019-02-18
Fix Nullable Type Parameter handling in KType.asTypeName()

### LINKED ISSUE #625: KType.asTypeName() Loses Nullability Information of Type Parameters
Given a `KType` which uses a type Parameter nullably, `KType.asTypeName()` will create a type name where the parameter is used non-nullably. 

I have provided [a failing test case](https://github.com/square/kotlinpoet/pull/626/files#diff-d1a4b945009047c3499f9bc4f89ba605R99) in #626.

jGleitz:
fixes #625

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
@@ -76,17 +76,20 @@ class ParameterizedTypeNameTest {
   }
 
   @Test fun arrayPlusPrimitiveParameter() {
-    val typeName = Array<Int>::class.createType(listOf(KTypeProjection(KVariance.INVARIANT, Int::class.createType()))).asTypeName()
+    val invariantInt = KTypeProjection(KVariance.INVARIANT, Int::class.createType())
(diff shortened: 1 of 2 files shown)


jGleitz:
This PR also includes two more test cases, which are not directly related to #625, but are still meaningful test cases from my point of view.

src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
```diff
@@ -84,6 +85,23 @@ class ParameterizedTypeNameTest {
     assertThat(typeName.toString()).isEqualTo("kotlin.Array<java.io.Closeable>")
   }
 
+  @Test fun arrayPlusNullableParameter() {
+    val typeName = Array<Unit>::class.createType(listOf(KTypeProjection(KVariance.INVARIANT, Closeable::class.createType(nullable = true)))).asTypeName()
```
Egorand: nit: this line looks too long, can you please wrap it to fit into the 100 characters limit?

jGleitz: Done. I also broke the lines in the other, similar test cases.

PR #628, opened 2019-02-20
#629 Demo: Introduce %C for strings that should be compile-time constants

ZacSweers:
This is a proposal solution for addressing #629 via adding a new `%C` placeholder API to `CodeBlock` to indicate that it should use compile-time-constant strings formats (i.e. not templated) for that String. Basically a constant-alternative for `%S`

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
@@ -82,8 +82,8 @@ class AnnotationSpecTest {
     var b = AnnotationSpec.builder(AnnotationC::class.java).build()
     assertThat(a == b).isTrue()
     assertThat(a.hashCode()).isEqualTo(b.hashCode())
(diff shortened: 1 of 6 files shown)


src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -246,16 +246,25 @@ class AnnotationSpecTest {
...
         "@kotlin.Deprecated(\"Nope\", kotlin.ReplaceWith(\"Nope\"))")
   }
 
+  @Test fun stringsAreConstants() {
+    val text = "This is a long string with a newline\nin the middle."
+    val builder = AnnotationSpec.builder(Deprecated::class)
+        .addMember("%C", text)
+
+    assertThat(builder.build().toString()).isEqualTo("" +
+        "@kotlin.Deprecated(\"${text.replace("\n", "\\n")}\")")
```
JakeWharton: Same feedback here as other PR

ZacSweers:
Continued in #654

Issue #629, opened 2019-02-21
Compile-time constant strings incorrectly generated with trimMargin() helper


ZacSweers:
Some places require that `%S` strings be compile-time constants, such as annotations or `const` properties. Right now, if a string has a newline char in it, kotlinpoet will convert it to a raw string template with `.trimMargin()` on the end, which will fail to compile since it's not a compile-time constant

ZacSweers:
I've put up two proposal PRs:

Use a `%C` placeholder: https://github.com/square/kotlinpoet/pull/628
Add `constantStrings` to `CodeBlock`'s API: https://github.com/square/kotlinpoet/pull/630

`%C` makes it very explicit and low overhead for kotlinpoet, `constantStrings` makes it easier for consumers at the expense of overhead in plumbing it down

swankjesse:
Ew. Can we ask the Kotlin folks to do this at build time?

swankjesse:
Ie. this problem isn't really a codegen problem.

ZacSweers:
There's a long standing youtrack for it: https://youtrack.jetbrains.com/issue/KT-17755

But it hasn't seen much activity and has been open for 3 years

JakeWharton:
Seems like we should disable the margin string feature until it's fixed in the compiler

Egorand:
Agree. As a workaround, pass the string as `%L` and escape it manually (e.g. `"\"hello\""`).

ZacSweers:
#654 for the new proposal. I know this is partially addressed in a future kotlin release per https://github.com/JetBrains/kotlin/pull/2159, but also think the new solution is a reasonable (even if temporary) path forward

PR #630, opened 2019-02-21
#629 Demo: Add constantStrings API to CodeBlock

ZacSweers:
This is a proposal solution for addressing #629 via adding a new `constantStrings` API to `CodeBlock` to indicate that it should use compile-time-constant strings formats (i.e. not templated).

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -428,7 +428,7 @@ class CodeBlock private constructor(
       formatParts += codeBlock.formatParts
       args.addAll(codeBlock.args)
       if (codeBlock.constantStrings) {
-        // Constant strings are sticky, if one part wants them then they all get them
+        // Constant strings are sticky, if one part wants them then they all get them.
         constantStrings = true
       }
     }
--- a/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -59,13 +59,22 @@ import kotlin.reflect.KClass
  */
 class CodeBlock private constructor(
   internal val formatParts: List<String>,
-  internal val args: List<Any?>
+  internal val args: List<Any?>,
+  internal val constantStrings: Boolean = false
 ) {
   /** A heterogeneous list containing string literals and value placeholders.  */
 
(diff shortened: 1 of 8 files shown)


src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -758,4 +758,20 @@ class FileSpecTest {
...
+        .addProperty(PropertySpec.builder("FOO", String::class, KModifier.CONST)
+            .initializer("%S", text)
+            .build())
+        .build()
+    assertThat(spec.toString()).isEqualTo("""
+      |package testsrc
+      |
+      |import kotlin.String
+      |
+      |const val FOO: String = "${text.replace("\n", "\\n")}"
```
JakeWharton: I think we'd prefer to see the string here directly. Right now the test relies on the implementation and `replace` and we only want it to depend on the implementation.

src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -256,6 +256,15 @@ class AnnotationSpecTest {
         "@kotlin.Deprecated(\"Nope\", kotlin.ReplaceWith(\"Nope\"))")
   }
 
+  @Test fun stringsAreConstants() {
+    val text = "This is a long string with a newline\nin the middle."
+    val builder = AnnotationSpec.builder(Deprecated::class)
+        .addMember("%S", text)
+
+    assertThat(builder.build().toString()).isEqualTo("" +
+        "@kotlin.Deprecated(\"${text.replace("\n", "\\n")}\")")
```
JakeWharton: Similarly

src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -417,6 +427,10 @@ class CodeBlock private constructor(
     fun add(codeBlock: CodeBlock) = apply {
       formatParts += codeBlock.formatParts
       args.addAll(codeBlock.args)
+      if (codeBlock.constantStrings) {
+        // Constant strings are sticky, if one part wants them then they all get them
```
JakeWharton:
```suggestion
        // Constant strings are sticky, if one part wants them then they all get them.
```

swankjesse:
I feel like we're fighting a compiler bug.

ZacSweers:
Continued in #654

PR #633, opened 2019-02-23
Make *Spec types taggable

ZacSweers:
This is a proposal PR to introduce tagging similar to OkHttp's tagging APIs. The idea here is that for an API that uses KotlinPoet as an IR, this allows consumers to tag arbitrary information to read at a later point. This also allows for pushing more information into individual specs (such as originating elements) without bolting them to the top-level API or coupling them to those frameworks.

One hypothetical example: Say Moshi's code gen was abstracted to be usable in non-apt environments. Right now there's a good deal of error handling and reporting in it that depends on having elements available to read from, but we could represent those elements as KotlinPoet types instead and track the original elements in tags for the processor's usage later.

From a high level - these are backed as a `Map<KClass<*>, Any>`. There's overloads in builders for `Class` and no-arg. There's no overload for the no-arg with `Any` though, as Java users will just see that as an `Object` anyway.

Usage example:

```kotlin
val function = FunSpec.builder("name")
    .tag<String>("my tag")
    // or .tag(String::class, "my tag")
    // or .tag(String::class.java, "my tag")
    .build()

val myTag = function.tag<String>()
    // or .tag(String::class)
    // or .tag(String::class.java)
assertThat(myTag).isEqualTo("my tag")
```

If this sounds reasonable, I can start on adding some tests.


Egorand:
I feel like this proposal targets a very specific use case and the proposed APIs will not make sense to the majority of the library users. However, I'm not familiar with OkHttp's tagging API and I might be missing context here. However, I also think that using KotlinPoet's (or any other 3rd party's) types directly as IR is not a good idea, cause you're coupling your code to those types - hence the need for those types to have properties specific to your use case. Would you consider doing this instead?
```
data class Function(
  val spec: FunSpec,
  val tags: Map<KClass<*>, Any>
)
```

ZacSweers:
I think KotlinPoet makes a great IR! @swankjesse and I had talked in the past about trying to make an APT-ish API for Kotlin with KotlinPoet as the type system. Having the toBuilder() APIs in particular is what enables it for IR cases, and we've used it to pretty good success for some major internal projects.

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -30,6 +30,25 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
...
+   */
+  fun tag(): Any? = tag(Any::class)
+
+  /** Returns the tag attached with [type] as a key, or null if no tag is attached with that key. */
+  fun <T : Any> tag(type: Class<out T>): T? = tag(type.kotlin)
+
+  /** Returns the tag attached with [type] as a key, or null if no tag is attached with that key. */
+  fun <T : Any> tag(type: KClass<out T>): T? {
+    @Suppress("UNCHECKED_CAST")
+    return tags[type] as? T
```
JakeWharton:
```suggestion
    return tags[type] as T?
```

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -30,6 +30,25 @@ class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
   val className: ClassName = builder.className
   val members = builder.members.toImmutableList()
   val useSiteTarget: UseSiteTarget? = builder.useSiteTarget
+  private val tags: Map<KClass<*>, Any> = builder.tags.toImmutableMap()
+
+  /**
+   * Returns the tag attached with [Any] as a key, or `null` if no tag is attached with that key.
+   */
+  fun tag(): Any? = tag(Any::class)
```
JakeWharton: I don't think we need this variant nor the builder method. Its use breaks composition of tags which seems like the only reason to have them given they're public API.

Same comment on all specs

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
JakeWharton: Same change on all specs

JakeWharton: Actually, `type.safeCast(tags[type])` is probably best here.

ZacSweers: Hm... I think your first suggestion (`tags[type] as T?`) is the right thing to do actually. I was trying to just make it null-safe but throw on type mismatches since they're programmer error. What do you think?

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
ZacSweers: 2c91a6b

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
JakeWharton: `safeCast` is the safer version of that because it actually performs a cast at runtime instead of being erased into a no-op cast to `Object`.

ZacSweers:
I started with that approach, but you can't have inline functions in an interface. Not sure if there would be confusion if we split the functions by having inline ones in the class and the rest in the interface. What do you think?

Egorand:
Ah, then I think it's fine to leave as is. Would there be a use-case where you'd want to process a collection of `Taggable` specs?

ZacSweers:
Actually I've got an idea. Will report back 👀

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
ZacSweers: Right, but it will always fail the cast silently and just return null rather than throw it back to the caller. Is that what we want?

image

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -26,7 +26,10 @@ import javax.lang.model.util.SimpleAnnotationValueVisitor7
 import kotlin.reflect.KClass
 
 /** A generated annotation on a declaration.  */
-class AnnotationSpec private constructor(builder: AnnotationSpec.Builder) {
+class AnnotationSpec private constructor(
+    builder: AnnotationSpec.Builder,
```
Egorand: nit: 2 space indent, same for all spec classes

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
Egorand: If we want to return null for missing keys then this is fine, otherwise we can do
```
type.safeCast(tags[type]) ?: throw NoSuchElementException("Missing tag for type $type")
```

JakeWharton: Wow safeCast does not do what I thought it did. And cast is kinda shitty. I guess I'm looking for `?.let(KClass::cast)` where there's an actual type check that tolerates null.

ZacSweers:
Updated with tests and fell on to the `cast only if nonnull` approach, meaning that presence will try to cast and null will skip the cast. I did have to tweak the generics a bit as I unintentionally set them up before to require the key and value be of the same type, which wasn't intended (but let me know if you think it should?). This means the storing tag API has two reified generics, which could be potentially confusing, so I've also renamed the generics to be more explicit `KeyType` and `ValueType` where appropriate.

I used a parameterized test runner to cover all the taggable types' flows, but let me know if you'd rather a different approach for these tests!

JakeWharton:
Why would you want the key and value types to be different? It makes no sense to restrict it to a class anymore for that. It could just be an opaque object like NameAllocator. That being said, I prefer the old behavior.

ZacSweers:
Fair point. I couldn't really think of a good example of why that would be useful after thinking on it. Switched back in 9503327

src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
ZacSweers: 386254c

Worth mentioning though that my IDE doesn't see a difference between primary constructor param indents and continuation indent (using the square intelliJ code style too)

PR #634, opened 2019-02-24
Attach primary constructor param/property KDoc to the element vs emitting it inside the type header

Egorand:


src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1208,16 +1208,21 @@ class TypeSpecTest {
...
+        |     * the Andes, or hot with lava-like cheese from the depths of
+        |     * the Ninth Circle.
+        |     */
         |    temperature: Double,
+        |    /**
+        |     * True for a soft flour tortilla; false for a crunchy corn tortilla.
+        |     */
         |    val soft: Boolean,
+        |    /**
+        |     * No one likes mild tacos.
```
jrodbx: hmm, would it be better to override the parameter kdoc with any property kdoc, possibly throwing an exception to alert the author of this potential conflict?

Egorand: Decided to go with putting parameter KDoc under the type/primary constructor header if there's a conflict.

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1190,7 +1190,7 @@ class TypeSpecTest {
             .initializer("soft")
             .build())
         .addProperty(PropertySpec.builder("mild", Boolean::class)
-            .addKdoc("This property doc comment should be overridden.\n")
+            .addKdoc("No one likes mild tacos.")
```
jrodbx: facts

PR #636, opened 2019-03-03
import members from default package

fluidsonic:
In order to use use symbols from the default package in Kotlin they must be imported.


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -501,7 +501,7 @@ class KotlinPoetTest {
             .initializer("%S", "a")
             .build())
         .addType(TypeSpec.classBuilder("B")
-            .superclass(ClassName("", "A"))
+            .superclass(ClassName("com.squareup.tacos", "A"))
```
Egorand: nit: `tacosPackage` instead of hardcoded package name?

fluidsonic: oh nice, done 👍

PR #637, opened 2019-03-04
Add some tests around jvmModifiers

shaishavgandhi05:
Unrelated: I'm not sure if this is intentional, but `jvmModifiers` in `FunSpec` doesn't return the `Builder` and breaks the chaining. There is a `jvmModifiers` method in `ParameterSpec`  which does return the Builder. I can fix that in a different PR if that was an oversight.


Egorand:
Yeah, not returning `Builder` looks like an oversight. It's a binary-incompatible change though, would be great to file a bug and we'll fix it for 2.0.

PR #641, opened 2019-03-07
Add non-wrapping spaces in control flow creation methods

Egorand:


Egorand:
Adding tests in a follow-up

PR #648, opened 2019-03-14
Fix inline Taggable builder APIs by not breaking builder chain

ZacSweers:
This was an unfortunate bug that I noticed too late - we can't have a general-use `inline` extension function for the builder tag APIs without losing context of the builder type. Open to suggestions on this, but haven't been able to figure out any other way right now. This breaks out the inline extensions to just be overloads for the different subclasses, but at the cost of not being able to make it something that each consumer control.


ZacSweers:
Good call, added in 7799a9d

PR #653, opened 2019-03-27
Add FunSpec#jvmModifiers overload to fix chaining

### LINKED ISSUE #638: `jvmModifiers` doesn't return the Builder
The method `jvmModifiers` in `FunSpec` breaks the chaining by not returning the `Builder`. There is a similar method in `ParameterSpec` which does return the `Builder`. 

This will be a breaking change so it will have to wait till 2.x.

ZacSweers:
Resolves #638

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -28,6 +28,7 @@ import javax.lang.model.type.DeclaredType
 import javax.lang.model.type.ExecutableType
 import javax.lang.model.type.TypeVariable
 import javax.lang.model.util.Types
+import kotlin.DeprecationLevel.HIDDEN
 import kotlin.reflect.KClass
 
 /** A generated function declaration.  */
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -286,7 +287,19 @@ class FunSpec private constructor(
       this.modifiers += modifiers
     }
 
-    fun jvmModifiers(modifiers: Iterable<Modifier>) {
+    @JvmName("jvmModifiers")
+    @Deprecated(
+        message = "This API was missing the Builder return type and breaks chaining. " +
```
Egorand: The message here should target library users, not maintainers, something like "Deprecated. Use jvmModifiersNew()." would do.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -286,7 +287,19 @@ class FunSpec private constructor(
...
+            "https://github.com/square/kotlinpoet/issues/638",
+        replaceWith = ReplaceWith("jvmModifiers(modifiers)"),
+        level = HIDDEN
+    )
+    fun jvmModifersOld(modifiers: Iterable<Modifier>) {
```
Egorand: typo: `Modifers` -> `Modifiers`

ZacSweers:
Talked offline, will leave to a future 2.x

PR #654, opened 2019-03-27
Introduce notion of "constant context" to CodeWriter

### LINKED ISSUE #629: Compile-time constant strings incorrectly generated with trimMargin() helper

### LINKED ISSUE #649: Applying annotation to a single statement seems to be not possible
I am trying to wrap my head around emitting code which would look like this:

```kotlin
fun operation() {
  @Supress("UNCHECKED_CAST")
  val value = otherValue as SomeClass<*>
  println(value)
}
```
So far I can't find a good way to do this without pasting literal text (which would not insert correct imports)

Neither `addStatement()` nor `CodeBlock.builder()` do not accept `AnnotationSpecs`, and there's no `%X` format modifier for annotations.

Is this not supported? 
What would you recommend to do in this situation?

ZacSweers:
This is a successor to #630 and #628, with the goal of resolving a couple issues around constant context emissions.

There are two use cases right now:
1 - Strings that should be constant (such as for `const` or as annotation members). Ref - #629
2 - Better supporting literal annotations emission depending on the context. As an annotation member, they need to be constant and can omit the `@` sign. As a literal in a `CodeBlock` that isn't in this context, they need the `@` sign. Adding this clarifies that context and can be controlled at the `AnnotationSpec` level. Ref - #649

By only adding this as an added parameter to `CodeWriter#emitCode()`, this is far less invasive than #630. It also doesn't do #630's property type validation for `KModifier.CONST` and instead just uses the presence of `CONST` in its modifiers as the sole heuristic, with the idea being that the developer should know if the type they're using can't be `const` and if they don't then the compiler will tell them later :).

Resolves #629
Resolves #649

I think this kills two birds with one stone quite nicely. Let me know what you all think!

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
@@ -256,6 +256,15 @@ class AnnotationSpecTest {
         "@kotlin.Deprecated(\"Nope\", kotlin.ReplaceWith(\"Nope\"))")
   }
 
+  @Test fun annotationStringsAreConstant() {
+    val text = "This is a long string with a newline\nin the middle."
+    val builder = AnnotationSpec.builder(Deprecated::class)
+        .addMember("%S", text)
+
+    assertThat(builder.build().toString()).isEqualTo("" +
+        "@kotlin.Deprecated(\"This is a long string with a newline\\nin the middle.\")")
+  }
+
   private fun toString(annotationSpec: AnnotationSpec) =
       toString(TypeSpec.classBuilder("Taco").addAnnotation(annotationSpec).build())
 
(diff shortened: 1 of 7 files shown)


src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -256,6 +256,54 @@ class AnnotationSpecTest {
...
+    assertThat(file.toString().trim()).isEqualTo("""
+      package test
+
+      import kotlin.Suppress
+
+      fun test() {
+          @Suppress("Things")
+          val annotatedString = "AnnotatedString"
+      }
+    """.trimIndent())
```
Egorand: Just for consistency with the rest of the codebase, can you please switch to `trimMargin()`?

ZacSweers: Done 56e3896

PR #655, opened 2019-03-27
Fix named "value" argument being omitted in annotation array types

### LINKED ISSUE #651: value parameter in AnnotationSpec not explicitly written with an array
Hello, 

When trying to add an annotation to a type, when the `value` parameter of the annotation is  an array, the `AnnotationSpec` skip the parameter name (`value = `).

So it produce this kind of code which is not valid:  

```
@XmlSeeAlso([Object::class, Boolean::class])
class Result
```

It would expected instead: 

```
@XmlSeeAlso(value = [Object::class, Boolean::class])
class Result
```

or 

```
@XmlSeeAlso(Object::class, Boolean::class)
class Result
```

Do you confirm that it's a bug?
You can reproduce the case using https://github.com/dwursteisen/kotlinpoet (I added a test)

Please note that I fix it by adding the parameter name during the generation of the annotation.
(see https://github.com/dwursteisen/kotlinpoet/commit/f07e3577c229d40b1a9c18d97398c285594d6d88#diff-063f0c8c9066c2227d34ae844f895fc2R164). It was the easier solution to implement for me, and check that it's working on my project.

Regards

ZacSweers:
Fixes #651

### DIFF
--- a/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
+++ b/src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
@@ -338,15 +338,15 @@ class AnnotationSpecTest {
         }
 
     assertThat(toString(classBuilder.build())).isEqualTo("""
-            |package com.squareup.tacos
-            |
-            |import com.squareup.kotlinpoet.JavaClassWithArrayValueAnnotation
-            |import java.lang.Object
-            |import kotlin.Boolean
-            |
-            |@JavaClassWithArrayValueAnnotation.AnnotationWithArrayValue(value = [Object::class, Boolean::class])
-            |class Result
-            |""".trimMargin())
+        |package com.squareup.tacos
+        |
+        |import com.squareup.kotlinpoet.JavaClassWithArrayValueAnnotation
+        |import java.lang.Object
+        |import kotlin.Boolean
+        |
+        |@JavaClassWithArrayValueAnnotation.AnnotationWithArrayValue(value = [Object::class, Boolean::class])
+        |class Result
+        |""".trimMargin())
   }
 
   @Test fun getOnValueArrayTypeAnnotationShouldNameValueArg() {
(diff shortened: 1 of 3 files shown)


src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -256,6 +256,93 @@ class AnnotationSpecTest {
...
+        .getTypeElement(JavaClassWithArrayValueAnnotation::class.java.canonicalName)
+    val classBuilder = TypeSpec.classBuilder("Result")
+
+    myClazz.annotationMirrors.map { AnnotationSpec.get(it) }
+        .forEach {
+          classBuilder.addAnnotation(it)
+        }
+
+    assertThat(toString(classBuilder.build())).isEqualTo("""
+            |package com.squareup.tacos
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
```diff
@@ -256,6 +256,93 @@ class AnnotationSpecTest {
...
+    val classBuilder = TypeSpec.classBuilder("Result")
+
+    myClazz.annotationMirrors.map { AnnotationSpec.get(it) }
+        .filter { it.className.simpleName == "AnnotationWithArrayValue" }
+        .forEach {
+          classBuilder.addAnnotation(it)
+        }
+
+    assertThat(toString(classBuilder.build())).isEqualTo("""
+            |package com.squareup.tacos
```
Egorand: nit: 4-space indent

src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
ZacSweers: Part of dad2308

src/test/java/com/squareup/kotlinpoet/AnnotationSpecTest.kt
ZacSweers: Part of dad2308

PR #664, opened 2019-04-07
Demo/proposal: FormattingFiler

ZacSweers:
Inspired by GJF's implementation of the same name - this demos an implementation of a `FormattingFiler` that formats kotlin code with [KtLint](https://github.com/pinterest/ktlint) while writing. This could also be applicable to the other `writeTo()` overloads, but would require slightly different support.

I _don't_ think this should be part of the first party KotlinPoet API, but it could be a useful extension artifact. This filer-only API alone may be more valid as an extension artifact on KtLint's project, but as mentioend above the implementation could be useful for other `writeTo()` overloads too.

Example usage:

```kotlin
val formattingFiler = filer.asFormatting()

fileSpec.writeTo(formattingFiler)
```

Possible other `writeTo()` overloads could be `FileSpec.formattedWriteTo()` extension functions, but would likely want to reuse the directory preparation logic.

Happy to also just put all this in a separate project and call it a day. Talked with @Egorand offline and wanted to give first refusal here :)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,6 +27,9 @@ afterEvaluate {
 
 dependencies {
   api("org.jetbrains.kotlin:kotlin-stdlib-jdk7")
+  implementation("com.github.shyiko.ktlint:ktlint-core:0.31.0")
+  implementation("com.github.shyiko.ktlint:ktlint-ruleset-standard:0.31.0")
+  implementation("com.github.shyiko.ktlint:ktlint-ruleset-experimental:0.31.0")
   implementation("org.jetbrains.kotlin:kotlin-reflect")
   testImplementation("org.jetbrains.kotlin:kotlin-test-junit")
   testImplementation("com.google.truth:truth:0.42")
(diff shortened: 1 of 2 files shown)


JakeWharton:
For those who think kotlinc isn't slow enough?

ZacSweers:
For those that generate models in separate, heavily-cached non-gradle subprojects

Egorand:
Agree that this shouldn't live in the main artifact, but we can definitely move this into a separate one. We do plan to improve Kotlinpoet's own formatting though.

ZacSweers:
Do you want me to set up a separate artifact as part of this? Or discuss it further/separately first?

ZacSweers:
I envision this to be more useful for consumers that are implementing possibly complex `CodeBlock`s, and not necessary as a replacement for KotlinPoet's own standard formatting for standard types :)

Egorand:
Yep, let's set up a separate artifact.

LouizFC:
> We do plan to improve Kotlinpoet's own formatting though.

@Egorand could you elaborate please? Do you plan to use Kotlin official formatting guide as a base or this is just a backlog idea that isn't put much thought into?

Egorand:
@LouizFC KotlinPoet already produces reasonably well-formatted code, but there are some issues we're aware of, such as function and constructor signature not being formatted according to the official style guide. This issue specifically is a top priority and I'm hoping to address it soon, if there's anything else that doesn't work for you - please file bugs.

swankjesse:
I'd prefer to not fix our formatting bugs by introducing support for pluggable formatters. One thing we learned with Dagger 2 is that nobody wants to wait for a formatting step.

ZacSweers:
Going to close this for now as I think the lack of any direct touch points with KotlinPoet means it probably should just be a separate project

PR #668, opened 2019-04-08
Add ClassName.constructorReference() and MemberName.reference()

### LINKED ISSUE #663: Support callable references
I frequently get tripped up generating function references due to a combination of the syntax required and how KotlinPoet handles imports.

For example, say I would like to generate a reference to the `Taco` constructor:

```kotlin
import com.example.Taco

val tacoFactory: () -> Taco = ::Taco
```

In KotlinPoet, I would try this, and it produces the expected output:

```kotlin

val TACO = ClassName("com.example", "Taco")

val file = FileSpec.builder("", "Taco.kt")
    .addProperty(
        PropertySpec.builder("tacoFactory", LambdaTypeName.get(null, emptyList(), TACO))
            .initializer("::%T", TACO)
            .build()
    )
    .build()
```

However, if Taco is a nested class, this doesn't work, because the syntax requires the `::` to come *after* the enclosing type. Replacing `TACO` above with `ClassName("com.example", "Food", "Taco")`:

```kotlin
import com.example.Food

val tacoFactory: () -> Food.Taco = ::Food.Taco      // This should be Food::Taco
```

To get this to work in the general case (e.g. for any `ClassName`, nested or not) requires a bunch of boilerplate that could be handled by KotlinPoet itself.

```kotlin
fun ClassName.constructorReference(): CodeBlock {
    val enclosingClassName = enclosingClassName()
    return if (enclosingClassName != null) {
        CodeBlock.of("%T::%L", enclosingClassName, simpleName)
    } else {
        CodeBlock.of("::%T", this)
    }
}
```

Of course, the `else` case isn't semantically correct, as it only works when KotlinPoet is able to import the type and use it without the qualifying package. I'm not sure how the above code could be modified to automatically import the non-nested type if I only use it in a constructor reference.

In addition to constructor references, there are also function and property references, which are syntactically similar to `MemberName`, but with the reference operator required before the member name in both qualified and unqualified usages. These are similarly difficult to generate if they are top-level members, as KotlinPoet wants to both emit the name and handle the import.

tadfisher:
Fixes #663 

Add support for emitting [callable references](https://kotlinlang.org/docs/reference/reflection.html#callable-references) for `ClassName` and `MemberName`. These emit names prefixed with the enclosing class name, if any, and the reference operator `::`.

Similar to `MemberName` references for top-level classes and members are not automatically aliased.


README.md
```diff
@@ -1203,6 +1203,49 @@ typealias FileTable<K> = Map<K, Set<File>>
 typealias Predicate<T> = (T) -> Boolean
 ```
 
+### Callable References
```
Egorand: Big kudos for the docs!

README.md
```diff
@@ -1203,6 +1203,49 @@ typealias FileTable<K> = Map<K, Set<File>>
...
+import com.example.hello.Hello
+
+fun factories() {
+  val hello = ::Hello
+  val world = Hello::world
+  val bye = Hello.World::bye
+}
+```
+
+Top-level classes and members with conflicting names may require aliased imports, as with
```
Egorand: Would be great to have tests covering these scenarios.

PR #669, opened 2019-04-09
Use %N internally to escape names

Egorand:


src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -338,7 +338,7 @@ class CodeBlock private constructor(
 
     private fun addArgument(format: String, c: Char, arg: Any?) {
       when (c) {
-        'N' -> this.args += argToName(arg).escapeIfKeyword()
+        'N' -> this.args += argToName(arg).escapeIfNecessary()
```
JakeWharton: Is there a test missing here? Something that wasn't escaped before but will be now?

Egorand: Yeah, will add tests for all specs that started using `%N`

Egorand: Actually the behavior of the specs hasn't changed, they're just going through the same mechanism to ensure names are escaped correctly. Added tests for `%N` usage.

PR #679, opened 2019-04-10
Fix bug with MemberNames in same package nested inside a class.

lowasser:
The test case pretty accurately reflects the bug I was encountering: within the same package, we still need to qualify members that are part of an enclosing class.


lowasser:
Done.

PR #680, opened 2019-04-12
Pass only unique originating elements to Filer

technoir42:
If let's say we generate 20 top-level elements from a single originating element there is no point in passing that originating element to Filer 20 times. It's inconvenient during debugging/logging where you are only interested in unique elements and [breaks](https://github.com/JetBrains/kotlin/blob/6381b3b1515fb8f1115610e7019fb7bd30bcbc6a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/incrementalProcessors.kt#L108-L109) incremental processing (which will be properly fixed in kapt).

### DIFF
--- a/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -99,8 +99,10 @@ class FileSpec private constructor(
   /** Writes this to `filer`.  */
   @Throws(IOException::class)
   fun writeTo(filer: Filer) {
-    val originatingElements = members.filterIsInstance<OriginatingElementsHolder>()
-        .flatMap(OriginatingElementsHolder::originatingElements)
+    val originatingElements = members.asSequence()
+        .filterIsInstance<OriginatingElementsHolder>()
+        .flatMap { it.originatingElements.asSequence() }
+        .toSet()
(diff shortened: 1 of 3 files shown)


technoir42:
I signed it.

src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -99,8 +99,10 @@ class FileSpec private constructor(
   /** Writes this to `filer`.  */
   @Throws(IOException::class)
   fun writeTo(filer: Filer) {
-    val originatingElements = members.filterIsInstance<OriginatingElementsHolder>()
-        .flatMap(OriginatingElementsHolder::originatingElements)
+    val originatingElements = members.asSequence()
+        .filterIsInstance<OriginatingElementsHolder>()
+        .flatMap { it.originatingElements.asSequence() }
```
JakeWharton: Why are you creating a sequence here? The entire collection will always be consumed.

technoir42: `Sequence` has only one (this) variant of `flatMap`. What would you suggest instead?
I used `members.asSequence()` to avoid allocating temporary lists.

JakeWharton: Ah, I missed the original asSequence call on the outer chain. Makes sense.

PR #687, opened 2019-04-21
Improve code and KDoc formatting in NameAllocator

Egorand:


Egorand:
Yep, that'd be great. Wanna take a stab at it?

shaishavgandhi05:
Yup, will do!

PR #689, opened 2019-04-21
Integrate Spotless

shaishavgandhi05:
5f58b3ce49d7fb8a9aafd141946a0340995894bf Integrates it into the build file and then subsequent commits are just applying it over the repo.


build.gradle.kts
```diff
@@ -25,6 +27,15 @@ afterEvaluate {
   }
 }
 
+spotless {
+  kotlin {
+    target("**/*.kt")
+    ktlint("0.31.0").userData(mapOf("indent_size" to "2", "continuation_indent_size" to "2"))
```
Egorand: Hmm, shouldn't `continuation_indent_size` be 4?

src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt
```diff
@@ -211,16 +211,16 @@ abstract class AbstractTypesTest {
...
+        throw UnsupportedOperationException()
 
-    override fun <A : Annotation> getAnnotation(annotationType: Class<A>): A
-        = throw UnsupportedOperationException()
+    override fun <A : Annotation> getAnnotation(annotationType: Class<A>): A =
+        throw UnsupportedOperationException()
 
-    override fun getAnnotationMirrors()
-        = throw UnsupportedOperationException()
+    override fun getAnnotationMirrors() =
```
Egorand: nit: this can fit on a single line

src/test/java/com/squareup/kotlinpoet/TestFiler.kt
```diff
@@ -53,27 +53,36 @@ internal class TestFiler(
...
   }
 
   override fun getResource(
-      location: JavaFileManager.Location, pkg: CharSequence, relativeName: CharSequence)
-      = throw UnsupportedOperationException("Not implemented.")
+    location: JavaFileManager.Location,
+    pkg: CharSequence,
+    relativeName: CharSequence
+  ) =
+      throw UnsupportedOperationException("Not implemented.")
```
Egorand: nit: merge this with previous line

build.gradle.kts
ZacSweers: I agree, but FYI that'll nuke your 2space indented primary constructor args that we talked about on another PR

shaishavgandhi05: Basically, yeah. [OkHttp also has that same size](https://github.com/square/okhttp/blob/master/build.gradle#L114) because of the primary constructor args.

shaishavgandhi05:
> Note this will rewrite existing copyrights to the current year. Not configurable currently
I've left out the copyright rule for Spotless. I had some problems configuring it with Gradle kts and it would also overwrite the copyright year, so I left it out.

src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt
shaishavgandhi05: 5fca64f772bad7052c1836d3fde8e485210d4395

src/test/java/com/squareup/kotlinpoet/TestFiler.kt
shaishavgandhi05: 5fca64f772bad7052c1836d3fde8e485210d4395

build.gradle.kts
Egorand: That sounds like a Spotless/KtLint bug

shaishavgandhi05: Seems that IntelliJ does the same, so i'm not sure if it's a bug. 

Screen Shot 2019-04-22 at 11 42 35 AM

Notice the `multilineMethod` with 4 continuation indent.

Egorand: Weeeeird.. The official style guide says that class header params should use same indentation as the properties inside the class, and I believe that function headers follow the same rules: https://git.sqcorp.co/projects/CASH/repos/cash-android/pull-requests/8428/overview

Anyway, fine with leaving this as `2` as long as the code passes the checks. Might change later.

shaishavgandhi05: Hmm that's interesting. I'll file a bug with ktlint soon and report it here.

ZacSweers: Fwiw I think it would be better to use 4 space for now since most everything else should be using that

Egorand: Would the check still pass?

ZacSweers: It will after the developer runs spotlessApply to autoformat :)

shaishavgandhi05: Hmm looks like changing it to 4 doesn't do anything 🤷‍♂️ 

Some relevant threads:
[Seems like it was removed in 0.22](https://github.com/pinterest/ktlint/blob/8ddea8d60d75ced601cd7c6d44ecff0b971e07eb/CHANGELOG.md#0220---2018-04-22)

Egorand: Should we just remove the rule then? I think it'll confuse people going forward.

shaishavgandhi05: 5bb2a5d13847b355d93e147f321caa42790fa103

Tested by creating a new class like so:
```kotlin
class SampleClass(
    val something: String
)
```
which triggered a spotless violation. Upon running `./gradlew spotlessApply` it corrected to:
```kotlin
class SampleClass(
  val something: String
)
```

Egorand: Sweet, :shipit:!

PR #691, opened 2019-04-21
Edit MemberName docs to not give an impression that it can only be us…

Egorand:
…ed for static context members

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,10 +385,10 @@ class HelloWorld {
 
 ### %M for Members
 
-Similar to types, KotlinPoet has a special placeholder for **members** (functions and properties) 
-that are accessible in static context - this includes top-level members and members declared inside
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -385,10 +385,10 @@ class HelloWorld {
 
 ### %M for Members
 
-Similar to types, KotlinPoet has a special placeholder for **members** (functions and properties) 
-that are accessible in static context - this includes top-level members and members declared inside
-objects. Use **`%M`** to reference members, pass an instance of `MemberName` as the argument for 
-the placeholder, and KotlinPoet will handle imports automatically:
+Similar to types, KotlinPoet has a special placeholder for **members** (functions and properties),
+which comes handy when your code needs to access  top-level members and members declared inside
```
JakeWharton:
```suggestion
which comes handy when your code needs to access top-level members and members declared inside
```

PR #696, opened 2019-05-27
Force new lines when emitting primary constructors

Egorand:


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -718,9 +722,11 @@ class KotlinPoetTest {
...
       |import com.example.SomeTypeAlias
       |
-      |class Taco(val foo: SomeTypeAlias = { arg: kotlin.Any ->
-      |  println("arg=${'$'}arg")
-      |}
+      |class Taco(
+      |  val foo: SomeTypeAlias = { arg: kotlin.Any ->
+      |    println("arg=${'$'}arg")
+      |  }
+      |
```
Egorand: This looks like a bug, will fix separately

Egorand:
Taking the path of least resistance, at least temporarily, cause solving this on the `LineWrapper` level is hard.

PR #697, opened 2019-05-27
Don't inline annotations in the primary constructor

Egorand:


src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -797,7 +798,11 @@ class KotlinPoetTest {
       |import kotlin.String
       |
       |data class Taco(
-      |  @WireField(tag = 1, adapter = "CustomStringAdapterWithALongNameThatCauses") val name: String
+      |  @WireField(
```
Egorand: This is how Wire will look

oldergod: YES!

PR #710, opened 2019-06-22
Expression Function for `throw`

### LINKED ISSUE #703: Generate single expression functions when the body starts with "throw"
Similar to `return`, we can omit the brackets if the body of the function starts with the `throw` keyword:
```kotlin
FunSpec.builder("newBuilder")
  .addStatement("throw %T()", AssertionError::class)
  .build()
```
now produces:
```kotlin
fun newBuilder() {
  throw AssertionError()
}
```
while it can be simplified to:
```kotlin
fun newBuilder() = throw AssertionError()
```

ShaishavGandhi:
Resolves #703

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -97,7 +97,7 @@ class FunSpec private constructor(
       return
     }
 
-    val asExpressionBody = asExpressionBody(body)
+    val asExpressionBody = body.asExpressionBody()
 
     if (asExpressionBody != null) {
       codeWriter.emitCode(" = %L", asExpressionBody)
@@ -108,17 +108,6 @@ class FunSpec private constructor(
       codeWriter.unindent()
       codeWriter.emit("}\n")
     }
-  }
-
-  private fun asExpressionBody(codeBlock: CodeBlock): CodeBlock? {
-    val asReturnExpressionBody = codeBlock.trim().withoutPrefix(RETURN_EXPRESSION_BODY_PREFIX)
-    if (asReturnExpressionBody != null) {
-      return asReturnExpressionBody
-    }
-    if (codeBlock.trim().withoutPrefix(THROW_EXPRESSION_BODY_PREFIX) != null) {
-      return codeBlock
-    }
-    return null
   }
 
   private fun emitSignature(codeWriter: CodeWriter, enclosingName: String?) {
@@ -201,13 +190,21 @@ class FunSpec private constructor(
    */
   private fun emitReturnType(returnType: TypeName?): Boolean {
     if (returnType != null) {
-      return returnType != Unit::class.asTypeName() || isExpressionBody()
+      return returnType != Unit::class.asTypeName() || body.asExpressionBody() != null
     }
     return false
   }
 
-  private fun isExpressionBody(): Boolean {
-    return asExpressionBody(body) != null
+  private fun CodeBlock.asExpressionBody(): CodeBlock? {
+    val codeBlock = this.trim()
+    val asReturnExpressionBody = codeBlock.withoutPrefix(RETURN_EXPRESSION_BODY_PREFIX)
+    if (asReturnExpressionBody != null) {
+      return asReturnExpressionBody
+    }
+    if (codeBlock.withoutPrefix(THROW_EXPRESSION_BODY_PREFIX) != null) {
+      return codeBlock
+    }
+    return null
   }
 
   override fun equals(other: Any?): Boolean {
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -97,7 +97,7 @@ class FunSpec private constructor(
       return
     }
 
-    val asExpressionBody = body.trim().withoutPrefix(EXPRESSION_BODY_PREFIX)
+    val asExpressionBody = asExpressionBody(body)
 
     if (asExpressionBody != null) {
       codeWriter.emitCode(" = %L", asExpressionBody)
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -196,7 +207,7 @@ class FunSpec private constructor(
   }
 
   private fun isExpressionBody(): Boolean {
```
JakeWharton: What's this used for? I'm tempted to make the callers just check for nullability of asExpressionBody similar to what is happening in the codeWriter interaction above.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -457,7 +468,8 @@ class FunSpec private constructor(
     internal val String.isConstructor get() = this == CONSTRUCTOR
     internal val String.isAccessor get() = this.isOneOf(GETTER, SETTER)
 
-    private val EXPRESSION_BODY_PREFIX = CodeBlock.of("return ")
+    private val RETURN_EXPRESSION_BODY_PREFIX = CodeBlock.of("return ")
+    private val THROW_EXPRESSION_BODY_PREFIX = CodeBlock.of("throw")
```
JakeWharton: Need trailing space. Please add a test case for a function body that calls a method whose name starts with `throw` like `throwOrPrint("hi")` to ensure we never regress this case.

Egorand: Adding a space to the end of the prefix is important to avoid cases like this:
```kotlin
    .returns(Int::class)
    .addStatement("throwsAFoo()")
    .addStatement("return 1 + 1")
    .build()
```
If the space is omitted, the output will be incorrect:
```kotlin
expected:
    fun foo(): kotlin.Int {
      throwsAFoo()
      return 1 + 1
    }
    
but was:
    fun foo(): kotlin.Int = throwsAFoo()
    return 1 + 1
```
Space here basically ensures that `throw` is matched only when it's used as a keyword.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -110,6 +110,17 @@ class FunSpec private constructor(
     }
   }
 
+  private fun asExpressionBody(codeBlock: CodeBlock): CodeBlock? {
```
Egorand: nit: maybe make it an extension function?

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -274,6 +274,16 @@ class FunSpecTest {
       |""".trimMargin())
   }
 
+  @Test fun functionWithThrows() {
+    val funSpec = FunSpec.builder("foo")
+        .addStatement("throw %T()", AssertionError::class)
+        .returns(ClassName.bestGuess("kotlin.Nothing"))
```
Egorand: nit: use the `NOTHING` constant

src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -692,7 +692,7 @@ class TypeSpecTest {
             .addParameter("label", t)
             .addParameter("x", p)
             .addParameter("y", p)
-            .addStatement("throw new %T(%S)", UnsupportedOperationException::class, "TODO")
```
Egorand: Ha, it's 2019 and we're still discovering Java relics in this codebase!

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -110,6 +110,17 @@ class FunSpec private constructor(
...
+    val asReturnExpressionBody = codeBlock.trim().withoutPrefix(RETURN_EXPRESSION_BODY_PREFIX)
+    if (asReturnExpressionBody != null) {
+      return asReturnExpressionBody
+    }
+    if (codeBlock.trim().withoutPrefix(THROW_EXPRESSION_BODY_PREFIX) != null) {
```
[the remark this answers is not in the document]
ShaishavGandhi: Good call 
2f3bcf3c6469a3b1e4ed5f232b0f003db7c308a2

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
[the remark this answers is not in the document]
ShaishavGandhi: 2f3bcf3c6469a3b1e4ed5f232b0f003db7c308a2

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
ShaishavGandhi: Have removed this method. It was used to check emitting of return types.

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
ShaishavGandhi: Good point, missed that. 
02ae906ac6a7deded0da407a271ee8a5cedd1205

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
ShaishavGandhi: 2f3bcf3c6469a3b1e4ed5f232b0f003db7c308a2

src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
ShaishavGandhi: 2ae906ac6a7deded0da407a271ee8a5cedd1205

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
ShaishavGandhi: Yup, good point! 
02ae906ac6a7deded0da407a271ee8a5cedd1205

src/main/java/com/squareup/kotlinpoet/FunSpec.kt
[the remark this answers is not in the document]
ShaishavGandhi: 2f3bcf3c6469a3b1e4ed5f232b0f003db7c308a2

Egorand:
Thanks!

PR #711, opened 2019-06-25
Start kotlinpoet-km artifact

ZacSweers:
Part of #707. No tests right now yet pending feedback on the implementation and also thoughts on what tests for this should look like!

Issue #707, opened 2019-06-21
Kotlinx-metadata integration


ZacSweers:
Working doc and notes from offline conversation with @Egorand. The goal is to incorporate Kotlinx-metadata and to:

* Create ABI representations of existing code with KotlinPoet types from metadata
* Optionally offer the underlying immutable types layer (explained below)

### Incorporate immutable km types work as a separate artifact. 

Offered as a standalone public, immutable API over kotlinx-metadata's without exposing it directly. Will be used for KotlinPoet but also available for others' use as a read-only API.

* Keeping a separate artifact as there is information that cannot easily be expressed in KotlinPoet today. Some can be added (see KotlinPoet segment below), but some things are based on source and not necessarily relevant to kotlinpoet
  * jvm signatures
  * Some raw data that we wouldn't express the same way in kotlinpoet
  * Synthetic property annotation holder methods
* Artifact name tbd
* Current impl: https://github.com/ZacSweers/copydynamic/blob/master/metric/src/main/kotlin/io/sweers/metric/immutableNodes.kt
* Will shade kotlinx-metadata dependency, not exposed in public API
  * Mostly there now, just need to make flags alternative (or remove need for them)

### Add support for reading ABIs from class/kclass/metadata instances in KotlinPoet

Depending on immutable km artifact to do so, and being able to create KotlinPoet *specs representations from km types.

* TypeSpecs (supports all three)
  * `val okHttpClientType: TypeSpec = TypeSpec.fromKotlin(OkHttpClient::class)`
* FileSpec (only with class/kclass, or metadata with a package name)
* Underlying immutable km types available via `tags()` API, guaranteed to be present
  * `val kmProperty = propertySpec.tag<ImmutableKmProperty>()`
* Metadata coerced to source representation
  * For instance - in metadata, `String.() -> Unit` looks like `Function<String, Unit>`. We can coerce this to be a `LambdaTypeName` with receiver from the available metadata though, and express it idiomatically for source
    * Same for suspend functions/types and other lambdas.
* Creates compile-able ABIs. Function bodies, default value expressions, etc all stubbed with `TODO("Stub")` as needed.
* Current implementations:
  * https://github.com/ZacSweers/copydynamic/blob/master/metric/src/main/kotlin/io/sweers/metric/Metric.kt
  * https://github.com/ZacSweers/copydynamic/blob/master/metric/src/main/kotlin/io/sweers/metric/MetricKotlinPoet.kt

[...]

PR #713, opened 2019-06-27
Implement support for contracts on functions

ZacSweers:
Part of building out parity for metadata (#707). This implements support for contracts on functions. Contracts are considered part of the signature for functions in the kotlin compiler and in kotlin metadata, and this leans on much of metadata's semantics to build that into an API for functions. Everything is still builder based, but I _think_ it should be possible to basically do everything you could need to do by just using the static helper functions.

Notes:
- Right now this avoids touching contracts code directly to avoid the experimental annotation requirement. Should it use it since they're experimental? It's hard to contain the virality of the experimental annotations usage once it reaches `FunSpec` though. I'm personally of the opinion to just keep it without any experimental annotations or direct references and instead trust the user to know what they're doing.
- Started tests, want to do a review pass before proceeding with finishing them.
- `ContractEffectExpressionSpec` looks gnarly, but it's basically the same API as how it looks in metadata and has to be meticulous about how it describes things due to the limited nature of contracts. The named static factories offer better semantics
- Unclear how and/or ordering can be interpreted from metadata. May need to tweak once we integrate this with #711. I wasn't able to find any examples of this in the kotlin stdlib

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ContractSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ContractSpec.kt
@@ -1,0 +1,606 @@
+package com.squareup.kotlinpoet
+
+import com.squareup.kotlinpoet.ContractEffectType.CALLS
+import com.squareup.kotlinpoet.ContractEffectType.RETURNS_CONSTANT
+import com.squareup.kotlinpoet.ContractEffectType.RETURNS_NOT_NULL
+import kotlin.reflect.KClass
+
+/** Represents a contract of a Kotlin function. */
+class ContractSpec private constructor(
+  builder: Builder,
+  private val tagMap: TagMap = builder.buildTagMap()
+) : Taggable by tagMap {
+
+  /** Effects of this contract. */
+  val effects: List<ContractEffectSpec> = builder.effects.toImmutableList()
+
+  init {
+    require(effects.isNotEmpty()) {
+      "Contract must have at least one effect!"
+    }
+  }
+
+  internal fun emit(codeWriter: CodeWriter, function: FunSpec) {
+    codeWriter.emitCode("%M·{\n", MemberName("kotlin.contracts", "contract"))
+    codeWriter.indent()
+    effects.joinEmissions(codeWriter, separator = "\n", postfix = "\n") { writer, effect ->
+      effect.emit(writer, function)
+    }
+    codeWriter.unindent()
+    codeWriter.emitCode("}\n")
+  }
+
+  private fun <T> List<T>.joinEmissions(
+    codeWriter: CodeWriter,
+    separator: String = ", ",
+    prefix: String = "",
+    postfix: String = "",
+    emitter: (CodeWriter, T) -> Unit
+  ) {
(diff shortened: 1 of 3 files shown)


JakeWharton:
This is a lot of machinery.

ZacSweers:
yeah, I don't think we can escape needing the machinery but we could probably get by with making most of it `internal`. Main problem with that is it makes it impossible for external users to read what a contract is.

In practice, I think it reads just fine

```kotlin
val contractSpec = ContractSpec.builder()
    .addEffect(ContractEffectSpec.calls(param, EXACTLY_ONCE))
    .addEffect(
        ContractEffectSpec.returnsNotNull(
            conclusion = ContractEffectExpressionSpec.nullCheck(
                parameterIndex = 1,
                isNegated = true
            )
        )
    )
    .build()
```

Naming helps a lot, suggestions welcome :)

JakeWharton:
So why is this different from other constructs which appear inside the body of a function? Even though it's part of the signature, it really isn't on the same level syntactically as the rest of the signature.

JakeWharton:
Different, meaning, why should we choose to model it when we don't model anything else inside bodies.

ZacSweers:
Contracts aren't like everything else inside bodies though.

- only one per function is allowed
- Some have specific rules that we can better enforce with a model:
  - `callsInPlace` requires the function to be inline
  - `callsInPlace` requires the target parameter be an invokable function
- It's contained in metadata about that function. If we use KotlinPoet's types as an IR without modeling this, representing metadata as a function ABI becomes lossy unless someone wants to manually scrape this out of the function body. At the same time - we still need to build this anyway to express it in the body anyway

Contracts have transitive implications on parameters instance at call sites and are as much a part of the signature from an API perspective as named parameters or the `internal` modifier are. It's not in bytecode, but it's part of Kotlin's model of that function that callers can leverage. That's a compelling reason to me to model this properly.

JakeWharton:
The first two bullet points don't really resonate with me. We're not building a complete compiler, `kotlinc` can validate these things. We don't validate that calls to `T::class` are only on `reified` type parameters either and I wouldn't want to. A model can be useful for modeling _anything_ inside a body, it's about what we choose to model.

For point 3, I'm not sure being an IR is a goal of the project. And you don't need to build this in the same way we don't need to build abstractions for `if`/`else`, `when`, or lambdas in general.

ZacSweers:
Sure, but at the end of the day a contract is a property of the function API that just happens to be defined in the body rather than the formal signature. I would put it on par with receivers, modifiers, etc.

> I'm not sure being an IR is a goal of the project

It is a goal :). At least from conversations with @Egorand and @swankjesse

JakeWharton:
I'm still opposed to it. Not to mention it might incompatibly change at any point prior to stability.

ZacSweers:
are you opposed to it because of the machinery nature of the implementation or because of the principle?

ZacSweers:
I'll can this for now to focus on #711. Would still be interested in hearing more detailed reasons for being against it, but in the meantime we can avoid making km conversions lossy by just storing the `KmContract` in a given FunSpec's tags()

PR #716, opened 2019-07-02
Add experimental typeNameOf() API for TypeName

ZacSweers:
This depends on the experimental `typeOf()` API from the stdlib and propagates the same experimental annotation to guard it.

This can be used as a shorthand for basically almost all of the current typename variants' static factories, and just leave the factories for when arguments are being passed in from somewhere else.


ZacSweers:
Yep will do, wanted to get a read of thoughts on adding the API first

ZacSweers:
The current TypeNameTest class is java. Do we want to move it to kotlin or start a new one just for this API? Let's do it in a followup pending that answer :)

Egorand:
Sounds good! Let's leave that one as is and start a `TypeNameKotlinTest`.

PR #719, opened 2019-07-03
Add a simple constructor for ParameterSpec

### LINKED ISSUE #717: Provide a simpler path for ParameterSpec creation
Approximately 90% of the time I am creating `ParameterSpec`s, it takes the form `ParameterSpec.builder(name, type).build()`.  Modifiers on `ParameterSpec` are rare -- `noinline`, `crossinline`, and `vararg` are the only possibilities I can think of, and it makes sense that they're not the most common.

It makes lots of sense to provide a simpler constructor, pseudoconstructor, or factory method for this common case, instead of forcing users to go through the builder.

lowasser:
Fixes #717.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -26,6 +26,9 @@ class ParameterSpec private constructor(
   builder: ParameterSpec.Builder,
   private val tagMap: TagMap = builder.buildTagMap()
 ) : Taggable by tagMap {
+  
+  constructor(name: String, type: TypeName, vararg modifiers: KModifier): this(builder(name, type, *modifiers))
+  
   val name = builder.name
   val kdoc = builder.kdoc.build()
   val annotations = builder.annotations.toImmutableList()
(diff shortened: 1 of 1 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -26,6 +26,9 @@ class ParameterSpec private constructor(
   builder: ParameterSpec.Builder,
   private val tagMap: TagMap = builder.buildTagMap()
 ) : Taggable by tagMap {
+  
+  constructor(name: String, type: TypeName, vararg modifiers: KModifier): this(builder(name, type, *modifiers))
```
Egorand: Do we want overloads that accept `java.lang.reflect.Type` and `KClass`, similar to `builder()` overloads?

Egorand:
@lowasser I'm really unsure what Spotless wants, the formatting looks correct and I can't repro the issue locally when I copy your change into master. If you can run `./gradlew spotlessApply` and push that'd be awesome, otherwise I'll try cloning your fork to figure out what's wrong.

lowasser:
@Egorand I did that, and it looks like it worked.  You should be able to merge now!

Egorand:
Perfect, thanks @lowasser!

PR #723, opened 2019-07-08
typeNameOf tests + move

ZacSweers:
This moves `typeNameOf()` to `TypeName.kt` (something I had committed locally but forgot to push before the last PR was merged), and adds tests for it in `TypeNameKotlinTest.kt`

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
@@ -21,10 +21,6 @@ import java.lang.reflect.ParameterizedType
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
 import kotlin.reflect.KTypeParameter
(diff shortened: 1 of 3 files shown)


ZacSweers:
> Otherwise we need to bound the T to Any if it's unsupported.

I'm not sure I followed this. Have an example?

JakeWharton:
`typeNameOf<String?> == "kotlin.String?"`

ZacSweers:
Ah, in that case we're good. Added a test and it worked fine 👍

JakeWharton:
Great! That would be a huge limitation if it wasn't supported.

Egorand:
I didn't realize how awesome this new API is until I saw the tests. Is this something we can use internally as well to solve the problem of Kotlin types being reported as Java types when read from bytecode?

ZacSweers:
Unfortunately not, this API is only useful for source where you directly define the type (i.e. `typeNameOf<String>()`). For reading from bytecode like elements and class, they'll still show up as the java forms. the metadata artifact should hopefully help with those cases though.

PR #725, opened 2019-07-09
Start metadata -> TypeSpec APIs

ZacSweers:
This is the final large piece of bringing over the Metric work, with APIs to convert the immutable km types to TypeSpec/FileSpec representations. Very much a rough draft right now and not feature complete, but I think it's a good starting point for an initial pass before I continue fleshing out the tests and implementation.

Some key details:
- Main entry point is APIs like `Class<*>.toTypeSpec()`
- Tries to represent source ABIs only. No synthetic, delegated, or otherwise undeclared elements are rendered even if they're in byte code since we want this to look like the idiomatic kotlin that it represents
- Tries to understand kotlin lambdas (i.e. `kotlin.Function*`) and suspend types to represent them idiomatically via `LambdaTypeName`.

One thing this maybe asks for is a tag API on `TypeName` directly. I'll open a separate proposal PR for that to see how it looks


kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
```diff
@@ -0,0 +1,405 @@
...
+fun KClass<*>.toTypeSpec(): TypeSpec = java.toTypeSpec()
+/** @return a [TypeSpec] ABI representation of this [KClass]. */
+@KotlinPoetKm
+fun Class<*>.toTypeSpec(): TypeSpec = toImmutableKmClass().toTypeSpec()
+/** @return a [TypeSpec] ABI representation of this [TypeElement]. */
+@KotlinPoetKm
+fun TypeElement.toTypeSpec(): TypeSpec = toImmutableKmClass().toTypeSpec()
+/** @return a [FileSpec] ABI representation of this [KClass]. */
+@KotlinPoetKm
+fun KClass<*>.toFileSpec(): FileSpec = java.toFileSpec()
```
Egorand: Do we care about cases when there are multiple types declared in a file?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
```diff
@@ -0,0 +1,405 @@
...
+private const val TODO_BLOCK = "TODO(\"Stub!\")"
+
+@KotlinPoetKm
+private fun ImmutableKmClass.toTypeSpec(): TypeSpec {
+  // Fill the parametersMap. Need to do sequentially and allow for referencing previously defined params
+  val parametersMap = mutableMapOf<Int, TypeName>()
+  val typeParamResolver = { id: Int -> parametersMap.getValue(id) }
+  typeParameters.forEach { parametersMap[it.id] = it.toTypeVariableName(typeParamResolver) }
+
+  val simpleName = name.substringAfterLast(if (isInline) "/" else ".")
```
Egorand: That's interesting, got an example of an inline class' FQCN?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
```diff
@@ -0,0 +1,405 @@
...
+      if (isConstructorParam || (!isDelegated && !isLateinit)) {
+        // TODO if hasConstant + elements, we could look up the constant initializer
+        when {
+          isConstructorParam -> initializer(name)
+          type.isNullable -> initializer("null")
+          else -> initializer(TODO_BLOCK)
+        }
+      }
+      if (hasGetter) {
+        val visibility = setterFlags.visibility
```
Egorand: Should this block be using `getterFlags`? Also, use it for both getter and setter to avoid duplication?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
```diff
@@ -0,0 +1,405 @@
...
+  return map {
+    when (it) {
+      IS_EXTERNAL -> KModifier.EXTERNAL
+      IS_INLINE -> KModifier.INLINE
+      IS_NOT_DEFAULT -> TODO("Wat")
+    }
+  }.toTypedArray()
+}
+
+@KotlinPoetKm
```
Egorand: Btw, why do we care about marking private methods with experimental annotations?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -0,0 +1,326 @@
...
+  interface BaseInterface
+  class Supertype : BaseType(), BaseInterface
+
+  @Test
+  fun properties() {
+    val typeSpec = Properties::class.toTypeSpec()
+
+    assertThat(typeSpec.propertySpecs).hasSize(4)
+
+    val fooProp = typeSpec.propertySpecs.find { it.name == "foo" } ?: throw AssertionError(
```
Egorand: nit: use `fail()` instead of `throw AssertionError`?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -0,0 +1,326 @@
...
+    checkNotNull(companionObject)
+    assertThat(companionObject.name).isEqualTo("Companion")
+  }
+
+  class CompanionObject {
+    companion object
+  }
+
+  @Test
+  fun namedCompanionObject() {
```
Egorand: Nice!

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -0,0 +1,326 @@
...
+
+  @Test
+  fun generics() {
+    val typeSpec = Generics::class.toTypeSpec()
+
+    assertThat(typeSpec.typeVariables).hasSize(3)
+    val tType = typeSpec.typeVariables[0]
+    assertThat(tType.name).isEqualTo("T")
+    assertThat(tType.isReified).isFalse()
+    assertThat(tType.variance).isNull() // we don't redeclare out variance
```
Egorand: Not clear what "redeclare" here means

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -0,0 +1,326 @@
...
+  @Test
+  fun collectionMutability() {
+    val typeSpec = CollectionMutability::class.toTypeSpec()
+
+    assertThat(typeSpec.primaryConstructor?.parameters).hasSize(2)
+
+    val (immutableProp, mutableListProp) = typeSpec.primaryConstructor!!.parameters
+    assertThat(immutableProp.type).isEqualTo(List::class.parameterizedBy(String::class))
+    assertThat(mutableListProp.type).isEqualTo(
+        ClassName.bestGuess("kotlin.collections.MutableList").parameterizedBy(
```
Egorand: nit: use `MUTABLE_LIST`

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
ZacSweers: for right now - no. Would definitely be something to iterate on

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
ZacSweers: yep! Good catch, will fix

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
ZacSweers: I think it's just the nature of how these experimental annotations work, you have to apply them everywhere. Or did you mean why use this vs `@UseExperimental`?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
ZacSweers: specifying `out` would be redundant

ZacSweers:
I think I need to reopen this, as my fork's master somehow got chomped, and I also don't know how to rebase on top of a master merge from a maintainer into a PR. Unless you want to merge this as-is and do comment nits in a followup

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
Egorand: No, using our own is the right way to go. I would expect them only be necessary for API, not for the private functions, but if that's how it works I'm fine. No big issue.

Egorand:
Fine with merging as is and fixing stuff in followups.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
Egorand: Why so? Not sure I get it.

ZacSweers: if you write `List<out T>`, the compiler will nag you about it being unnecessary. This checks that we don't emit it that way

ZacSweers:
Cool, let's do that then I'll re-fork and update

PR #729, opened 2019-07-25
Flesh out more KmSpecs tests

ZacSweers:
This reveals a few (now-documented) WIP areas and also addresses the open comments from the previous PR

The missing pieces unfortunately would require inspecting the corresponding classes/subtypes (detecting overrides, companion objects, enum member types)

### DIFF
--- a/kotlinpoet-km/src/main/kotlin/com/squareup/kotlinpoet/km/Flags.kt
+++ b/kotlinpoet-km/src/main/kotlin/com/squareup/kotlinpoet/km/Flags.kt
@@ -257,7 +257,9 @@ val ImmutableKmValueParameter.isNoInline: Boolean get() = Flag.ValueParameter.IS
 
 // Property flags.
 @KotlinPoetKm
-val Flags.isOverrideProperty: Boolean get() = Flag.Property.IS_FAKE_OVERRIDE(this)
+val Flags.isFakeOverrideProperty: Boolean get() = Flag.Property.IS_FAKE_OVERRIDE(this)
+@KotlinPoetKm
+val Flags.isOverrideProperty: Boolean get() = false // TODO This isn't contained in metadata. Likely have to read an annotation
 @KotlinPoetKm
 val KmProperty.hasConstant: Boolean get() = Flag.Property.HAS_CONSTANT(flags)
 @KotlinPoetKm
@@ -277,7 +279,7 @@ val KmProperty.isExpect: Boolean get() = Flag.Property.IS_EXPECT(flags)
 @KotlinPoetKm
 val KmProperty.isExternal: Boolean get() = Flag.Property.IS_EXTERNAL(flags)
 @KotlinPoetKm
-val KmProperty.isOverride: Boolean get() = flags.isOverrideProperty
+val KmProperty.isFakeOverride: Boolean get() = flags.isFakeOverrideProperty
 @KotlinPoetKm
 val KmProperty.isLateinit: Boolean get() = Flag.Property.IS_LATEINIT(flags)
 @KotlinPoetKm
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -27,17 +34,15 @@ class KmSpecsTest {
...
-    assertThat(barParam.name).isEqualTo("bar")
-    assertThat(barParam.modifiers).contains(KModifier.VARARG)
-    assertThat(barParam.type).isEqualTo(INT)
+    //language=kotlin
+    assertThat(typeSpec.trimmedToString()).isEqualTo("""
+      class ConstructorClass(
+        val foo: kotlin.String,
+        vararg bar: kotlin.Int
+      ) {
+        constructor(bar: kotlin.Int)
```
Egorand: Oh, so the secondary constructor isn't ignored as stated in the comment inside `ConstructorClass`?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
```diff
@@ -62,24 +69,18 @@ class KmSpecsTest {
...
-    val listProp = typeSpec.propertySpecs.find { it.name == "aList" } ?: throw AssertionError(
-        "Missing baz property")
-    assertThat(listProp.type).isEqualTo(List::class.parameterizedBy(Int::class))
-    assertThat(listProp.mutable).isTrue()
+    //language=kotlin
+    assertThat(typeSpec.trimmedToString()).isEqualTo("""
+      class Properties {
+        var aList: kotlin.collections.List<kotlin.Int> = TODO("Stub!")
+
+        val bar: kotlin.String? = null
```
Egorand: Should nullable fields also get assigned a `TODO` instead of null?

kotlinpoet-km/src/main/kotlin/com/squareup/kotlinpoet/km/Flags.kt
```diff
@@ -257,7 +257,9 @@ val ImmutableKmValueParameter.isNoInline: Boolean get() = Flag.ValueParameter.IS
 
 // Property flags.
 @KotlinPoetKm
-val Flags.isOverrideProperty: Boolean get() = Flag.Property.IS_FAKE_OVERRIDE(this)
+val Flags.isFakeOverrideProperty: Boolean get() = Flag.Property.IS_FAKE_OVERRIDE(this)
```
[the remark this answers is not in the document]
ZacSweers: From the doc: 
```
/**
 * A member kind flag, signifying that the corresponding function exists in the containing class because a function with a suitable
 * signature exists in a supertype. This flag is not written by the Kotlin compiler and its effects are unspecified.
 */
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
```diff
@@ -289,7 +299,17 @@ private fun ImmutableKmProperty.toPropertySpec(
         mutable(false)
       }
       if (isDelegated) {
-        delegate("") // Placeholder
+        // Placeholders for these are tricky
+        addKdoc("Note: delegation is ABI stub only and not guaranteed to match source code.")
```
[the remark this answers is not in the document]
ZacSweers: I think it's useful just since this is a case where we're clearly fabricating behavior for the sake of compilation and not actually matching source behavior. Similar to the `TODO("Stub!")` holders, but the only way we can express that here is a comment :/

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
ZacSweers: I think semantics are maybe mixed up. The comment is in reference to the `: this(...)` that the secondary constructor should call. Not sure how to read that from metadata

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KmSpecsTest.kt
ZacSweers: I don't think so, unless you think that'd make it more clear that it's a placeholder

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KmSpecs.kt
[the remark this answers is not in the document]
Egorand: Okay, let's leave it

PR #731, opened 2019-07-27
add ClassName constructors

### LINKED ISSUE #730: please give me a constructor to make ClassName with List<String>
Now I can new `ClassName` with two `String` and vararg `String`s. but I would like constructor with `List<String>`.

Of course, I know that way to make  `ClassName` with `List<String>`. but it is not good for performance:
```
val names = listOf("kotlin.collections", "Map", "Entry")
val className = ClassName(names[0], names[1], *names.drop(2).toTypedArray())
```

anatawa12:
Fixes #730

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -288,8 +288,27 @@ class ClassName internal constructor(
    * Returns a class name created from the given parts. For example, calling this with package name
    * `"java.util"` and simple names `"Map"`, `"Entry"` yields `Map.Entry`.
    */
+  @Deprecated("", level = DeprecationLevel.HIDDEN)
   constructor(packageName: String, simpleName: String, vararg simpleNames: String) :
       this(listOf(packageName, simpleName, *simpleNames))
+
+  /**
+   * Returns a class name created from the given parts. For example, calling this with package name
+   * `"java.util"` and simple names `"Map"`, `"Entry"` yields `Map.Entry`.
+   */
+  constructor(packageName: String, vararg simpleNames: String) :
+      this(listOf(packageName, *simpleNames)) {
+    require(simpleNames.isNotEmpty()) { "simpleNames must not be empty" }
+  }
+
+  /**
+   * Returns a class name created from the given parts. For example, calling this with package name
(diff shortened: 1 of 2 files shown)


anatawa12:
@JakeWharton you seem to be waiting for that I marge this. But I can't marge this because I don't have write access. 

If you aren't waiting, could you tell me why you don't marge?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -288,9 +288,28 @@ class ClassName internal constructor(
    * Returns a class name created from the given parts. For example, calling this with package name
    * `"java.util"` and simple names `"Map"`, `"Entry"` yields `Map.Entry`.
    */
+  @Deprecated("", level = DeprecationLevel.HIDDEN)
```
Egorand: This is a binary incompatible change that will affect Java callers:
```java
String[] names = { "simple2", "simple3" };
ClassName cn3 = new ClassName("package", "simple", names);
```
This code compiles before the change and won't compile after. Normally we would require a major version change to merge binary-incompatible code, and given that this change is an API enhancement, not a bug fix, I would rather postpone merging this until we're ready for 2.0. @JakeWharton what are your thoughts?

ZacSweers:
Why not give it a different JvmName for Java callers?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
JakeWharton: It's source incompatible for that specific case, not binary. And I highly doubt anyone is using it that way. It's far more common to do `get("package", "Map", "Entry")` which remains source compatible.

I don't have any problem breaking source compatibility for a usability improvement.

Egorand:
@ZacSweers It's a constructor.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
Egorand: Yeah, that's fair.

anatawa12:
I signed CLA.

anatawa12:
Thank you for accepting my future request and merging PR.

Egorand:
@anatawa12 thanks for your PR!

PR #739, opened 2019-07-30
add overrie with list for each function with vararg parameters

### LINKED ISSUE #737: add override with List for each function with vararg parameters
I often write source codes like this for many function invocations. I think it's not good for performance and it's not good readable code.
```
type.parameterizedBy(*typeArguments.toTypedArray())
```
if there is override with List, I can write like this.
```
type.parameterizedBy(typeArguments)
```

anatawa12:
Resolves #737
Some override using List to less allocation.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -378,26 +378,52 @@ class FunSpec private constructor(
       parameters += parameterSpec
     }
 
+    @JvmName("callThisConstructorWithStrings")
+    fun callThisConstructor(args: Iterable<String>) = apply {
+      callConstructor("this", args.map { CodeBlock.of(it) })
+    }
+
+    fun callThisConstructor(args: List<CodeBlock>) = apply {
+      callConstructor("this", args)
+    }
+
+    fun callThisConstructor(args: Iterable<CodeBlock>) = apply {
+      callConstructor("this", args.toList())
+    }
+
     fun callThisConstructor(vararg args: String) = apply {
-      callConstructor("this", *args.map { CodeBlock.of(it) }.toTypedArray())
+      callConstructor("this", args.map { CodeBlock.of(it) })
     }
 
     fun callThisConstructor(vararg args: CodeBlock = emptyArray()) = apply {
-      callConstructor("this", *args)
+      callConstructor("this", args.toList())
+    }
+
+    @JvmName("callSuperConstructorWithStrings")
+    fun callSuperConstructor(args: Iterable<String>) = apply {
+      callConstructor("super", args.map { CodeBlock.of(it) })
+    }
+
+    fun callSuperConstructor(args: Iterable<CodeBlock>) = apply {
+      callConstructor("super", args.toList())
+    }
+
+    fun callSuperConstructor(args: List<CodeBlock>) = apply {
+      callConstructor("super", args)
     }
 
     fun callSuperConstructor(vararg args: String) = apply {
-      callConstructor("super", *args.map { CodeBlock.of(it) }.toTypedArray())
+      callConstructor("super", args.map { CodeBlock.of(it) })
     }
 
     fun callSuperConstructor(vararg args: CodeBlock = emptyArray()) = apply {
-      callConstructor("super", *args)
+      callConstructor("super", args.toList())
     }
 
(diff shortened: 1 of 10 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -378,26 +378,52 @@ class FunSpec private constructor(
       parameters += parameterSpec
     }
 
+    @JvmName("callThisConstructorWithStrings")
```
Egorand: I really don't like the name and I can't see a better alternative. I think we should just drop the overloads that require changing method names.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -378,26 +378,52 @@ class FunSpec private constructor(
...
+    fun callThisConstructor(args: Iterable<CodeBlock>) = apply {
+      callConstructor("this", args.toList())
+    }
+
     fun callThisConstructor(vararg args: String) = apply {
-      callConstructor("this", *args.map { CodeBlock.of(it) }.toTypedArray())
+      callConstructor("this", args.map { CodeBlock.of(it) })
     }
 
     fun callThisConstructor(vararg args: CodeBlock = emptyArray()) = apply {
```
Egorand: It's interesting that we use an `emptyArray()` default arg here but not in the `vararg args: String` overload. This is also missing a `@JvmOverloads`. Out of scope of this PR, so no need to change.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -378,26 +378,52 @@ class FunSpec private constructor(
...
-      callConstructor("this", *args)
+      callConstructor("this", args.toList())
+    }
+
+    @JvmName("callSuperConstructorWithStrings")
```
Egorand: Same here, let's remove this overload.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/jvm/JvmAnnotations.kt
```diff
@@ -69,12 +69,24 @@ fun FunSpec.Builder.jvmName(name: String) = apply {
...
 
 fun FunSpec.Builder.throws(vararg exceptionClasses: Type) =
-    throws(*(exceptionClasses.map(Type::asTypeName).toTypedArray()))
+    throws(exceptionClasses.toList())
 
-fun FunSpec.Builder.throws(vararg exceptionClasses: TypeName) = addAnnotation(
+fun FunSpec.Builder.throws(vararg exceptionClasses: TypeName) =
+    throws(exceptionClasses.toList())
+
+@JvmName("throwsKClasses")
```
Egorand: Let's remove all these cause they require name changes.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
anatawa12: should I keep `private fun callConstructor(constructor: String, vararg args: CodeBlock)`? I think it makes duplicated code with `private fun callConstructor(constructor: String, vararg args: CodeBlock)` and `private fun callConstructor(constructor: String, args: List<CodeBlock>)`

kotlinpoet/src/main/java/com/squareup/kotlinpoet/jvm/JvmAnnotations.kt
anatawa12: Fixed in 4f12e7d. could you check?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
anatawa12: Fixed in 4f12e7d. could you check?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
Egorand: No, I'm fine with the change you made. The arguments are ultimately kept as a list, so it doesn't matter at which point we convert a `vararg` to a list.

anatawa12:
Thank you for merging my PR!

PR #740, opened 2019-08-03
Switch to openjdk8 for CI

Egorand:


JakeWharton:
Change deploy script too

Egorand:
Ah good catch, changed.

PR #741, opened 2019-08-03
Large chunk of misc km work

ZacSweers:
- Functions now correctly report type arguments
- Safer handling of enums with class bodies
- Transient/volatile/synchronized support
- New `ElementHandler` API for retrieving information about related elements
  - Full companion object support
  - Full enum entry class body support
  - Full annotations support
  - Full constant value literals support
  - Full nested classes support
  - Some override support
  - Heavily caches lookups
- Fixed interface generation
  - Don't generate bodies unless it's JvmDefault or has body in source
    - This involved adding a `CodeBlock.Builder#clear()` function, which I think is a useful addition.
  - Don't extend from `Any`
  - Resolve which supertypes are actually interfaces via `ElementHandler`

Talked with @Egorand offline and we'll plan to do an in-person CR of this since it's unfortunately all a bit stuck together and interwoven

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ElementHandler.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ElementHandler.kt
@@ -27,6 +27,7 @@ import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
 import java.util.concurrent.ConcurrentHashMap
+import javax.lang.model.element.Element
 import javax.lang.model.element.ElementKind.ENUM_CONSTANT
 import javax.lang.model.element.ElementKind.INTERFACE
 import javax.lang.model.element.ExecutableElement
@@ -383,11 +384,14 @@ interface ElementHandler {
           }
         }
 
-        private fun lookupMethod(classJvmName: String, methodSignature: JvmMethodSignature): ExecutableElement? {
+        private fun lookupMethod(classJvmName: String,
+            methodSignature: JvmMethodSignature,
+            elementFilter: (Iterable<Element>) -> List<ExecutableElement>
+        ): ExecutableElement? {
           return lookupTypeElement(classJvmName)?.let {
             val signatureString = methodSignature.asString()
             methodCache.getOrPut(it to signatureString) {
-              ElementFilter.methodsIn(it.enclosedElements)
+              elementFilter(it.enclosedElements)
                   .find { signatureString == it.jvmMethodSignature(types) }.toOptional()
             }.nullableValue
           }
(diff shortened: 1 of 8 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -428,6 +428,11 @@ class CodeBlock private constructor(
       formatParts += "⇤"
     }
 
+    fun clear() = apply {
```
JakeWharton: This could land separately. Also needs a test.

ZacSweers: Good point @742

ZacSweers:
@Egorand pushed the updated bits from what we talked about in the in-person review today. Was able to resolve every outstanding issue but the super constructor calls. 

Since we're going to split it out to a separate artifact, I've added autocommon API assistance in handling override detection. We can shade this in the new artifact, or I can copy all the relevant APIs from AutoCommon over if we really want to avoid the dependency entirely.

Otherwise, this is in a reaaaaaally good place now 👍. We could even look at splitting out the elementhandler implementations to different artifacts if we really want to have maximum separation of concerns

ZacSweers:
spoke too soon -_-. Will figure out what I accidentally broke

ZacSweers:
Fixed 🤞

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ElementHandler.kt
```diff
@@ -388,7 +395,22 @@ interface ElementHandler {
         methodSignature: JvmMethodSignature
       ): Boolean {
         val clazz = lookupClass(classJvmName) ?: error("No class found for: $classJvmName.")
-        return clazz.lookupMethod(methodSignature)?.declaringClass == clazz
+        val signatureString = methodSignature.asString()
+        val classPackage = clazz.`package`.name
+        val interfaceMethods = clazz.interfaces.asSequence()
+            .flatMap { it.methods.asSequence() }
+        val superClassMethods = clazz.superclass?.methods.orEmpty().asSequence()
+        return interfaceMethods.plus(superClassMethods)
```
Egorand: Looks good!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ElementHandler.kt
```diff
@@ -0,0 +1,468 @@
...
+          return null
+        }
+        return field
+            .get(null) // Constant means we can do a static get on it.
+            .asLiteralCodeBlock()
+      }
+    }
+
+    /** @return an [Elements]-based implementation of [ElementHandler]. */
+    fun fromElements(elements: Elements, types: Types): ElementHandler {
```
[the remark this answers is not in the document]
ZacSweers: It's in an interface, so only possible if we make KotlinPoet target jvm 1.8

PR #743, opened 2019-08-08
Ensure trailing newlines in kdocs and functions

ZacSweers:
Analogous to https://github.com/square/javapoet/pull/732

There appears to be some existing logic for this for kdocs, but I couldn't really grok how it was intended to work just from reading the code alone. Suggestions welcome and happy to merge them if possible

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -332,7 +332,7 @@ internal class CodeWriter constructor(
         }
       }
     }
-    if (ensureTrailingNewline && out.lastChar != '\n') {
+    if (ensureTrailingNewline && out.hasPendingSegments) {
       emit("\n")
     }
   }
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -213,7 +213,11 @@ internal class CodeWriter constructor(
 
   fun emitCode(format: String, vararg args: Any?) = emitCode(CodeBlock.of(format, *args))
 
-  fun emitCode(codeBlock: CodeBlock, isConstantContext: Boolean = false) = apply {
+  fun emitCode(
+    codeBlock: CodeBlock,
+    isConstantContext: Boolean = false,
+    ensureTrailingNewline: Boolean = false
```
Egorand: Both call sites pass `true`, should we just make it the default? Or even remove this parameter, would there be cases where we don't want to ensure the new line?

ZacSweers: Both that specify it do, but not every call site necessarily needs it. Was trying to be least invasive as possible starting out

image

Egorand: Ah, makes sense!

PR #747, opened 2019-08-11
Followup fixes for #741 comments

ZacSweers:
Note that JvmStatic isn't covered here yet because it's an interface and requires jvm target 1.8, but when they're moved to separate artifacts as extension functions it won't really matter


Egorand:
Thanks!

I don't see a reason not to update to jdk8. @JakeWharton is this an incompatible change that would require bumping the major version?

JakeWharton:
OkHttp did it. I think you can argue either way. The blocker was IntelliJ running JDK 6 which was fixed 3 years ago. JavaPoet [did it](https://github.com/square/javapoet/blob/master/CHANGELOG.md#javapoet-1100-2018-01-27) in a minor version. And of course OkHttp and Retrofit are doing it in minor versions as well.

As far as the semver lies we tell ourselves, the API shape isn't changing incompatibly in any way so it immediately becomes a judgement call.

Egorand:
Let's do it.

PR #748, opened 2019-08-12
Split km artifacts

### LINKED ISSUE #745: Come up with a better artifact layout
Libraries that process annotations and parse Kotlin metadata represent only a subset of KotlinPoet consumers, whereas other consumers, such as SQLDelight and Wire, have other types of input. Therefore, helpers for Metadata and Elements processing should be moved out of the main artifact, which should help keep it small and generic.

- Move recently introduced helpers for working with Kotlin metadata into a new artifact that depends on both `kotlinpoet` and `kotlinpoet-km`. **Milestone**: 1.4.
- Find good names for `kotlinpoet-km` and the new artifact. **Milestone**: 1.4.
- Move helpers for processing Elements into a new artifact. **Milestone**: 2.0.

ZacSweers:
Resolves #745 

This splits artifacts down the following way:
- `kotlinpoet-km-specs`: `KmSpecs` and `ElementHandler`
- `kotlinpoet-elementhandler-elements`: Elements-based implementation of `ElementHandler`
- `kotlinpoet-elementhandler-reflective`: Reflective-based implementation of `ElementHandler`
- `km-specs-test`: `KmSpecsTest` extracted so it can depend on the 3 above

Per offline discussion with @Egorand - keeping the tiny bit of duplication in elementhandlers for now.


ZacSweers:
Up for naming/structure suggestions. I think these are roughly the four artifacts we want though

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -18,7 +18,7 @@ object versions {
   const val spotless = "3.22.0"
   const val ktlint = "0.31.0"
   const val mavenPublish = "0.8.0"
-  const val shadowPlugin = "5.0.0"
+  const val shadowPlugin = "5.1.0"
```
Egorand: Is this an unrelated change?

km-specs-tests/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+  testImplementation(project(":kotlinpoet-km"))
+  testImplementation(project(":kotlinpoet-km-specs"))
+  testImplementation(project(":kotlinpoet-elementhandler-elements"))
+  testImplementation(project(":kotlinpoet-elementhandler-reflective"))
+  testImplementation(deps.kotlin.junit)
+  testImplementation(deps.test.truth)
+  testImplementation(deps.test.compileTesting)
+}
+
+repositories {
```
Egorand: Root `build.gradle` applies a `repositories` block to all subproject, so I believe re-applying it here is not necessary.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -475,6 +475,10 @@ class FunSpec private constructor(
       body.addStatement(format, *args)
     }
 
+    fun clearBody() = apply {
```
Egorand: Separate PR?

buildSrc/src/main/java/Dependencies.kt
Egorand: Ah, found the shading config which probably requires a version bump.

km-specs-tests/build.gradle.kts
ZacSweers: c2c64bd

ZacSweers:
> Looks good + a couple small things. Did shading work as expected?

Yup! Manually verified looking at the jar, also it's tested by virtue of the separate test project consuming it

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
ZacSweers: #752

Egorand:
Spotless is not happy, ready to merge once it's fixed.

ZacSweers:
fixed and rebased 👍

PR #751, opened 2019-08-17
Allow customization of line wrapping limit

vikasmurthy:
Simple change to expose columnLimit to allow customization of line wrapping


Egorand:
Agree. Thanks for your PR though, @vikasmurthy!

PR #754, opened 2019-08-19
Support `@JvmName` annotations in metadata parsing

### LINKED ISSUE #753: Support JvmName annotations in reflective metadata support
Right now - `@JvmName` annotations are recognized when using the elements-based API, but not with reflection since they're only `BINARY`-retained. We should be able to infer this fairly easily though - if the field or method signature names don't match the metadata name, infer and add the annotation with their names in it.

Tricky part will be trying to only do this for element handlers that require runtime retention for annotations. We could maybe add a `supportsNonRuntimeRetainedAnnotations` property to `ElementHandler` and, if that's `false`, do the custom inference. There are probably a couple of other annotations that would benefit from this that are being duplicated in the elements implementation right now (this is why they currently use `Set` 😓)

ZacSweers:
This resolves #753 

Elements API supports it out of the box, for reflection we just compare simple names of methods. Note that `JvmFieldSignature` support (mentioned in the issue) isn't done here, because there's no support for `@JvmName` on fields (TIL)

This also catches a couple corner cases around `val` properties with constant initializers having getters, where to look up field signatures when the enclosing class is an interface

### DIFF
--- a/kotlinpoet-km-specs/src/main/kotlin/com/squareup/kotlinpoet/km/specs/KmSpecs.kt
+++ b/kotlinpoet-km-specs/src/main/kotlin/com/squareup/kotlinpoet/km/specs/KmSpecs.kt
@@ -335,10 +335,19 @@ private fun ImmutableKmClass.toTypeSpec(
                       annotations += elementHandler.methodJvmModifiers(jvmInternalName, getterSignature)
                           .map { it.annotationSpec() }
                     }
-                    if (!elementHandler.supportsNonRuntimeRetainedAnnotations) {
+                    if (!isInterface && !elementHandler.supportsNonRuntimeRetainedAnnotations) {
                       // Infer if JvmName was used
+                      // We skip interface types for this because they can't have @JvmName.
+                      // For annotation properties, kotlinc puts JvmName annotations by default in
+                      // bytecode but they're implicit in source, so we expect the simple name for
+                      // annotation types.
+                      val expectedMetadataName = if (isAnnotation) {
+                        property.name
+                      } else {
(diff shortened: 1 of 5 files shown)


kotlinpoet-km-specs/src/main/kotlin/com/squareup/kotlinpoet/km/specs/KmSpecs.kt
```diff
@@ -660,6 +704,20 @@ private fun Set<PropertyAccessorFlag>.toKModifiersArray(): Array<KModifier> {
   }.toTypedArray()
 }
 
+private fun JvmMethodSignature.jvmNameAnnotation(
+  metadataName: String,
+  useSiteTarget: UseSiteTarget? = null
+): AnnotationSpec? {
+  return if (name == metadataName) {
```
Egorand: Nice

kotlinpoet-km-specs/src/main/kotlin/com/squareup/kotlinpoet/km/specs/KmSpecs.kt
```diff
@@ -681,6 +739,25 @@ private fun ImmutableKmWithFlags.addVisibility(body: (KModifier) -> Unit) {
   }
 }
 
+// TODO This is a copy of the stdlib version. Use it directly once it's out of experimental
```
Egorand: Maybe mark with `KotlinPoetKm`?

ZacSweers: No need to, it's private and doesn't use any metadata APIs

PR #756, opened 2019-08-20
invoke Kotlin compiler in FileSpecTest

sullis:


sullis:
Travis CI Build is failing:  "java.lang.NoSuchMethodError"

Egorand:
I'm actually not sure we want our tests to compile generated code. KotlinPoet doesn't guarantee that the code it produces will compile.

PR #766, opened 2019-08-29
Add test and notes for `@JvmOverloads`

ZacSweers:
Thought of this today. Works as expected in elements, but not from reflection due to the annotation retention being only `BINARY`. Would need more work for reflection in the future, perhaps if we just use `kotlin-reflect` directly

### DIFF
--- a/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
+++ b/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
@@ -1152,6 +1152,47 @@ class KotlinPoetMetadataSpecsTest(
       }
     }
   }
+
+  @IgnoreForHandlerType(
+      reason = "JvmOverloads is not runtime retained and thus not visible to reflection.",
+      handlerType = REFLECTIVE
+  )
+  @Test
+  fun overloads() {
+    val typeSpec = Overloads::class.toTypeSpecWithTestHandler()
+
+    //language=kotlin
+    assertThat(typeSpec.trimmedToString()).isEqualTo("""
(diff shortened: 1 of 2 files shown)


kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
```diff
@@ -1152,6 +1152,47 @@ class KotlinPoetMetadataSpecsTest(
       }
     }
   }
+
+  @IgnoreForHandlerType(
```
Egorand: I like this!

kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
```diff
@@ -1152,6 +1152,47 @@ class KotlinPoetMetadataSpecsTest(
...
+      reason = "JvmOverloads is not runtime retained and thus not visible to reflection.",
+      handlerType = REFLECTIVE
+  )
+  @Test
+  fun overloads() {
+    val typeSpec = Overloads::class.toTypeSpecWithTestHandler()
+
+    //language=kotlin
```
Egorand: Does Spotless require this comment?

ZacSweers: Nope this is just a helper so the IDE does syntax highlighting on the below code. Normally you can use a `@Language("Kotlin")` annotation, but since this is a parameter we're passing in, this comment applies it on the contents of that parameter instead

ZacSweers:
Fixed!

PR #770, opened 2019-08-31
Support detection of synthetic elements

ZacSweers:
This adds support to kotlin-metadata-specs for detecting and generating `@JvmSynthetic` constructs. This is a bit tricky, as it works fine in reflection but revealed an interesting limitation in elements where synthetic elements are not included in the elements API.

From the docs: https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/package-summary.html

> ...synthetic constructs in a class file, such as accessor methods used in implementing nested classes and bridge methods used in implementing covariant returns, are translation artifacts outside of this model.

As such, some aspects we read off of the bytecode directly like static, synchronized, etc are not available when using elements. We leave a comment in those cases now to indicate that this information may be missing.


ZacSweers:
This is mostly complete, only one thing I need to adjust is to ignore synthetic elements we know are copied in to enclosing classes from companion objects.

Egorand:
Looks like I introduced conflicts by not merging in order, sorry about that.

ZacSweers:
Updated!

ZacSweers:
...how are there more conflicts? I just rebased 🤔

PR #771, opened 2019-08-31
Add support for detecting Kotlin default interface functions

ZacSweers:
This adds support for detecting interface functions in Kotlin that have default implementations. This probably merits a separate deep dive blogpost, but in short they way they are implement is a synthetic `DefaultImpls` class in the interface with static functions containing the default implementation. The rules are pretty simple:

- method name is the same
- parameters are the same
- the target interface instance is added as a parameter to the beginning of these parameters

So by looking for the existing of that `DefaultImpls` class + method with the modified signature, we can infer if there is a default implementation for Kotlin. There are slightly different rules for when default parameter arguments are in play, but these methods still exist in those cases too

In short:

```kotlin
interface Foo {
  fun bar(input: String) {
    println("I'm a " + this + " saying " + input)
  }
}
```

Becomes roughly this in bytecode:

```java
interface Foo {

  void bar();

  final class DefaultImpls {
    public static void bar(Foo instance, String input) {
      System.out.println("I'm a " + instance + " saying " + input);
    }
  }
}
```

### DIFF
--- a/kotlinpoet-elementhandler-elements/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/elements/ElementsElementHandler.kt
+++ b/kotlinpoet-elementhandler-elements/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/elements/ElementsElementHandler.kt
@@ -196,6 +196,10 @@ class ElementsElementHandler private constructor(
     return method.isOverriddenIn(typeElement)
   }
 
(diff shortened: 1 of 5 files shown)


ZacSweers:
I think Kotlin wouldn't break these because changing this would break binary compatibility. There is an interesting thing about JvmDefault compatibility behavior, but even in that case we'd just defer to `@JvmDefault`

PR #772, opened 2019-09-01
Add support for Kotlin `@Throws` in metadata

ZacSweers:
This adds support for detecting `@Throws` in metadata. Like some other jvm-specific things (certain modifiers, etc), these have to be inferred from the jvm code as the annotations are not present

### DIFF
--- a/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
+++ b/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
@@ -1534,7 +1534,7 @@ class KotlinPoetMetadataSpecsTest(
 
     //language=kotlin
     assertThat(typeSpec.trimmedToString()).isEqualTo("""
-      class Throwing {
+      class Throwing @kotlin.jvm.Throws(exceptionClasses = [java.lang.IllegalStateException::class]) constructor() {
         @get:kotlin.jvm.Throws(exceptionClasses = [java.lang.IllegalStateException::class])
         @set:kotlin.jvm.Throws(exceptionClasses = [java.lang.IllegalStateException::class])
         var getterAndSetterThrows: kotlin.String? = null
--- a/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
+++ b/kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
@@ -1312,13 +1312,10 @@ class KotlinPoetMetadataSpecsTest(
       ) {
         @kotlin.jvm.JvmField
         val fieldProp: kotlin.String = TODO("Stub!")
-
         companion object {
           @kotlin.jvm.JvmField
           val companionProp: kotlin.String = ""
-
           const val constCompanionProp: kotlin.String = ""
-
           @kotlin.jvm.JvmStatic
           val staticCompanionProp: kotlin.String = ""
         }
@@ -1342,13 +1339,10 @@ class KotlinPoetMetadataSpecsTest(
       ) {
         @field:kotlin.jvm.JvmField
         val fieldProp: kotlin.String = ""
-
         companion object {
           @kotlin.jvm.JvmField
           val companionProp: kotlin.String = ""
-
           const val constCompanionProp: kotlin.String = ""
-
           @kotlin.jvm.JvmStatic
           val staticCompanionProp: kotlin.String = ""
         }
@@ -1520,6 +1514,7 @@ class KotlinPoetMetadataSpecsTest(
     interface InterfaceWithJvmName {
       @JvmSynthetic
       fun interfaceFunction()
+
       companion object {
         @JvmStatic
         @get:JvmSynthetic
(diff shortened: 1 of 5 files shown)


kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/ElementHandler.kt
```diff
@@ -129,6 +130,21 @@ interface ElementHandler {
...
   fun methodAnnotations(classJvmName: String, methodSignature: JvmMethodSignature): List<AnnotationSpec>
 
+  /**
+   * Looks up a given class method given its [JvmMethodSignature] and returns any thrown types
+   * found on it. Used for [Throws][@Throws]
+   *
+   * @param classJvmName The JVM name of the class (example: `"org/foo/bar/Baz$Nested"`).
+   * @param methodSignature The method with annotations to look up.
+   * @param isConstructor Indicates if [methodSignature] is a constructor.
+   * @return the set of found thrown types, or null.
```
Egorand: nit: Any upside in returning null vs an empty set?

kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
```diff
@@ -1192,6 +1192,47 @@ class KotlinPoetMetadataSpecsTest(
...
     }
   }
+
+  @Test
+  fun throws() {
+    val typeSpec = Throwing::class.toTypeSpecWithTestHandler()
+
+    //language=kotlin
+    assertThat(typeSpec.trimmedToString()).isEqualTo("""
+      class Throwing {
```
Egorand: Constructor exception didn't get recognized?

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -816,6 +826,19 @@ private inline fun <E> setOf(body: MutableSet<E>.() -> Unit): Set<E> {
...
 private val JVM_DEFAULT = JvmDefault::class.asClassName()
 private val JVM_STATIC = JvmStatic::class.asClassName()
+private fun createThrowsSpec(
+  exceptions: Set<TypeName>,
+  useSiteTarget: UseSiteTarget? = null
+): AnnotationSpec {
+  return AnnotationSpec.builder(Throws::class)
+      .addMember(
+          "exceptionClasses = %L",
+          exceptions.map { CodeBlock.of("%T::class", it) }
```
Egorand: nit: These won't get new lines automatically if the array is too long, this would require some manual formatting. Not important for this PR.

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/ElementHandler.kt
ZacSweers: null allows us to lean on kotlin nullability rather than checking if the returned annotations are empty

kotlinpoet-metadata-specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
ZacSweers: oooo good catch! Will look

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/ElementHandler.kt
JakeWharton: Collections should almost never be nullable unless there's some semantic meaning between non-null vs. null and empty vs. non-empty. For things like an allow/disallow list, null might mean allow all whereas empty list means allow nothing.

Since you don't need the ternary state, empty list makes more sense here. You already always have the problem of dealing with empty lists in collections. Adding null just complicates things.

ZacSweers: fair enough, will change 👍

ZacSweers:
Fixed missing constructor annotations and changed to use a non-nullable return type.

I also found a bug in `ParameterSpec` where primary constructors with no keyword emission would still emit parentheses even with no parameters. Technically legal in kotlin, but redundant

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
```diff
@@ -183,9 +183,13 @@ class ParameterSpec private constructor(
 internal fun List<ParameterSpec>.emit(
   codeWriter: CodeWriter,
   forceNewLines: Boolean = false,
+  forceParensOnEmpty: Boolean = true,
```
Egorand: I'd really prefer these changes to be in a separate PR, looks they're not strictly related to this one. Can you please pull them out and rebase this PR on top? Should be an easy merge.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -473,7 +473,7 @@ class TypeSpec private constructor(
 
     fun primaryConstructor(primaryConstructor: FunSpec?) = apply {
       check(kind == Kind.CLASS) {
-        "$kind can't have initializer blocks"
+        "$kind can't have a primary constructor"
```
Egorand: Also feels like a separate PR, even if a very small one.

ZacSweers: https://github.com/square/kotlinpoet/pull/773

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
ZacSweers: https://github.com/square/kotlinpoet/pull/774

Egorand:
Looks good! Wanna rebase against master before I merge?

ZacSweers:
Rebased!

Issue #776, opened 2019-09-04
ElementHandler API


ZacSweers:
This is an issue to discuss any changes/tuning we should make to the `ElementHandler` API for metadata parsing.

Quick context: `ElementHandler` is a basic API for assisting kotlinpoet-metadata-specs APIs in fully describing the constructs around the code the metadata represents. There are pre-built reflective and `Elements`-based implementations right now.

The API has grown a little ad-hoc, so we should try to tease out where it can be better.

Current API:

```kotlin
/**
 * A basic interface for looking up information about JVM elements.
 */
@KotlinPoetMetadataPreview
interface ElementHandler {

  interface JvmModifier {
    fun annotationSpec(): AnnotationSpec
  }

  /** Modifiers that are annotations in Kotlin but modifier keywords in bytecode. */
  enum class JvmFieldModifier : JvmModifier {
    STATIC {
      override fun annotationSpec(): AnnotationSpec = AnnotationSpec.builder(JvmStatic::class.asClassName()).build()
    },
    TRANSIENT {
      override fun annotationSpec(): AnnotationSpec = AnnotationSpec.builder(Transient::class.asClassName()).build()
    },
    VOLATILE {
      override fun annotationSpec(): AnnotationSpec = AnnotationSpec.builder(Volatile::class.asClassName()).build()
    };
  }

  /** Modifiers that are annotations in Kotlin but modifier keywords in bytecode. */
  enum class JvmMethodModifier : JvmModifier {
    STATIC {
      override fun annotationSpec(): AnnotationSpec = AnnotationSpec.builder(JvmStatic::class.asClassName()).build()
    },
    SYNCHRONIZED {
      override fun annotationSpec(): AnnotationSpec = AnnotationSpec.builder(Synchronized::class.asClassName()).build()
    }
  }

  /**
   * Indicates if this element handler supports [AnnotationRetention.RUNTIME]-retained annotations.
   * This is used to indicate if manual inference of certain non-RUNTIME-retained annotations should
   * be done, such as [JvmName].
   */
  val supportsNonRuntimeRetainedAnnotations: Boolean

  /**
   * Looks up other classes, such as for nested members. Note that this class would always be
   * Kotlin, so Metadata can be relied on for this.
   *
   * @param jvmName The JVM name of the class (example: `"org/foo/bar/Baz$Nested"`).
   * @return the read [ImmutableKmClass] from its metadata. If no class was found, this should throw
   *         an exception.
   */
[...]

PR #777, opened 2019-09-04
Fix class name splicing

ZacSweers:
This is a bit of an oversight - before, every test was assuming that every class name in metadata followed this format:

`package/to/class.MyClass`

When in fact this was just coincidence due to all the test classes being nested classes. The actual format is:

`package/to/class/MyClass.Nested` or `package/to/class/MyClass$Nested`

This also made a red herring out of `inline` classes, which can _only_ be top level classes. Thus when that test was set up, it was presumed that the class naming scheme was special, and thus special-cased in the spec generation 😅

### DIFF
--- a/kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
+++ b/kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
@@ -208,10 +208,11 @@ private fun ImmutableKmClass.toTypeSpec(
 
   // package/of/class/MyClass.InnerClass
   // Sometimes it's package/of/class/MyClass$InnerClass
-  val simpleName = name
-      .substringAfterLast("/") // Drop the package name, e.g. "package/of/class/"
-      .substringAfterLast(".") // Drop any enclosing classes, e.g. "MyClass."
-      .substringAfterLast("$") // Drop any enclosing classes, e.g. "MyClass$"
+  val simpleName = name.substringAfterLast(charArrayOf(
+      '/', // Drop the package name, e.g. "package/of/class/"
+      '.', // Drop any enclosing classes, e.g. "MyClass."
+      '$'  // Drop any enclosing classes, e.g. "MyClass$"
+  ))
   val jvmInternalName = name.jvmInternalName
   val builder = when {
     isAnnotation -> TypeSpec.annotationBuilder(simpleName)
@@ -923,6 +924,11 @@ private fun createThrowsSpec(
       )
       .useSiteTarget(useSiteTarget)
       .build()
+}
+
+private fun String.substringAfterLast(delimiters: CharArray): String {
+  val index = lastIndexOfAny(delimiters)
+  return if (index == -1) this else substring(index + 1, length)
 }
 
 @PublishedApi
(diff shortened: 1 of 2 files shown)


kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -206,7 +206,13 @@ private fun ImmutableKmClass.toTypeSpec(
 ): TypeSpec {
   val classTypeParamsResolver = typeParameters.toTypeParamsResolver()
 
-  val simpleName = name.substringAfterLast(if (isInline) "/" else ".")
+  // package/of/class/MyClass.InnerClass
+  // Sometimes it's package/of/class/MyClass$InnerClass
+  val simpleName = name.substringAfterLast(charArrayOf(
+      '/', // Drop the package name, e.g. "package/of/class/"
+      '.', // Drop any enclosing classes, e.g. "MyClass."
+      '$'  // Drop any enclosing classes, e.g. "MyClass$"
```
JakeWharton: What happens for
```kotlin
class `Foo$Bar`
```
?

ZacSweers:
I can't actually find any rhyme or reason to when it uses a dollar sign for the nested class version dot. The classes nested in the test suite are dots and other nested ones in the new test are dollar signs

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -920,6 +926,11 @@ private fun createThrowsSpec(
       .build()
 }
 
+private fun String.substringAfterLast(delimiters: CharArray): String {
```
JakeWharton: `vararg Char` is more common in the stdlib

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
ZacSweers: It becomes `com/squareup/kotlinpoet/metadata/specs/test/Foo$Bar`

I'm not really sure what to do about this though :|. Any suggestions?

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
ZacSweers: 970994b

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
JakeWharton: There will be an InnerClass attribute in the classfile when it's an actually-nested type. Not sure if that's exposed to you or not from wherever you're reading.

ZacSweers: Hacking on this - We can solve 95% of this case by just asking the source `Class`/`TypeElement`/`KClass` for its simple name, and the metadata gives us the simple name when we iterate over subtypes. The remaining 5% would just be cases where someone calls `ImmutableKmClass.toTypeSpec`, in which case we can accept an optional `simpleName` that defaults to the current implementation as a "best guess" approach. We can tune that best guess approach via element handler as well in the future to check the existence of that type and whatnot.

I have a mostly working implementation of this locally, but due to how Elements tries to canonicalize names, I think we should push to a separate PR  pending its API cleanup #776, where we could give it a hint of what the known simple name is.

PR #780, opened 2019-09-09
Parse parameter annotations in metadata

ZacSweers:
This implements support for parsing and rendering parameter annotations on metadata. It requires another `ElementHandler` API (but can be cleaned up as part of #776).

Interesting insight - Kotlin does roughly the following decision tree for implicit parameter annotation site targets
1. If an annotation is applicable to parameters, then the implicit site-target is `param`
2. If the annotation is applicable to properties, then the implicit target is `property`
3. If the annotation is applicable to fields, then the implicit target is `field`

This tries to make that explicit by, if a site-target-less annotation was found on a property but that property is a constructor param, it applies `property:` designation to it. This is legal for field-only annotations as well like `@JvmField`, but redundant. We can maybe do a lookup to read the given annotation's targets, but for now I figure it's better to err for correctness rather than avoiding redundancy in the rendered source. This is implemented in 42abb39

### DIFF
--- a/kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
+++ b/kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
@@ -200,7 +200,7 @@ class ReflectiveElementHandler private constructor() : ElementHandler {
     return try {
       parameters[index]
           .declaredAnnotations
-          .map { AnnotationSpec.get(it, true) }
+          .map { AnnotationSpec.get(it, includeDefaultValues = true) }
           .filterOutNullabilityAnnotations()
     } catch (e: ClassNotFoundException) {
       emptyList()
--- a/kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
+++ b/kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
@@ -186,6 +186,23 @@ class ReflectiveElementHandler private constructor() : ElementHandler {
     }
   }
 
+  override fun parameterAnnotations(classJvmName: String, methodSignature: JvmMethodSignature,
+      index: Int, isConstructor: Boolean): List<AnnotationSpec> {
+    val parameters = if (isConstructor) {
+      lookupConstructor(classJvmName, methodSignature)!!.parameters
+    } else {
+      lookupMethod(classJvmName, methodSignature)!!.parameters
+    }
+    return try {
+      parameters[index]
+          .declaredAnnotations
+          .map { AnnotationSpec.get(it, true) }
+          .filterOutNullabilityAnnotations()
+    } catch (e: ClassNotFoundException) {
+      emptyList()
+    }
+  }
+
   override fun isMethodSynthetic(
     classJvmName: String,
     methodSignature: JvmMethodSignature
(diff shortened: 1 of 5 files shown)


kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
```diff
@@ -186,6 +186,27 @@ class ReflectiveElementHandler private constructor() : ElementHandler {
...
+  ): List<AnnotationSpec> {
+    val parameters = if (isConstructor) {
+      lookupConstructor(classJvmName, methodSignature)!!.parameters
+    } else {
+      lookupMethod(classJvmName, methodSignature)!!.parameters
+    }
+    return try {
+      parameters[index]
+          .declaredAnnotations
+          .map { AnnotationSpec.get(it, true) }
```
Egorand: nit: Make `true` a named arg for better readability.

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -639,14 +661,21 @@ private fun ImmutableKmConstructor.toFunSpec(
...
   return FunSpec.builder(name)
       .apply {
         addAnnotations(annotations)
         addVisibility { addModifiers(it) }
         if (valueParameters.isNotEmpty()) {
-          addParameters(valueParameters.map { it.toParameterSpec(typeParamResolver) })
+          addParameters(valueParameters.mapIndexed { index, param ->
+            param.toParameterSpec(
+                typeParamResolver,
+                param.extractAnnotations(elementHandler, classJvmName, signature, index, false)
```
Egorand: nit: Named arg `false`.

kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
```diff
@@ -186,6 +186,27 @@ class ReflectiveElementHandler private constructor() : ElementHandler {
...
+    isConstructor: Boolean
+  ): List<AnnotationSpec> {
+    val parameters = if (isConstructor) {
+      lookupConstructor(classJvmName, methodSignature)!!.parameters
+    } else {
+      lookupMethod(classJvmName, methodSignature)!!.parameters
+    }
+    return try {
+      parameters[index]
+          .declaredAnnotations
```
Egorand: Does `Metadata` contain use-site targets info? If yes we should honor them.

ZacSweers: It does, but afaik only for declared ones. For implicit ones, it's not specified afaik

kotlinpoet-elementhandler-reflective/src/main/kotlin/com/squareup/kotlinpoet/elementhandler/reflective/ReflectiveElementHandler.kt
ZacSweers: 85ae0a5

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
ZacSweers: 85ae0a5

PR #783, opened 2019-09-13
Restore ABI compatibility to TypeName#copy

### LINKED ISSUE #782: Adding a new default parameter to TypeName#copy was a breaking change
`tags` was added to the signature of `copy()` below.

```kotlin
abstract fun copy(
    nullable: Boolean = this.isNullable,
    annotations: List<AnnotationSpec> = this.annotations.toList(),
    tags: Map<KClass<*>, Any> = this.tagMap.tags.toMap()
  ): TypeName
```

While this isn't a breaking change for source, it is actually a binary breaking change because the synthetic defaults method signature did change, meaning any previous users of this function would break.

```
e: [kapt] An exception occurred: java.lang.NoSuchMethodError: com.squareup.kotlinpoet.TypeName.copy$default(Lcom/squareup/kotlinpoet/TypeName;ZLjava/util/List;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/TypeName;
        at com.squareup.moshi.kotlin.codegen.MetadataKt.asTypeName(metadata.kt:205)
        at com.squareup.moshi.kotlin.codegen.MetadataKt.declaredProperties(metadata.kt:381)
        at com.squareup.moshi.kotlin.codegen.MetadataKt.targetType(metadata.kt:322)

```

Possible solutions are to revert the change or add an overload of copy

ZacSweers:
Resolves #782

### DIFF
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
@@ -206,13 +206,13 @@ class CodeBlockTest {
     val typeBlock = CodeBlock.of("%T", type)
     assertThat(typeBlock.toString()).isEqualTo("kotlin.String?")
 
-    val list = List::class.asClassName().copy(nullable = true)
+    val list = (List::class.asClassName().copy(nullable = true) as ClassName)
         .parameterizedBy(Int::class.asTypeName().copy(nullable = true))
         .copy(nullable = true)
     val listBlock = CodeBlock.of("%T", list)
     assertThat(listBlock.toString()).isEqualTo("kotlin.collections.List<kotlin.Int?>?")
 
-    val map = Map::class.asClassName().copy(nullable = true)
+    val map = (Map::class.asClassName().copy(nullable = true) as ClassName)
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -206,13 +206,13 @@ class CodeBlockTest {
     val typeBlock = CodeBlock.of("%T", type)
     assertThat(typeBlock.toString()).isEqualTo("kotlin.String?")
 
-    val list = List::class.asClassName().copy(nullable = true)
+    val list = (List::class.asClassName().copy(nullable = true) as ClassName)
```
Egorand: We should fix this to return self-types instead of `TypeName`, this is pretty awkward to use at the moment.

ZacSweers: Yep, a v2 could be something like

```kotlin
sealed class TypeName<T : TypeName> {
  abstract fun copy(...): T
}

class ClassName : TypeName<ClassName> {

}
```

PR #792, opened 2019-09-19
Rename ElementHandler -> ClassInformer

ZacSweers:


kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/ClassInspector.kt
```diff
@@ -20,14 +20,12 @@ import com.squareup.kotlinpoet.metadata.ImmutableKmClass
 import com.squareup.kotlinpoet.metadata.KotlinPoetMetadataPreview
 import kotlinx.metadata.jvm.JvmMethodSignature
 
-/**
- * A basic interface for looking up information about JVM elements.
- */
+/** A basic interface for looking up information about JVM information about a given Class. */
```
Egorand:
```suggestion
/** A basic interface for looking up JVM information about a given Class. */
```

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/ClassInspector.kt
```diff
@@ -20,14 +20,12 @@ import com.squareup.kotlinpoet.metadata.ImmutableKmClass
...
 @KotlinPoetMetadataPreview
-interface ElementHandler {
+interface ClassInspector {
 
   /**
-   * Indicates if this element handler supports [AnnotationRetention.RUNTIME]-retained annotations.
+   * Indicates if this informer supports [AnnotationRetention.RUNTIME]-retained annotations.
```
Egorand:
```suggestion
   * Indicates if this [ClassInspector] supports [AnnotationRetention.RUNTIME]-retained annotations.
```

kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -271,11 +271,11 @@ private fun ImmutableKmClass.toTypeSpec(
       builder.addModifiers(INNER)
     }
     builder.addTypeVariables(typeParameters.map { it.toTypeVariableName(classTypeParamsResolver) })
-    // If we have an element handler, we can check exactly which "supertype" is an interface vs
+    // If we have an informer, we can check exactly which "supertype" is an interface vs
```
Egorand:
```suggestion
    // If we have an inspector, we can check exactly which "supertype" is an interface vs
```

ZacSweers:
Updated with the suggestions 👍. Good catch

PR #794, opened 2019-09-21
Throw IllegalArgumentException when trying to escape illegal characters

### LINKED ISSUE #701: Improve name escaping logic
As demonstrated by #698, we incorrectly assume that any characters escaped by backticks form a correct name. This is not true, e.g. this will not compile:
```kotlin
class `My.Class` // dots are illegal, even after escaping
```
We should fix name validation logic and throw if the name contains characters that can't be escaped.

Wasabi375:
fixes #701

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -295,12 +295,19 @@ class TypeSpec private constructor(
     for (property in propertySpecs) {
       val parameter = primaryConstructor.parameter(property.name) ?: continue
       if (parameter.type != property.type) continue
-      if (CodeBlock.of("%N", parameter) != CodeBlock.of("%N", property.initializer.toString()))
+      if (!isPropertyInitializerConstructorParameter(property, parameter))
         continue
 
       result[property.name] = property.fromPrimaryConstructorParameter(parameter)
     }
     return result
+  }
+
+  /**
+   * Returns `true if the property can be declared inline as a constructor parameter
+   */
+  private fun isPropertyInitializerConstructorParameter(property: PropertySpec, parameter: ParameterSpec): Boolean {
+    return CodeBlock.of("%N", parameter).toString() == property.initializer.toString().escapeIfNotJavaIdentifier().escapeIfKeyword()
   }
 
   /**
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -304,10 +304,15 @@ class TypeSpec private constructor(
   }
 
   /**
-   * Returns `true if the property can be declared inline as a constructor parameter
+   * Returns true if the property can be declared inline as a constructor parameter
    */
-  private fun isPropertyInitializerConstructorParameter(property: PropertySpec, parameter: ParameterSpec): Boolean {
-    return CodeBlock.of("%N", parameter).toString() == property.initializer.toString().escapeIfNotJavaIdentifier().escapeIfKeyword()
+  private fun isPropertyInitializerConstructorParameter(
+    property: PropertySpec,
+    parameter: ParameterSpec
+  ): Boolean {
+    val parameterName = CodeBlock.of("%N", parameter).toString()
+    val initializerCode = property.initializer.toString().escapeIfNotJavaIdentifier().escapeIfKeyword()
+    return parameterName == initializerCode
   }
 
   /**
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -218,3 +225,6 @@ private val KEYWORDS = setOf(
     "interface",
     "typeof"
 )
+
+// https://github.com/JetBrains/kotlin/blob/master/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/checkers/JvmSimpleNameBacktickChecker.kt
```
Egorand: Nice!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -295,14 +295,21 @@ class TypeSpec private constructor(
...
 
       result[property.name] = property.fromPrimaryConstructorParameter(parameter)
     }
     return result
   }
 
+  /**
+   * Returns `true if the property can be declared inline as a constructor parameter
+   */
+  private fun isPropertyInitializerConstructorParameter(property: PropertySpec, parameter: ParameterSpec): Boolean {
```
Egorand: nit: Feels like these lines are too long, I'm surprised that Spotless didn't catch it.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -15,6 +15,7 @@
  */
 package com.squareup.kotlinpoet
 
+import java.lang.IllegalArgumentException
```
Egorand: I think we should throw `kotlin.IllegalArgumentException` instead. Just removing the import should do the job.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -151,7 +152,13 @@ internal fun String.escapeIfKeyword() = if (isKeyword) "`$this`" else this
 internal fun String.escapeIfNotJavaIdentifier() =
   if (!Character.isJavaIdentifierStart(first()) || drop(1).any { !Character.isJavaIdentifierPart(it) }) "`$this`" else this
 
-internal fun String.escapeIfNecessary() = escapeIfNotJavaIdentifier().escapeIfKeyword()
+internal fun String.escapeIfNecessary() = escapeIfNotJavaIdentifier().escapeIfKeyword().failIfEscapeInvalid()
+
+internal fun String.failIfEscapeInvalid() =
+  if (any { it in illegalCharactersToEscape })
```
Egorand: nit: Let's use `{}` with this if-else block.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -218,3 +225,6 @@ private val KEYWORDS = setOf(
     "interface",
     "typeof"
 )
+
+// https://github.com/JetBrains/kotlin/blob/master/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/checkers/JvmSimpleNameBacktickChecker.kt
+private val illegalCharactersToEscape = setOf('.', ';', '[', ']', '/', '<', '>', ':', '\\')
```
Egorand: nit: Should the name be in UPPER_SNAKE_CASE, since it's a constant?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/KotlinPoetTest.kt
```diff
@@ -884,4 +884,13 @@ class KotlinPoetTest {
       |}
       |""".trimMargin())
   }
+
+  // https://github.com/square/kotlinpoet/issues/701
+  @Test fun noIllegalCharacterInIdentifier() {
+    assertThrows<java.lang.IllegalArgumentException> {
```
Egorand: nit: Remove `java.lang.`.

Wasabi375: Not sure how `java.lang` got in there in the first place. Same above

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
Wasabi375: I changed it to a require. After adding a few braces intellij suggested it and I think it's better this way.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Wasabi375: I guess it's just short enough. Should be better now though

Wasabi375:
If the CLA is the same as for Moshi I signed it.

PR #798, opened 2019-09-24
Disallow empty ClassName simple names

### LINKED ISSUE #795: ClassName empty name ambiguity
This will pass right now on master

```kotlin
val name1 = ClassName("", "", "Taco")
val name2 = ClassName("", "Taco")
assertThat(name1).isEqualTo(name2)
```

This is because their `toString()` is used for equality, and their `toString()`s are surprisingly the same. Their canonical names are different (`".Taco"` and `"Taco"`, respectively).

Not sure if canonical name should match the toString() behavior, toString() should account for this, or if equality should just compare names in an override instead. Or some combination of these.

Egorand:
Closes #795

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -309,6 +309,9 @@ class ClassName internal constructor(
   constructor(packageName: String, vararg simpleNames: String) :
       this(listOf(packageName, *simpleNames)) {
     require(simpleNames.isNotEmpty()) { "simpleNames must not be empty" }
+    require(simpleNames.none { it.isEmpty() }) {
+      "simpleNames must not contain empty items: ${simpleNames.contentToString()}"
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -309,6 +309,10 @@ class ClassName internal constructor(
   constructor(packageName: String, vararg simpleNames: String) :
       this(listOf(packageName, *simpleNames)) {
     require(simpleNames.isNotEmpty()) { "simpleNames must not be empty" }
+    require(simpleNames.none { it.isEmpty() }) {
+      "simpleNames must not contain empty items: " +
+          simpleNames.joinToString(prefix = "[", postfix = "]")
```
ZacSweers: Kotlin's list toString() does this format automatically

Egorand: Yeah good catch, copy-pasta from the vararg constructor.

PR #803, opened 2019-09-26
adding ability for fun receivers to be nullable

drmaas:


JakeWharton:
Doesn't this already work with nullable types? The type is what carries the nullness information, not the location the type is used.

drmaas:
Thanks @JakeWharton - I found I can use ```.receiver(className.copy(nullable = true))```

PR #805, opened 2019-09-28
Setup an action to build every pull request

Egorand:


ShaishavGandhi:
There's some prior art for jdk11 in RxDogTag that @ZacSweers did. 
https://github.com/uber/RxDogTag/pull/50/files#diff-e9f950f17198d3d5e3122a44230a09b9R12

ZacSweers:
I found it simplest to just set up one push action and just limit the uploadArchives task to only run on master pushes. Unless you've got separate workflows that you need the separate build actions, you're just going to duplicate things and have to keep them in sync.

Egorand:
Yeah, that's a good idea. I kinda wanted to have two different actions running side by side, similar to Travis, but that's not a big deal.

PR #806, opened 2019-09-30
Read type annotations from metadata

ZacSweers:
Most of the time we can just read annotations from `ClassInspector`, but there are certain cases where the annotation is only available in metadata. Namely (per the doc): 

> annotations on type parameters and types are serialized to the Kotlin metadata.

Types can mean `KmType`, `KmTypeParameter`, and `KmTypeAlias`. We don't support KmTypeAlias -> TypeAliasSpec yet, but that'll be a different PR


ZacSweers:
I'm planning to add an integration test in the specs test project as well. Some real world examples and also make sure the code we're generating actually compiles 😬

PR #811, opened 2019-10-15
Add explicit "Extension functions" section to README

petedmarsh:
I was looking for an example of how to generate an extension function in the README recently and I missed it due to it being part of the example of single-statement functions. This splits the two out into separate sections - I hope this will help future READM-ers Ctrl-F their way to the right information.


README.md
```diff
@@ -614,12 +614,36 @@ The other modifiers work where permitted.
...
-Also, KotlinPoet can recognize single-expression functions and print them out properly. It treats
+#### Extension functions
+
+Extension functions can be generated by specifying a `receiver`.
+
+```kotlin
+val square = FunSpec.builder("square")
+    .receiver(Int::class)
+    .returns(Int::class)
+    .addStatement("var s = this * this")
```
Egorand: nit: The variable name is `s` here and `p` in the snippet below.

petedmarsh:
@Egorand I've signed the CLA and I've fixed the code examples (there was a second mistake!)

PR #815, opened 2019-10-19
Properly handle % sign as KDoc suffix

### LINKED ISSUE #814: Percentage at the end of parameter kdoc removes line break
If you do something like this:
```
val parameterSpec1 = parameterSpecBuilder1.addKdoc("Progress in %%").build()
val parameterSpec2 = parameterSpecBuilder2.addKdoc("Some other parameter with %%").build()
funSpecBuilder.addParameters(listOf(parameterSpec1, parameterSpec2))
```

The output will be (one line)
```
* @param param1Name Progress in %@param param2Name Some other parameter with % */
```
Instead of (three lines)
```
* @param param1Name Progress in %
* @param param2Name Some other parameter with % 
*/
```

Egorand:
Fixes #814

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -172,17 +172,15 @@ internal val String.isKeyword get() = this in KEYWORDS
 internal val String.isName get() = split("\\.").none { it.isKeyword }
 
 internal fun CodeBlock.ensureEndsWithNewLine() = if (isEmpty()) this else with(toBuilder()) {
-  val lastFormatPart = formatParts.last()
-  if (CodeBlock.isPlaceholder(lastFormatPart)) {
-    if (args.isEmpty()) {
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -172,17 +172,15 @@ internal val String.isKeyword get() = this in KEYWORDS
 internal val String.isName get() = split("\\.").none { it.isKeyword }
 
 internal fun CodeBlock.ensureEndsWithNewLine() = if (isEmpty()) this else with(toBuilder()) {
-  val lastFormatPart = formatParts.last()
-  if (CodeBlock.isPlaceholder(lastFormatPart)) {
-    if (args.isEmpty()) {
-      return@with build()
-    }
+  val lastFormatPart = trim().formatParts.last()
```
Egorand: `trim()` ensures we're ignoring indentation and statement placeholders.

PR #817, opened 2019-10-19
Setup Github Actions

Egorand:
Largely copied from https://github.com/uber/RxDogTag/blob/master/.github/workflows/ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+    name: JDK ${{ matrix.java_version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java_version: [1.8, 11]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Install JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
+      - name: Configure Gradle
+        # Initial gradle configuration, install dependencies, etc
+        run: ./gradlew help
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  build:
+    name: JDK ${{ matrix.java_version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java_version: [1.8, 11]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Install JDK
```
ZacSweers: GitHub fixed a bug that prevented using replacements in names, so you could do `Install JDK ${{ matrix.java_version }}` here if you want. Haven't verified it works yet but someone else on the community forum confirmed it

ZacSweers:
Fun thing to do on these - compare how fast the build was on github vs travis :)

image

.github/workflows/ci.yml
Egorand: Oh cool, yeah, I'll do it!

Egorand:
> compare how fast the build was on github vs travis
Totally, I'm very hyped about it!

Egorand:
One thing I'm not sure about is whether we need both push and pr, cause they're different - while push simply builds the branch, pr rebases on the latest master and builds, which can uncover different types of issues. @ZacSweers do you know if there's a way to declare both in YAML without duplicating steps setup?

ZacSweers:
Didn't realize pr rebases against latest master - that's neat! Unfortunately I don't think there's an easy way to share them unfortunately. Been a month or so since I last tried though

PR #824, opened 2019-10-28
Emit initializers in order with properties

### LINKED ISSUE #800: Order matters for class initializers, but kotlinpoet always emits them after properties
Consider the following class

```kotlin
class Taco(vararg seasonings: String) {
  private val firstSeasoning = seasonings[0]
  init {
    require(seasonings.isNotEmpty()) {
      "Seasonings cannot be empty!"
    }
  }
}

class TacoTest {
  @Test fun runTest() {
    Taco()
  }
}
```

This would throw an `IndexOutOfBoundsException` because `firstSeasoning` is initialized first

Now if we change the class

```kotlin
class Taco(vararg seasonings: String) {
  init {
    require(seasonings.isNotEmpty()) {
      "Seasonings cannot be empty!"
    }
  }
  private val firstSeasoning = seasonings[0]
}
```

Now the test would throw an `IllegalArgumentException` from the `require()` call

ZacSweers:
This change adds tracking for initializer blocks to know what property index it should be emitted with respect to.

Resolves #800

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -62,6 +62,7 @@ class TypeSpec private constructor(
   val enumConstants = builder.enumConstants.toImmutableMap()
   val propertySpecs = builder.propertySpecs.toImmutableList()
   val initializerBlock = builder.initializerBlock.build()
+  val initializerIndex = builder.initializerIndex
   val funSpecs = builder.funSpecs.toImmutableList()
   val typeSpecs = builder.typeSpecs.toImmutableList()
   internal val nestedTypesSimpleNames = typeSpecs.map { it.name }.toImmutableSet()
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -419,6 +443,7 @@ class TypeSpec private constructor(
     internal var primaryConstructor: FunSpec? = null
     internal var superclass: TypeName = ANY
     internal val initializerBlock = CodeBlock.builder()
+    var initializerIndex = -1
```
Egorand: Should be `internal`

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -291,8 +305,16 @@ class TypeSpec private constructor(
   private fun constructorProperties(): Map<String, PropertySpec> {
     if (primaryConstructor == null) return emptyMap()
 
+    // Properties added after the initializer are not permitted to be inlined into the constructor
```
Egorand: Something that we can improve in future?

ZacSweers: Nope it's not allowed in kotlin

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
ZacSweers: I kept it public because someone calling `toBuilder()` could want to change this order

PR #826, opened 2019-11-01
Replace compile with implementation as compile is deprecated

ligi:


ligi:
Oh dam - this form asks for a phone number - I do not have one (and neither would I put one in a google form If I would have one)

Egorand:
@ligi I _think_ it shouldn't be mandatory, but let me ask around and get back to you.

ligi:
Thanks - there is the required star on this field:
image

Egorand:
Superseded by #951

PR #831, opened 2019-11-09
Enable actions on PRs

ZacSweers:
Apparently push will only run when a collaborator on the repo opens a PR, and pull_request is necessary to have CI run for external contributors. 

This doesn't duplicate the build jobs though if the PR author is a collaborator - [example](https://github.com/uber/AutoDispose/pull/399).


ZacSweers:
Didn't notice this before because travis CI was still on and catching failures before :)

image

PR #834, opened 2019-11-16
Build against stdlib-jdk8

Egorand:
@JakeWharton what are backwards compat implications of this change?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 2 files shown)


JakeWharton:
Is there a motivation for this? You would really only do this if the bytecode was targeting 1.8.

Egorand:
No strong motivation, just feels weird to target jdk7.

JakeWharton:
I would expect this change to come with a bytecode target update to match. Otherwise there's really no advantage to doing this.

PR #849, opened 2019-12-25
Clean up bestGuessClassName usage

ZacSweers:
This switches us to use `bestGuessClassName` to be used anywhere that a Km classname is looked up, and also renames it to `createClassName` since it's no longer a best guess effort.


ZacSweers:
Ran into this bug locally in an internal annotation processor due to a class that had a capitalized package name 🙄

kotlinpoet-metadata-specs/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/internal/ClassInspectorUtilTest.kt
```diff
@@ -11,56 +11,56 @@ import kotlin.test.Test
 class ClassInspectorUtilTest {
 
   @Test fun bestGuess_simple() {
```
Egorand: nit: rename tests too?

ZacSweers: good call 14bffb3

PR #852, opened 2019-12-26
Escape typealias names

Egorand:


Egorand:
Hmm, that's interesting, the following test:

```kotlin
@Test fun implTypeAlias() {
  val typeName = AtomicReference::class.asClassName().parameterizedBy(TypeVariableName("V"))
  val typeAliasSpec = TypeAliasSpec
      .builder("AtomicRef<V>", typeName)
      .addModifiers(KModifier.ACTUAL)
      .build()

  assertThat(typeAliasSpec.toString()).isEqualTo("""
      |actual typealias AtomicRef<V> = java.util.concurrent.atomic.AtomicReference<V>
      |""".trimMargin())
}
```

now fails with:

```
java.lang.IllegalArgumentException: Can't escape identifier `AtomicRef<V>` because it contains illegal characters: <>
```

We should probably allow passing a `TypeName` as the name of the typealias and treat it differently than opaque strings.

JakeWharton:
Or teach typealias about parameterization. The name of that typealias is
still "AtomicRef".

Egorand:
Ha, forgot that `TypeAliasSpec` already supports type parameters. Rewrote the test to use proper APIs.

PR #865, opened 2020-01-08
Tag AnnotationSpec with the annotation mirror where available

ansman:
This will make it possible to read annotation arguments in an annotations processor.


kotlinpoet-metadata-specs/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/internal/KmAnnotations.kt
```diff
@@ -34,6 +35,7 @@ internal fun KmAnnotation.toAnnotationSpec(): AnnotationSpec {
           addMember("%L = %L", name, arg.toCodeBlock())
         }
       }
+      .tag(this)
```
ansman: This seemed like an oversight, most other specs are tagged with the metadata object.

PR #871, opened 2020-01-15
Support import operator

### LINKED ISSUE #870: Import extension operators
I have defined an extension operator like this:
```Kotlin
package foo.bar.baz

operator fun MyClass.plus(other: MyClass): MyClass {}
```

How I can import it in kotlinpoet?
```Kotlin
FunSpec.builder("myFunction")
    .addParameter("value", MyClass::class)
    .returns(MyClass::class)
    .addStatement("return this.value + value")
    .build()
```
It will generate:
```Kotlin
fun myFunction(value: MyClass): MyClass {
   return this.value + value // foo.bar.baz.plus not import
}
```

In my use case, I can't get `FileSpec.Builder`  in this timing. How can I import it?

devkanro:
# Compose poetry with operators
This PR adds `%O` placeholder for operators to fix #870.

## Checklist
- [x] Implementation
- [x] Test
- [x] Documents
- [x] CI build pass
- [ ] Review

## Description
Similar to `MemberName`, KotlinPoet has a special placeholder for **operators**, which comes handy when 
your code needs to access top-level and members declared operators. Use **`%O`** to  reference operators,
pass an instance of `MemberName` as the argument for the placeholder, and KotlinPoet will handle imports
automatically in most cases:

```kotlin
val rangeTo = MemberName("com.squareup.tacos", "rangeTo")
val not = MemberName("com.squareup.tacos", "not")
val file = FileSpec.builder("com.squareup.example", "TacoTest")
        .addStatement("%OAny()", not)
        .addStatement("Any()%OAny()", rangeTo)
        .build())
    .build()
println(file)
```

The code above generates the following file:

```kotlin
package com.squareup.example

import com.squareup.tacos.not
import com.squareup.tacos.rangeTo

fun main() {
    !Any()
    Any()..Any()
}
```

As you can see, it's also possible to use `%O` to reference extension operator functions. KotlinPoet converts function name to operator correctly in the most cast. But sometimes, a function could be mapping to multi operators, like: `compareTo` could be mapping to `>`, `<`, `>=` and `<=`. In those cases, you must assign operator to `MemberName`.

```kotlin
val compareTo = MemberName("com.squareup.tacos", "compareTo", ">=")
val equals = MemberName("com.squareup.tacos", "equals")
val equals2 = MemberName("com.squareup.tacos", "equals", "!=")
val file = FileSpec.builder("com.squareup.example", "TacoTest")
        .addStatement("Any() %O Any()", compareTo)
        .addStatement("Any() %O Any()", equals)
        .addStatement("Any() %O Any()", equals2)
        .build())
    .build()
println(file)
```

The code above generates the following file:

```kotlin
package com.squareup.example

import com.squareup.tacos.compareTo
import com.squareup.tacos.equals

fun main() {
    Any() >= Any()
    Any() == ()
    Any() != Any()
}
```


devkanro:
You must use the `%O` to reference operators.

If you have a function named `plus`, and reference it by `%O`, I guess the function must be an operator function.

devkanro:
@ZacSweers 
We can't get a `MemberName` referenced function is an operator or not, we just have it's package, name, mapping it to operator by name is the last resort.

ZacSweers:
Right now there is no type safety for the user that the membername they are trying to use as an operator is a valid operator. It would fail not even at runtime, but at compile time of the generated code (even later!)

Also, as mentioned already, there is nothing here that checks appropriate signatures. This code appears to assume that operators are only defined in name but they are more than that.

devkanro:
Yes, it would fail at the compile time of the generated code, but I think developers need to responsible for this failure. When a developer wants to use `%O` to reference an operator, the developer needs to check the signature of `MemberName`.

As same as the `%M` and `%N`, if developers want to use the `%M` or `%N` to reference a `MemberName`. `%M` will add the import for it, but `%N` will not. 
Developers need to check the `MemberName` is an extension or not, and use `%N` for non-extension, `%M` for extension.
If developers use the wrong placeholder, the generated code will fail at compile time, I think developers need to responsible for this failure.

In most cases, the name-based operator mapping is working fine, most function names are one-to-one correspondence with the operator. Just `compareTo` and `equals` have multi mapping.

ZacSweers:
Sorry but I'm not really convinced by this in its current form. A major feature of kotlinpoet is that it's strongly typed where possible, and this is not in that spirit.

devkanro:
OK, got it. Sorry about that.
You can close this PR, and also for issue #870

Egorand:
Thanks for your PR @devkanro! I still see value in providing better support for importable operator functions, but I don't think we should be introducing a new placeholder to achieve this. Here's what we can do:

- Add a new `operator` field to `MemberName`. You can either provide the `operator` or `simpleName` but never both.
- Add an enum representing all supported operator functions and use it as the type for `operator`. Every enum value has the operator (`+`) and the name (`plus`).
- `CodeWriter` will emit the operator if it's set and use the name for imports.

What do you think?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -320,4 +320,35 @@ class MemberNameTest {
...
+    assertThat(file.toString()).isEqualTo("""
+      |package com.example
+      |
+      |import com.squareup.tacos.compareTo
+      |import com.squareup.tacos.equals
+      |import com.squareup.tacos.not
+      |import com.squareup.tacos.rangeTo
+      |
+      |fun makeTastyTacos() {
+      |  !Any()..Any()
```
Egorand: `!Any()..Any()` doesn't make much sense to me, something like `for (i in 0..10)` would be much easier to understand. Unit tests are part of KotlinPoet's documentation that we expect people to read, so writing them in the form of recipes that somebody can just copy into their codebase is great.

devkanro:
@Egorand 
I have changed the implementation for this PR, please reopen this PR and issue #870.

Check it.

Egorand:
@devkanro Github won't allow me to reopen this PR ("operator branch was force-pushed or recreated"), can you please open another one?

devkanro:
ok

Issue #878, opened 2020-02-08
KModifier.EXPECT type modifier is not propagated more than one TypeSpec deep


JustinMullin:
I have a `FunSpec` nested underneath a few layers of `TypeSpec`s, with the top level being marked with the EXPECT modifier. When the `FunSpec` is more than two layers deep from the top-level type marked EXPECT, the generated method is being emitted with an empty body, rather than the expected declaration only.

A minimal example is the following:
```
TypeSpec.classBuilder("Outermost").addModifiers(KModifier.EXPECT)
    .addType(TypeSpec.classBuilder("Innermore")
        .addType(TypeSpec.classBuilder("Innermost")
            .addFunction(FunSpec.builder("foo").returns(ClassName("", "String")).build())
            .build())
        .build())
    .build()
```

Which generates:
```
expect class Outermost {
  class Innermore {
    class Innermost {
      fun foo(): String { // Compiler error due to implementation inside an `expect`
      }
    }
  }
}
```

The same example with one layer of TypeSpec removed:
```
TypeSpec.classBuilder("Outermost").addModifiers(KModifier.EXPECT)
    .addType(TypeSpec.classBuilder("Innermost")
        .addFunction(FunSpec.builder("foo").returns(ClassName("", "String")).build())
        .build())
    .build()
```

This does properly respect the outer EXPECT modifier, resulting in:
```
expect class Outermost {
  class Innermost {
    fun foo(): String
  }
}
```

JustinMullin:
It appears https://github.com/square/kotlinpoet/issues/732 addressed this case previously for single-nesting of types, but does not address further nesting.

ZacSweers:
Want to send a PR?

Egorand:
Closed with #879

PR #879, opened 2020-02-08
Propagate type implicit modifiers to constructors and nested types as well as functions.

JustinMullin:
This allows deeply nested classes and constructors to respect a top-level modifier.


JustinMullin:
Contains proposed fix for https://github.com/square/kotlinpoet/issues/878.

JustinMullin:
Whoops! Should be signed now.

PR #882, opened 2020-02-29
Add support for escaping $ in names

### LINKED ISSUE #841: char '$' in type name escape fail
`A$B` is a valid Java identifier, but not a valid Kotlin type name.

`escapeIfNotJavaIdentifier().escapeIfKeyword().failIfEscapeInvalid()`
none of these method respond to '$'.

`TypeSpec.classBuilder("A\$B").build()` build an invalid type name!
``TypeSpec.classBuilder("`A\$B`").build()`` also build an invalid type name!

ShaishavGandhi:
Resolves #841 

Technically `$` is allowed in names till it's escaped. 

```kotlin
class `A$B`
```

Most other special characters are covered by the Java identifiers.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -383,7 +383,7 @@ class ClassName internal constructor(
     }
     // concat top level class name and nested names
     return buildString {
-      append(topLevelClassName())
+      append(topLevelClassName().canonicalName)
       for (name in simpleNames.subList(1, simpleNames.size)) {
         append('$').append(name)
       }
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -199,7 +199,7 @@ private const val ALLOWED_CHARACTER = '$'
 
 internal val String.isKeyword get() = this in KEYWORDS
 
-internal val String.isAllowedCharacter get() = this.any { it == ALLOWED_CHARACTER }
+internal val String.hasAllowedCharacters get() = this.any { it == ALLOWED_CHARACTER }
 
 // https://github.com/JetBrains/kotlin/blob/master/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/checkers/JvmSimpleNameBacktickChecker.kt
 private val ILLEGAL_CHARACTERS_TO_ESCAPE = setOf('.', ';', '[', ']', '/', '<', '>', ':', '\\')
@@ -211,7 +211,7 @@ private fun String.failIfEscapeInvalid() {
 }
 
 internal fun String.escapeIfNecessary(validate: Boolean = true): String {
-  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfAllowedCharacter()
+  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfHasAllowedCharacters()
   if (validate) {
     escapedString.failIfEscapeInvalid()
   }
@@ -222,7 +222,7 @@ private fun String.alreadyEscaped() = startsWith("`") && endsWith("`")
 
 private fun String.escapeIfKeyword() = if (isKeyword && !alreadyEscaped()) "`$this`" else this
 
-private fun String.escapeIfAllowedCharacter() = if (isAllowedCharacter && !alreadyEscaped()) "`$this`" else this
+private fun String.escapeIfHasAllowedCharacters() = if (hasAllowedCharacters && !alreadyEscaped()) "`$this`" else this
 
 private fun String.escapeIfNotJavaIdentifier(): String {
   return if (!Character.isJavaIdentifierStart(first()) ||
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -195,13 +195,11 @@ private val KEYWORDS = setOf(
     "typeof"
 )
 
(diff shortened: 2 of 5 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -195,8 +195,14 @@ private val KEYWORDS = setOf(
     "typeof"
 )
 
+private val ALLOWED_CHARACTERS = setOf(
```
JakeWharton: Can we just use a const string here?

Egorand:
Hey @ZacSweers, here's the test that this change broke:

```kotlin
@Test
fun classNamesAndNesting_pathological() {
  // Make sure we parse class names correctly at all levels
  val typeSpec = `Fuzzy$ClassNesting`::class.toTypeSpecWithTestHandler()
  //language=kotlin
  assertThat(typeSpec.trimmedToString()).isEqualTo("""
    class Fuzzy${'$'}ClassNesting {
      class `-Nested` {
        class SuperNestedClass {
          inner class `-${'$'}Fuzzy${'$'}Super${'$'}Weird-Nested${'$'}Name`
        }
      }
    }
    """.trimIndent())
}
```

`Fuzzy$ClassNesting` is an incorrect class name and the fixture code won't compile. Should we just update the fixture, or is there something I'm missing?

ZacSweers:
yeah we should update the fixture to be properly escaped

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
ShaishavGandhi: Done!

ShaishavGandhi:
@Egorand Can you take a look again? Had to make some changes in the code to make the test pass correctly.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -383,7 +383,7 @@ class ClassName internal constructor(
     }
     // concat top level class name and nested names
     return buildString {
-      append(topLevelClassName())
+      append(topLevelClassName().canonicalName)
```
ShaishavGandhi: Without the `canonicalName` it would escape the topLevelClassName. Since this is used in reflection, escaping isn't required and would throw an exception

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -207,18 +211,22 @@ private fun String.failIfEscapeInvalid() {
...
-  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword()
+  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfAllowedCharacter()
   if (validate) {
     escapedString.failIfEscapeInvalid()
   }
   return escapedString
 }
 
-private fun String.escapeIfKeyword() = if (isKeyword) "`$this`" else this
+private fun String.alreadyEscaped() = startsWith("`") && endsWith("`")
```
ShaishavGandhi: Added this since we're chaining escapes like `escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfAllowedCharacter()`, there are chances of double escaping a string if two of the conditions meet. 

Happy to know if there's a more idiomatic way of representing this (i.e execute next statement if this predicate matches)

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -195,8 +195,12 @@ private val KEYWORDS = setOf(
     "typeof"
 )
 
+private const val ALLOWED_CHARACTER = '$'
+
 internal val String.isKeyword get() = this in KEYWORDS
 
+internal val String.isAllowedCharacter get() = this.any { it == ALLOWED_CHARACTER }
```
Egorand: How about
```suggestion
internal val String.hasAllowedCharacters get() = this.any { it == ALLOWED_CHARACTER }
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -207,18 +211,22 @@ private fun String.failIfEscapeInvalid() {
...
   }
   return escapedString
 }
 
-private fun String.escapeIfKeyword() = if (isKeyword) "`$this`" else this
+private fun String.alreadyEscaped() = startsWith("`") && endsWith("`")
+
+private fun String.escapeIfKeyword() = if (isKeyword && !alreadyEscaped()) "`$this`" else this
+
+private fun String.escapeIfAllowedCharacter() = if (isAllowedCharacter && !alreadyEscaped()) "`$this`" else this
```
Egorand:
```suggestion
private fun String.escapeIfHasAllowedCharacters() = if (hasAllowedCharacters && !alreadyEscaped()) "`$this`" else this
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -207,18 +211,22 @@ private fun String.failIfEscapeInvalid() {
 }
 
 internal fun String.escapeIfNecessary(validate: Boolean = true): String {
-  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword()
+  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfAllowedCharacter()
```
Egorand:
```suggestion
  val escapedString = escapeIfNotJavaIdentifier().escapeIfKeyword().escapeIfHasAllowedCharacters()
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
Egorand: I think it's fine, we could also just refactor the "escape" functions to return booleans, `||` them together and escape based on the value.

PR #885, opened 2020-03-01
Support generics on AnnotationSpecs

### LINKED ISSUE #881: Support type parameterized annotations
`AnnotationSpec` can only be created from a `ClassName`. Since Kotlin (as opposed to Java) supports type parameters on annotations, there should be a way to create an `AnnotationSpec` from a `ParameterizedTypeName`.

Example use case described here: https://stackoverflow.com/questions/60383694/cant-figure-out-how-to-generate-a-typeparcelert-p-annotation-with-kotlinpoe

Egorand:
Closes #881

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -30,7 +30,13 @@ class AnnotationSpec private constructor(
   builder: AnnotationSpec.Builder,
   private val tagMap: TagMap = builder.buildTagMap()
 ) : Taggable by tagMap {
-  val className: ClassName = builder.className
+  @Deprecated(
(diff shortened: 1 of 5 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -30,7 +30,13 @@ class AnnotationSpec private constructor(
   builder: AnnotationSpec.Builder,
   private val tagMap: TagMap = builder.buildTagMap()
 ) : Taggable by tagMap {
-  val className: ClassName = builder.className
+  @Deprecated(
```
ZacSweers: Error?

Egorand: I think warning is fine, there's no immediate need to change use sites upon upgrading, `className` would continue to work until you start creating parameterized `AnnotationSpec`s.

PR #888, opened 2020-03-06
Remove kotlin-reflect dependency

Egorand:


ZacSweers:
I think this will fail at runtime or it's still present on the test clssspath via something? Or it did when I tried doing this awhile back. Will dig up the branch

Egorand:
Added reflect as `testImplementation` dependency to a bunch of modules, will merge on success.

ZacSweers:
Is the goal to make it optional? Some APIs definer require it here, not sure how we signal that on a per API level

Egorand:
No real goal here, I was browsing through 1.3.70 release notes and noticed they've moved some API from `kotlin-reflect` into `kotlin-stdlib`, so I was curious if we can rely on stdlib only. Turns out `kotlinpoet` works fine without reflect, or am I missing something?

ZacSweers:
typeNameOf(), almost anything touching ktype, etc likely don't work

Egorand:
Ah, you're right, will revert. Sucks that this only fails at runtime.

PR #894, opened 2020-03-31
Validate Gradle wrapper on pushes to master

Egorand:
Aligns the behavior with SQLDelight: https://github.com/cashapp/sqldelight/blob/master/.github/workflows/gradle-wrapper-validation.yml


JakeWharton:
Don't like this. Prefer what we have.

Egorand:
Guess I'd agree, will send a PR to SQLDelight then.

PR #895, opened 2020-03-31
Align main Github Actions script with SQLDelight

Egorand:
Reference: https://github.com/cashapp/sqldelight/blob/master/.github/workflows/PR.yml

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -1,6 +1,10 @@
-name: CI
+name: Test
 
-on: [push, pull_request]
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/PR.yml
```diff
@@ -11,7 +15,9 @@ jobs:
         java_version: [1.8, 11]
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
```
JakeWharton, line 20: Aren't we running this twice then?

Egorand: Looks like it. Since a separate workflow would run the validation in parallel with the build, I'd say it's better to validate here to avoid calling the wrapper if the validation fails. WDYT?

ZacSweers: That's what I do in other projects, otherwise you're racing them and the consequential build may get farther before the other fails

Egorand: Agree. I like the modularity of being able to just add a file to the repo, makes contributing workflows much easier, but there should also be a way to define the ordering or the dependency graph between the workflows.

PR #896, opened 2020-03-31
Setup Release.yml workflow

Egorand:
Borrowed from https://github.com/cashapp/sqldelight/blob/master/.github/workflows/Release.yml

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,0 +1,23 @@
+name: Publish a release
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v2
+      - name: Publish the artifacts
+        env:
+          ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/Release.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v2
+      - name: Publish the artifacts
+        env:
+          ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME }}
+          ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD: ${{ secrets.ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD }}
```
JakeWharton, line 19: Did you initialize these in the repo?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,26 @@
...
+          # TODO(egorand): Figure out if the publishing plugin needs GPG keys and how they get fed
```
JakeWharton, line 20: You cannot publish without the artifacts being signed

.github/workflows/Release.yml
Egorand: Alec did

.github/workflows/Release.yml
Egorand: Yeah makes sense. I'll add the keys.

JakeWharton:
Obsolete now, right?

Egorand:
Yes, closing

PR #903, opened 2020-04-22
Fix references to publishing secrets

Egorand:


JakeWharton:
I'd rather normalize on the existing naming across our projects since it's agnostic to the build tool in use. I can update the secrets when I get to a computer.

Egorand:
@JakeWharton that works, can you close this PR once you've updated the secrets?

JakeWharton:
Yep

PR #906, opened 2020-04-25
Deploy website from CI

Egorand:


.github/workflows/build.yml
```diff
@@ -58,12 +58,10 @@ jobs:
...
-      #
-      # - name: Deploy docs
-      #   uses: mhausenblas/mkdocs-deploy-gh-pages@master
-      #   env:
-      #     PERSONAL_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
+      - name: Prep mkdocs
+        run: .buildscript/prepare_mkdocs.sh
+
+      - name: Deploy docs
+        uses: mhausenblas/mkdocs-deploy-gh-pages@master
```
JakeWharton, line 65: Should we do this only on tags? If I add a new feature and document it it'll show up on the website before it's released. The counterargument, of course, is that if you fix something you have to wait until the next release before it's deployed (or do it manually).

Egorand: Publishing on tags makes sense, although as Kotlinpoet is mostly feature-complete and we're mostly fixing things rather than adding new APIs, I think we'll have more instances of forgetting to publish the website than people complaining about unreleased APIs showing up in the docs. I'm happy with either though.

JakeWharton: Let's just merge as-is and we can change if needed.

Egorand:
Superseded by #990

PR #909, opened 2020-05-06
Don't emit own annotations in LambdaTypeName

### LINKED ISSUE #908: Potential bug when using lambdas, annotation & collections
Firstly, thank you for even creating this library! I'm very grateful as it makes generating code super easy to do! 

I was using this library to generate the following code
```
val map: Map<String, @CustomAnnotation()() -> Unit>
```
Here's the KotlinPoet code that I used

```
val customAnnotationClassName = ClassName("package_path", "Annotation")
val customAnnotation = AnnotationSpec.builder(customAnnotationClassName).build()
val predicate = LambdaTypeName.get(returnType = UNIT).copy(annotations = arrayListOf(customAnnotation))
val mapType = Map::class.asClassName().parameterizedBy(String::class.asTypeName(), predicate)
val mapProperty = PropertySpec.builder("map", mapType).build()
```

and here's the code that it generated

```
val map: Map<String, @CustomAnnotation() @CustomAnnotation() -> Unit>
```

There are a couple issues here:
- It's showing the name of the annotation twice
- Instead of the annotation showing up twice, I need the round brackets to show up twice`@CustomAnnotation()() -> Unit`

Is there a better way to generate this code?

Egorand:
Fixes #908 

Based on the current design, TypeNames should not emit their own
annotations. Instead, annotations should be emitted by CodeWriter
(as part of the %T printout logic) or enclosing TypeNames (such
as ParameterizedTypeName).

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
@@ -519,8 +519,6 @@ class LambdaTypeName private constructor(
   }
 
   override fun emit(out: CodeWriter): CodeWriter {
-    emitAnnotations(out)
-
     if (isNullable) {
       out.emit("(")
     }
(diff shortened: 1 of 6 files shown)


Egorand:
Good idea. @vinaygaba, since you've been generating code with annotated lambda types, do you have any additional test cases in mind? Please feel free to submit PRs.

vinaygaba:
@Egorand Wow thank you for taking care of this issue! I'll be able to add some more tests for this. Really appreciate the effort you'll put into this library 🙏

PR #910, opened 2020-05-07
Add support for Kotlin 1.4's functional interfaces

ZacSweers:
This adds a new `KModifier.FUN` and orders it alongside `enum` and `annotation` (figured this was the most conventional spot).

PR #912, opened 2020-05-09
Fun interface followups

ZacSweers:
Followups from #910

One thing I'm wondering is if we should make the function names match `isFunctionalInterface` or vice versa. Right now we have `funInterfaceBuilder()` (like the syntax) and `isFunctionalInterface()` (like the construct)


ZacSweers:
Gonna merge with the naming as-is

PR #916, opened 2020-05-24
Fixes #194 : FuncSpec.overriding now supports overriding suspend function

shivanshs9:
I haven't been able to support suspend vararg functions for now. But I've tested with regular suspend functions... maybe we could also have tests for overriding suspend functions?

Sorry, but I couldn't find any suitable alternative for the extension function I created, `TypeName.belongsToType`. If there already exists a better alternative, I'll switch over to it.


ZacSweers:
We're planning to deprecate this function in favor of pointing folks to using the metadata APIs, as it's wrong too often to be worth continuing to promote

PR #918, opened 2020-05-24
Open up FileSpec.Builder for more mutation

ZacSweers:
Similar towhat we do in other spec builders, this allows mutation of builder bits directly (so it's no longer just additive).

This covers:
- `comment`
- `members`
- `memberImports`. Specifically - this makes `Import` public but keeps its constructor `internal` and adds an adder. This allows reading Imports and re-adding them later if they're cleared in between, and seemed strictly better than just making the `memberImports` property itself public since we want folks to use the `addImport` functions.

### DIFF
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
@@ -770,11 +770,11 @@ class FileSpecTest {
 
     assertThat(builder.build().toString()).isEqualTo("""
       package com.taco
-      
+
       import com.foo.Foo
       import com.foo.Foo2
-      
-      
+
+
     """.trimIndent())
   }
 
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
@@ -754,6 +754,52 @@ class FileSpecTest {
     assertThat(builder.build().annotations).containsExactly(javaWord)
   }
 
+  @Test fun modifyImports() {
+    val builder = FileSpec.builder("com.taco", "Taco")
+        .addImport("com.foo", "Foo")
+
+    val currentImports = builder.imports
+    builder.clearImports()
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -754,6 +754,52 @@ class FileSpecTest {
...
+
+  @Test fun modifyMembers() {
+    val builder = FileSpec.builder("com.taco", "Taco")
+        .addFunction(FunSpec.builder("aFunction").build())
+        .addFunction(FunSpec.builder("aSecondFunction").build())
+        .addType(TypeSpec.classBuilder("AClass").build())
+
+    builder.members.removeAll { it !is TypeSpec }
+
+    check(builder.build().members.all { it is TypeSpec })
```
Egorand, line 789: Maybe `assertThat(builder.build().members).containsExactly(TypeSpec.classBuilder("AClass").build())`?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -754,6 +754,52 @@ class FileSpecTest {
...
+  }
+
+  @Test fun clearIndent() {
```
Egorand, line 792: nit: `clearComment`

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
ZacSweers: oops, I meant to actually add a typealias and property too to this test

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
ZacSweers: 755ccae

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
ZacSweers: 755ccae

PR #919, opened 2020-05-24
Deprecate mirror and element APIs

ZacSweers:
This deprecates existing mirror and element APIs, as they don't (and can't) return completely correct information about Kotlin types. This is just a first pass to soft-deprecate them and communicate that metadata should be used instead.

One thing not covered - `CodeBlock` accepts `TypeMirror` and `Element` as possible arguments. We could remove them, but it would be a behavior change. Thoughts?


ZacSweers:
This also doesn't cover reflective APIs. I think we might be able to make those work since we're using kotlin-reflect anyway

Egorand:
We can log a warning if someone uses `TypeMirror` or `Element` as args to `CodeBlock`.

Issue #927, opened 2020-06-04
Enable library mode in KotlinPoet


Egorand:
Also, track the issues we've encountered to have a list of things to fix in generated code.

Egorand:
Closed with #934

Issue #926, opened 2020-06-04
Incompatibility with library mode


JakeWharton:
KotlinPoet currently cannot be used with library mode since it omits optional visibility modifiers. I'm somewhat tempted to just always emit them rather than making this some kind of option. With an option, it means every consumer has to either detect library mode or expose a flag to turn it on which will probably never happen in practice. If we just emit library-mode-compatible code always, it'll always work regardless of the setting. We already do this with return types anyway.

Egorand:
Agree. Is missing visibility modifiers the only issue preventing emitted code to be used with library mode?

JakeWharton:
I think we also have to emit `Unit` which we currently omit. I've only been toying with it in side projects that don't use code gen so I'm not actually sure what (if anything) else is missing.

Egorand:
Should we enable library mode in KP itself and compile a list of issues we bump into?

JakeWharton:
Sounds like a good plan.

Egorand:
List of issues:

- Visibility must be specified explicitly
- Return type must be specified explicitly

ShaishavGandhi:
For library mode should the return type be specified only for public functions or internal/private as well?

JakeWharton:
I would probably just do everything for simplicity. Currently the only return type we omit is `Unit`, right?

ShaishavGandhi:
We also emit return types on expression bodies but not sure if we can do anything about that unless we do some type inference as discussed in #407.

From the tests:
```kotlin
@Test fun whenReturn() {
    val taco = TypeSpec.classBuilder("Taco")
        .addFunction(FunSpec.builder("toppingPrice")
            .addParameter("topping", String::class)
            .beginControlFlow("return when(topping)")
            .addStatement("%S -> 1", "beef")
            .addStatement("%S -> 2", "lettuce")
            .addStatement("%S -> 3", "cheese")
            .addStatement("else -> throw IllegalToppingException(topping)")
            .endControlFlow()
            .build())
        .build()
    assertThat(toString(taco)).isEqualTo("""
        |package com.squareup.tacos
        |
        |import kotlin.String
        |
        |class Taco {
        |  fun toppingPrice(topping: String) = when(topping) {
        |    "beef" -> 1
        |    "lettuce" -> 2
        |    "cheese" -> 3
        |    else -> throw IllegalToppingException(topping)
        |  }
        |}
        |""".trimMargin())
  }
```

Egorand:
[...]

PR #932, opened 2020-06-07
Emit Unit as return type

ShaishavGandhi:
Ref #927 #926 

This emits return types for FunSpec unless:
- It's a constructor
- Getter/Setter on a property 
- Expression body

### DIFF
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
@@ -146,7 +146,7 @@ class FunSpecTest {
             .build())
         .build()
     assertThat(funSpec.toString()).isEqualTo("""
-      |fun foo(string: kotlin.String?) {
+      |fun foo(string: kotlin.String?): kotlin.Unit {
       |}
       |""".trimMargin())
   }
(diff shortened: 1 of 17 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -690,7 +690,7 @@ class FunSpecTest {
 
     assertThat(builder.build().toString()).isEqualTo("""
       |@kotlin.jvm.JvmStatic
-      |internal fun staticMethod() {
+      |internal fun staticMethod(): kotlin.Unit {
```
Egorand, line 693: Does explicit API mode require `Unit` on non-public methods?

ShaishavGandhi: I don't think so but it's probably easier to just emit it always. Had a similar discussion in #926 with @JakeWharton

Egorand: Ah, cool, agree.

fluidsonic:
> Ref #927 #926
> 
> This emits return types for FunSpec unless:
> 
> * It's a constructor
> * Getter/Setter on a property
> * Expression body ⬅️

This seems to be backward.
The return type is **required** in library mode for functions with expression bodies and **not required** for regular bodies if it's `Unit`. According to your comment it was implemented the other way round.

Then again I was surprised that my code output now differs because I got a lot of useless `Unit` return types where they're clearly aren't needed: regular functions without expression body.

The IDE actually highlights that, even in library mode:
> Redundant 'Unit' return type

PR #933, opened 2020-06-08
Emit public modifier

ShaishavGandhi:
Ref #926 #927

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -150,13 +150,16 @@ internal class CodeWriter constructor(
 
   /**
    * Emits `modifiers` in the standard order. Modifiers in `implicitModifiers` will not
-   * be emitted.
+   * be emitted except for [KModifier.PUBLIC]
    */
   fun emitModifiers(
     modifiers: Set<KModifier>,
     implicitModifiers: Set<KModifier> = emptySet()
   ) {
-    if (modifiers.isEmpty()) return
+    if (shouldEmitPublicModifier(modifiers, implicitModifiers)) {
+      emit(KModifier.PUBLIC.keyword)
+      emit(" ")
+    }
     for (modifier in modifiers.toEnumSet()) {
       if (modifier in implicitModifiers) continue
       emit(modifier.keyword)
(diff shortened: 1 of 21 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -37,7 +37,7 @@ class TypeAliasSpec private constructor(
   internal fun emit(codeWriter: CodeWriter) {
     codeWriter.emitKdoc(kdoc.ensureEndsWithNewLine())
     codeWriter.emitAnnotations(annotations, false)
-    codeWriter.emitModifiers(modifiers)
+    codeWriter.emitModifiers(modifiers, setOf(PUBLIC))
```
ShaishavGandhi, line 40: TypeAliasSpec didn't have an implicit public modifier

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -566,6 +569,30 @@ internal class CodeWriter constructor(
...
   }
 
+  /**
+   * Returns whether a [KModifier.PUBLIC] should be emitted.
+   *
+   * This will return true when [KModifier.PUBLIC] is one of the [implicitModifiers] and
+   * there are no other opposing modifiers (like [KModifier.PROTECTED] etc.) supplied by the
+   * consumer.
+   */
+  private fun shouldEmitPublicModifier(
```
Egorand, line 579: Would it be possible to instead remove `PUBLIC` from every kind's implicit modifiers? That would hopefully eliminate the need for this code.

ShaishavGandhi: Yeah that was another approach I considered. We'd still have to do the check for clashing modifiers(if user entered PRIVATE as one of the modifiers) so this function would still exist, just with two less if statements. 

If that's fine, I can update the PR

Egorand: I realized there are actually two ways we can approach this:

- Emit `public` if there are no clashing modifiers
- Require `KModifier.PUBLIC` to be passed in explicitly for it to be emitted

I'd actually prefer the latter. Our current logic is that we will never emit `public`, even when it's passed in explicitly - since it's an implicit modifier. I'd want us to instead honor any visibility modifier if present, and we won't emit one if none are provided. What do you think?

PR #943, opened 2020-07-28
Do not merge properties with custom get or set (#942)

### LINKED ISSUE #942: Properties with custom getter are still "merged" into primary constructor properties
When a property in a class
- Is initialized with a _same name_ primary constructor parameter
- has a getter

It produces wrong code (tries to merge into a primary constructor property).

Miminal example:
```kotlin
TypeSpec.classBuilder("Greeter")
        .build()
    )
    .addProperty(PropertySpec.builder("name", String::class)
        .initializer("name")
        .getter(FunSpec.getterBuilder()
            .addStatement("return field")
            .build()
        )
        .build()
    ).build()
```
Which produces the output
```kotlin
class Greeter(
  val name: kotlin.Stringget() = field

)
```

Tested in version 1.6.0

enjoydambience:
Fixes #942


Egorand:
Thanks!

PR #957, opened 2020-07-31
Improve preconditions

GlassBricks:
- Checks for some specs were in inconsistent locations: in both `init` blocks of the Spec class and in the `build` function. There was even a redundant repeated check in FunSpec. I moved these checks into the build method. Since the Spec constructor is called from the build method, this shouldn't change behavior.
- Added some missing tests for FunSpec preconditions.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -41,18 +41,6 @@ public class PropertySpec private constructor(
   public val setter: FunSpec? = builder.setter
   public val receiverType: TypeName? = builder.receiverType
 
-  init {
-    require(typeVariables.none { it.isReified } ||
-        (getter != null || setter != null) &&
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -258,12 +246,21 @@ public class PropertySpec private constructor(
...
 
     public fun build(): PropertySpec {
-      if (KModifier.INLINE in modifiers) {
-        throw IllegalArgumentException("KotlinPoet doesn't allow setting the inline modifier on " +
-            "properties. You should mark either the getter, the setter, or both inline.")
+      require(KModifier.INLINE !in modifiers) {
+        "KotlinPoet doesn't allow setting the inline modifier on " +
+            "properties. You should mark either the getter, the setter, or both inline."
+      }
+      val getter = getter
```
Egorand, line 253: Not sure I get why local variables are needed here

Issue #939, opened 2020-07-09
add doc about `reified` keyword


z-950:
if want to create a function like this:
```kotlin
inline fun <reified T> foo()
```
the correct but not obvious code of kotlinpoet should be (1.6.0):
```kotlin
FunSpec.builder("foo")
  .addModifiers(KModifier.INLINE)
  .addTypeVariable(TypeVariableName("T").copy(reified = true))
  .build()
```

the wrong but obvious way is :
```kotlin
addTypeVariable(TypeVariableName("T", KModifier.REIFIED))
```
and will throw:
```
Execution failed for task ':example:kaptKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution
   > java.lang.reflect.InvocationTargetException (no error message)
```

it's hard to know `KModifier.REIFIED` use case. or it will never be used?

ZacSweers:
What's the full stacktrace? That snippet isn't enough context

Egorand:
The signature of the function you're trying to use is (see [`TypeVariableName` docs](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/-type-variable-name/)):

```kotlin
operator fun invoke(name: String, variance: KModifier? = null): TypeVariableName
``` 

So although it accepts a `KModifier` as the second argument, the subset of values that are allowed is limited to `KModifier.IN` and `KModifier.OUT`. Agree that it's unfortunate that there's no compile-time check to prevent you from misusing this API, but hopefully the parameter name should suggest that `KModifier.REIFIED` is not a supported argument.

z-950:
> The signature of the function you're trying to use is (see [`TypeVariableName` docs](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/-type-variable-name/)):
> 
> ```kotlin
> operator fun invoke(name: String, variance: KModifier? = null): TypeVariableName
> ```
> 
> So although it accepts a `KModifier` as the second argument, the subset of values that are allowed is limited to `KModifier.IN` and `KModifier.OUT`. Agree that it's unfortunate that there's no compile-time check to prevent you from misusing this API, but hopefully the parameter name should suggest that `KModifier.REIFIED` is not a supported argument.

 `reified`'s use case is the same as `in` and `out`. So it's strange that `KModifier.REIFIED` could not be use here

z-950:
[...]

PR #961, opened 2020-08-02
Use checkTarget on modifiers

GlassBricks:
This probably needs proofreading.

Other comments:
- Parameters can have some modifiers that properties can (that they otherwise shouldn't); this is to keep consistent behavior when parameters are turned into primary constructor properties.
  - A test case was found that doesn't produce compiling code (`final` does not apply to non-property parameter). This is fixed in a separate commit.
- In/Out (type variable) modifiers already have their own check.
- The Reified KModifier isn't used at all. See #939

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
@@ -124,11 +124,7 @@ public class TypeAliasSpec private constructor(
         addAnnotation(annotation.asClassName())
 
     public fun build(): TypeAliasSpec {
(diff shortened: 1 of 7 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeAliasSpec.kt
```diff
@@ -124,11 +124,7 @@ public class TypeAliasSpec private constructor(
         addAnnotation(annotation.asClassName())
 
     public fun build(): TypeAliasSpec {
-      for (it in modifiers) {
-        require(it in ALLOWABLE_MODIFIERS) {
```
JakeWharton, line 128: This variable now unused?

PR #963, opened 2020-08-09
Delete scripts

Egorand:


.buildscript/deploy_snapshot.sh
```diff
@@ -1,26 +0,0 @@
-#!/bin/bash
```
Egorand, line 1: This is Travis-specific and not needed anymore, `.github/workflows/build.yml` checks preconditions and `./gradlew uploadArchives` is enough to publish.

.buildscript/deploy_website.sh
```diff
@@ -1,41 +0,0 @@
-#!/bin/bash
```
Egorand, line 1: `./gradlew dokka && mkdocs gh-deploy` are enough to publish docs, no need to clone the repo.

.buildscript/prepare_mkdocs.sh
```diff
@@ -1,43 +0,0 @@
...
-# The website is built using MkDocs with the Material theme.
-# https://squidfunk.github.io/mkdocs-material/
-# It requires Python to run.
-# Install the packages with the following command:
-# pip install mkdocs mkdocs-material
-
-# Generate the API docs
-./gradlew dokka
-
-# Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
```
Egorand, line 12: I haven't found any symptoms of the problem described here, was it fixed in MkDocs at some point? The rest of the commands in this file aren't relevant anymore.

PR #970, opened 2020-08-23
Remove language version override

Egorand:


Egorand:
@ZacSweers looks like there are some behavior changes in metadata and reflection, I updated the tests to pass but haven't looked any deeper - any idea what we should do here?

Egorand:
IIUC there will be no changes to a consumer who's still using kotlinpoet-metadata API in a 1.3 codebase, since the output changes are most likely caused by bytecode- and metadata-level changes introduced in 1.4? If that's the case I think we should be good to merge.

ZacSweers:
Thing is these constant values are not stored in metadata, which adds to my confusion.

Egorand:
Are we filling them in?

ZacSweers:
Yeah we do 

https://github.com/square/kotlinpoet/blob/master/kotlinpoet-classinspector-elements/src/main/kotlin/com/squareup/kotlinpoet/classinspector/elements/ElementsClassInspector.kt#L197

https://github.com/square/kotlinpoet/blob/master/kotlinpoet-classinspector-reflective/src/main/kotlin/com/squareup/kotlinpoet/classinspector/reflective/ReflectiveClassInspector.kt#L223-L229

PR #977, opened 2020-08-31
prevent inserting new line at space in escaped keyword

### LINKED ISSUE #976: When calling a function that includes a space in the name, a new line is sometimes inserted in the middle of the function name.
For example, this code:

```kotlin
import com.squareup.kotlinpoet.*

val builder = StringBuilder()

@OptIn(ExperimentalStdlibApi::class)
FileSpec.builder("a", "b")
    .addFunction(FunSpec.builder("foo").apply {
        addParameter("aaa bbb", typeNameOf<(Int) -> String>())
        addStatement(
            StringBuilder().apply {
                repeat(10) {
                    append("`aaa bbb`($it) + ")
                }
                append("`aaa bbb`()")
            }.toString()
        )
    }.build())
    .build()
    .writeTo(builder)

println(builder.toString())

```

produces: 

```kotlin
package a

import kotlin.Function1
import kotlin.Int
import kotlin.String

fun foo(`aaa bbb`: Function1<Int, String>) {
  `aaa bbb`(0) + `aaa bbb`(1) + `aaa bbb`(2) + `aaa bbb`(3) + `aaa bbb`(4) + `aaa bbb`(5) + `aaa
      bbb`(6) + `aaa bbb`(7) + `aaa bbb`(8) + `aaa bbb`(9) + `aaa bbb`()
}
```

toliner:
Resolve #976


toliner:
Example Code:
```kotlin
import com.squareup.kotlinpoet.*

val builder = StringBuilder()

@OptIn(ExperimentalStdlibApi::class)
FileSpec.builder("a", "b")
    .addFunction(FunSpec.builder("foo").apply {
        addParameter("aaa bbb", typeNameOf<(Int) -> String>())
        val arg = mutableListOf<String>()
        addStatement(
            StringBuilder().apply {
                repeat(10) {
                    append("%N($it) + ")
                    arg += "aaa bbb"
                }
                append("%N(100)")
                arg += "aaa bbb"
            }.toString(),
            *arg.toTypedArray()
        )
    }.build())
    .build()
    .writeTo(builder)

println(builder.toString())
```

Output before PR:
```kotlin
package a

import kotlin.Function1
import kotlin.Int
import kotlin.String

fun foo(`aaa bbb`: Function1<Int, String>) {
  `aaa bbb`(0) + `aaa bbb`(1) + `aaa bbb`(2) + `aaa bbb`(3) + `aaa bbb`(4) + `aaa bbb`(5) + `aaa
      bbb`(6) + `aaa bbb`(7) + `aaa bbb`(8) + `aaa bbb`(9) + `aaa bbb`()
}
```
Output after PR:
```kotlin
package a

import kotlin.Function1
import kotlin.Int
import kotlin.String
import kotlin.Unit

public fun foo(`aaa bbb`: Function1<Int, String>): Unit {
  `aaa bbb`(0) + `aaa bbb`(1) + `aaa bbb`(2) + `aaa bbb`(3) + `aaa bbb`(4) + `aaa bbb`(5) +
      `aaa bbb`(6) + `aaa bbb`(7) + `aaa bbb`(8) + `aaa bbb`(9) + `aaa bbb`()
}
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -227,7 +227,7 @@ private fun String.escapeIfHasAllowedCharacters() = if (hasAllowedCharacters &&
 private fun String.escapeIfNotJavaIdentifier(): String {
   return if (!Character.isJavaIdentifierStart(first()) ||
       drop(1).any { !Character.isJavaIdentifierPart(it) } && !alreadyEscaped()) {
-    "`$this`"
+    "`$this`".replace(" ", "·")
```
JakeWharton, line 230: suuuuuuper nit:

```suggestion
    "`$this`".replace(' ', '·')
```

toliner:
I fixed the broken test. Should I add more tests?

JakeWharton:
I would prefer a test doing exactly like your example above which actually exercises the feature in practice. It's perfect.

toliner:
I added test using my sample code.

1. Dependency on `kotlin-scripting-jvm` and `kotlin-scripting-jvm-host` is added to :kotlinpoet.test
1. New module :test-util is added

To check generated code, I used kotlin-scripting. It is official, embeddable way to compile kotlin.
According to the specification of kotlin-script, a specific class must be defined in a separate module. To do it, I added :test-util module.

Egorand:
There's no need for all of the extra setup, you can `toString` a `FileSpec` and compare it to the hardcoded expected output - see how other tests in the project work.

Egorand:
@toliner just checking in: I'd love to merge this but would need the tests to be changed to be more in line with the other tests in the project.

toliner:
I'm so sorry for my late work. I added the test that compare generated code with hard-coded code.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
```diff
@@ -89,11 +89,47 @@ class UtilTest {
...
 
+  @OptIn(ExperimentalStdlibApi::class)
   @Test fun escapeNonJavaIdentifiers() {
     assertThat("8startWithNumber".escapeIfNecessary()).isEqualTo("`8startWithNumber`")
     assertThat("with-hyphen".escapeIfNecessary()).isEqualTo("`with-hyphen`")
-    assertThat("with space".escapeIfNecessary()).isEqualTo("`with space`")
+    assertThat("with space".escapeIfNecessary()).isEqualTo("`with·space`")
     assertThat("with_unicode_punctuation\u2026".escapeIfNecessary()).isEqualTo("`with_unicode_punctuation\u2026`")
+
+    val generated = FileSpec.builder("a", "b")
```
Egorand, line 99: I suggest moving this into a separate test method under `FileSpecTest`, tests inside `UtilTest` should only be testing utilities.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/compile/ParameterWithSpaceCompileTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlin.script.experimental.api.ResultWithDiagnostics
+import kotlin.script.experimental.host.toScriptSource
+import kotlin.script.experimental.jvm.dependenciesFromCurrentContext
+import kotlin.script.experimental.jvm.jvm
+import kotlin.script.experimental.jvmhost.BasicJvmScriptingHost
+import kotlin.script.experimental.jvmhost.createJvmCompilationConfigurationFromTemplate
+import kotlin.test.assertTrue
+import org.junit.Test
+
+class ParameterWithSpaceCompileTest {
```
Egorand, line 16: We don't do compilation tests, so this is redundant - please delete

settings.gradle.kts
```diff
@@ -29,3 +29,4 @@ include(
     ":kotlinpoet-metadata-specs",
     ":kotlinpoet-metadata-specs-tests"
 )
+include("test-util")
```
Egorand, line 32:
```suggestion
```

test-util/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
```
Egorand, line 1: Please delete the `test-util` module

Egorand (changes requested):

Egorand:
@toliner thanks! I think the compilation test and the supporting setup you added is redundant, can you please remove it?

toliner:
Oh, sorry, I reverted the commit but it seems not work.
I added revert-commit.

Egorand:
No worries, we'll be squashing the commits anyway. There's a couple more issues to address, plus a failing spotless check - please fix that one as well. Thanks!

toliner:
In my local, the test is passed. It looks that the blank line cause problem, so I added code to ignore blank lines.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
```diff
@@ -89,11 +89,47 @@ class UtilTest {
     assertThat("".isIdentifier).isFalse()
   }
 
+  @OptIn(ExperimentalStdlibApi::class)
```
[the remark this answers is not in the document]
Egorand: The question still stands, is this annotation needed?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
Egorand: I would still like this to be in a separate test method in a different file.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
```diff
@@ -89,11 +89,48 @@ class UtilTest {
...
+            append("%N(100)")
+            arg += "aaa bbb"
+          }.toString(),
+          *arg.toTypedArray()
+        )
+      }.build())
+      .build()
+      .toString()
+      .lines()
+      .filter { it.isNotBlank() }
```
Egorand, line 118: What's reason for stripping non-blank lines from the output?

Egorand: Sorry, just noticed your comment. This is strange, all other tests in the project don't have that issue. Is it just the trailing empty line that's causing problems? I think stripping non-empty lines is fine as a last resort, but ideally this shouldn't be needed.

Egorand:
Thanks, we're getting there! A few more things before I can merge this:

1. `@OptIn(ExperimentalStdlibApi::class)` - is it needed? What for?
2. Please move the new test case into a separate method.
3. Please sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1).

kotlinpoet/src/test/java/com/squareup/kotlinpoet/UtilTest.kt
toliner: I read the error log again, and I found the error is not caused by test, but spotless.
Indent inserted by IntelliJ is reported as error by spotless :(
I run `gradlew spotlessKotlinApply` command, then it was fixed.
I deleted the code to remove empty lines.

toliner:
1. `@OptIn(ExperimentalStdlibApi::class)` is required to use `typeNameOf<(Int) -> String>()` at L103
2. Done in [7d69de0](https://github.com/square/kotlinpoet/pull/977/commits/7d69de0cb4a0671f00f6719475b83d2b65a229c8)
3. Ok, I signed and sent.

Egorand (approved):
Looking good now, thanks! Merging.

PR #994, opened 2020-10-08
Escape soft and modifier keywords

### LINKED ISSUE #991: Modified keywords not escape when generating enums.
For instance, KP will generate the following and this doesn't compile because of `open`.
```kt
enum class Sort {
  open,
  sort
}
```

Egorand:
Fixes #991 

Feels overly restrictive, but the only alternative would be to encode the compiler rules of allowing soft keywords in specific contexts.


Egorand:
Failure is due to ktlint, my guess is that it doesn't like trailing commas. Will see if there's a newer version, otherwise will remove the comma.

ZacSweers:
I'm uneasy about this to be honest. Could we scope this down at all to only specific problematic spots?

JakeWharton (approved):
I'm fine with this. Fix the problem, and then pull it back where we can. Better than playing whack-a-mole long-term.

Egorand:
> Could we scope this down at all to only specific problematic spots?

Since there's an unknown number of problematic spots (enum value with constructor arguments is just an example) I think we'll be in a better spot if we catch all potential problems. Agree this is not needed in a lot of constructs, but maybe we can figure out a way to allow-list constructs that only require escaping the hard keywords? Open to ideas, but gonna merge this for now as it fixes the bug.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -165,36 +165,87 @@ private val IDENTIFIER_REGEX =
 
 internal val String.isIdentifier get() = IDENTIFIER_REGEX.matches(this)
 
-// https://github.com/JetBrains/kotlin/search?q=KeywordStringsGenerated.java
+// https://kotlinlang.org/docs/reference/keyword-reference.html
```
oldergod, line 168: 👍

kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
```diff
@@ -165,36 +165,87 @@ private val IDENTIFIER_REGEX =
...
 private val KEYWORDS = setOf(
-  "package",
+  // Hard keywords
```
oldergod, line 170: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

oldergod (commented):
thank you

abakes:
@Egorand I was wondering if you could expand on what this does exactly? This seems to have broken one of my clients

Egorand:
@abakes we used to only escape hard keywords before this change, but the issue this PR fixes described a scenario where that approach was failing, hence the decision to start escaping all Kotlin keywords. Wanna open a new issue and describe what broke in your code? A runnable test case would be ideal.

PR #995, opened 2020-10-08
ktlint 0.39.0

Egorand:


JakeWharton (approved):
There's a way to fix the import order so it doesn't use the nonsensical default, but it can be a follow-up.

PR #996, opened 2020-10-09
Use pull_request_target

ZacSweers:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request]
+on: [push, pull_request_target]
 
 env:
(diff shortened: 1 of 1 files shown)


ZacSweers:
More details https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

Egorand (approved):
Nice, thanks!

Egorand:
@ZacSweers I think this should be reverted, re-reading https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/, it explicitly states that:

> instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request

`pull_request_target` is probably a bad fit for actions that actually build the SHA, since it would just build the base SHA which is unhelpful, but probably a good fit for other types of actions.

PR #1000, opened 2020-10-15
Set source and target compatibility for KotlinCompile tasks

### LINKED ISSUE #999: Cannot use KotlinPoet version `1.7.0` with Java 8
The latest `1.7.0` version declares `"org.gradle.jvm.version": 11` which makes it impossible to compile for Java 8:

```
      > No matching variant of com.squareup:kotlinpoet:1.7.0 was found. The consumer was configured to find an API of a library compatible with Java 8, preferably in the form of class files, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm' but:
          - Variant 'apiElements' capability com.squareup:kotlinpoet:1.7.0 declares an API of a library, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm':
              - Incompatible because this component declares a component compatible with Java 11 and the consumer needed a component compatible with Java 8
          - Variant 'runtimeElements' capability com.squareup:kotlinpoet:1.7.0 declares a runtime of a library, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm':
              - Incompatible because this component declares a component compatible with Java 11 and the consumer needed a component compatible with Java 8
```

Egorand:
Fixes #999

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -44,6 +44,11 @@ subprojects {
       freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
     }
   }
+  // Ensure "org.gradle.jvm.version" is set to "8" in Gradle metadata.
+  tasks.withType<JavaCompile> {
+    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
+    targetCompatibility = JavaVersion.VERSION_1_8.toString()
+  }
   pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
     configure<KotlinProjectExtension> {
       explicitApi()


JakeWharton:
I don't think this affects the metadata, or at least i'm not sure. I would think that you would have to set it on the java convention / extension.

Egorand:
I'll test it some time later.

Egorand:
@JakeWharton PTAL, verified this locally.

build.gradle.kts
```diff
@@ -44,6 +44,11 @@ subprojects {
       freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
     }
   }
+  // Ensure "org.gradle.jvm.version" is set to "8" in Gradle metadata.
+  tasks.withType<JavaCompile> {
+    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
```
JakeWharton, line 49: I'm not sure you need `toString()`, but it doesn't really matter.

JakeWharton (approved):

build.gradle.kts
Egorand: With Kotlin DSL you do, sadly. We need setter overloads.

PR #1003, opened 2020-10-18
Omit visibility modifiers on custom accessors

### LINKED ISSUE #1002: Accessors get public visibility on non-public properties
Test case:

```kotlin
val file = FileSpec.builder("com.squareup.tacos", "Taco")
    .addProperty(PropertySpec.builder("foo", String::class, PRIVATE)
        .mutable()
        .getter(FunSpec.getterBuilder()
            .addStatement("return %S", "foo")
            .build())
        .setter(FunSpec.setterBuilder()
            .addParameter("foo", String::class)
            .build())
        .build())
    .build()
```

Prints:

```kotlin
package com.squareup.tacos
    
import kotlin.String
    
private var foo: String
  public get() = "foo"
  public set(foo) {
  }
```

Which doesn't compile (`Getter visibility must be the same as property visibility`).

Properties should either propagate their visibility to the accessors, or check that the accessors have visibility specified explicitly. My preference is for the former approach.

Egorand:
Fixes #1002

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/KModifier.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/KModifier.kt
@@ -15,6 +15,12 @@
  */
 package com.squareup.kotlinpoet
 
+import com.squareup.kotlinpoet.KModifier.INTERNAL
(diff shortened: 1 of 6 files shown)


oldergod (approved):
Nice

PR #1007, opened 2020-10-19
Fix tests and formatting

Egorand:
Really weird that CI didn't catch this in the PR.


Egorand:
I think something broke with the CI setup, here's what it complains about:
```
> The following files had format violations:
      kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
          @@ -221,7 +221,8 @@
           ······return·asReturnExpressionBody
           ····}
           ····if·(codeBlock.withoutPrefix(THROW_EXPRESSION_BODY_PREFIX_SPACE)·!=·null·||
          -······codeBlock.withoutPrefix(THROW_EXPRESSION_BODY_PREFIX_NBSP)·!=·null)·{
          +······codeBlock.withoutPrefix(THROW_EXPRESSION_BODY_PREFIX_NBSP)·!=·null
          +····)·{
           ······return·codeBlock
           ····}
           ····return·null
  Run 'gradlew spotlessApply' to fix these violations.
```
My PR fixes this issue exactly, it feels like CI is building the target commit instead of the commit in the PR.

ZacSweers (approved):

PR #1004, opened 2020-10-18
Detect expression bodies on 'return·' or 'throw·' prefix

ark-1:
Expression bodies were not recognized when the code starts with "return·"
I was surprised, because naturally, I used the "·" character after "return" keyword instead of space, because the line cannot be broken in this location. Before my changes, only space should be used.

PR #1008, opened 2020-10-19
Revert "Use pull_request_target"

Egorand:
This reverts commit ef5fc849

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request_target]
+on: [push, pull_request]
 
 env:
   GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"


Egorand:
CI failures are due to broken master.

ZacSweers:
I think the right fix would have been to just gate the publish job on push events, not reverting this

Egorand:
The issue wasn't in the publish job but in the fact that `pull_request_target` builds the base branch instead of the PR branch.

ZacSweers:
I don't think that's quite right. My understanding is it uses the base branch workflow to build the PR. Prevents people from coming in and adding "echo this secret to my email" to the workflow and having it run on your repo

Egorand:
Here's what happened:

- We received a PR (#1004) that had formatting issues. The CI passed and I merged the PR, which immediately broke master.
- I submitted a PR to fix master (#1007), verifying locally that it passes all tests - and the CI failed on it. Once I merged this PR and rebased, #1007 went green.

Would be awesome if you could investigate further, for now reverting the change at least unblocks further development. Another thought I had is that we can drop`pull_request` altogether since it more or less duplicates `push`.

PR #1009, opened 2020-10-29
Fix compiler args overwriting

LouisCAD:
See this issue: https://youtrack.jetbrains.com/issue/KT-41985

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -41,7 +41,7 @@ subprojects {
   tasks.withType<KotlinCompile> {
     kotlinOptions {
       jvmTarget = "1.8"
-      freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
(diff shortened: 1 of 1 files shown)


ZacSweers (approved):

Egorand:
Thanks for the fix @LouisCAD, appreciate it! Does it only affect KotlinPoet contributors, or it has potential to propagate to KotlinPoet consumers? Do we need a release for this fix?

JakeWharton:
It affects nothing locally and does not affect consumers. It just defends against copy/pasta.

LouisCAD:
@Egorand I think at the current state of KotlinPoet, it doesn't affect the project, but if you had to add any other compiler arguments explicitly, or implicitly (e.g. via a call to `explicitApi()`), these could be overridden, which would cause very hard to diagnose issues.

If there's arguments that affect the publication, it could/would affect consumers as well, but I can't find such an example from the top of my head.

PR #1013, opened 2020-11-01
Emit abstract modifier for abstract interface functions

### LINKED ISSUE #1011: Explicitly abstract properties in interface emit 'abstract' whereas explictly abstract functions do not

Egorand:
Fixes #1011


JakeWharton (approved):
Thanks!

Issue #1031, opened 2021-01-05
Type name clash detection does not work if clashing type is only used as superclass


luqasn:
It was a bit hard to describe this accurately, so please have a look at this example:

```
import com.squareup.kotlinpoet.ClassName
import com.squareup.kotlinpoet.FileSpec
import com.squareup.kotlinpoet.FunSpec
import com.squareup.kotlinpoet.TypeSpec

val namespace = "test"

val kotlinExceptionName = ClassName("kotlin", "Exception")
val customExceptionName = ClassName(namespace, "Exception")
val customException = TypeSpec
        .classBuilder("Exception")
        .superclass(kotlinExceptionName)
        .addFunction(
                FunSpec
                        .builder("test")
                        .addParameter("e", customExceptionName)
                        .build()
        )
        .build()


FileSpec.builder(namespace, "a.kt")
        .addType(customException)
        .build()
        .toString()

val funn = FunSpec.builder("test")
        .addParameter("e", customExceptionName)
        .addParameter("f", kotlinExceptionName)
        .build()

FileSpec.builder(namespace, "b.kt")
        .addType(customException)
        .addFunction(funn)
        .build()
        .toString()
```

this outputs
```
package test

import kotlin.Exception

class Exception : Exception() {
  fun test(e: Exception) {
  }
}
```

and

```
package test

class Exception : kotlin.Exception() {
  fun test(e: Exception) {
  }
}

fun test(e: Exception, f: kotlin.Exception) {
}
```

So in the second case where the clashing names are also used as parameters, the disambiguation works correctly, even for the inheritance. But in the first case, where the superclass should also be fully qualified, it does not work.

Egorand:
Looks like a bug - feel free to send us a PR!

PR #1032, opened 2021-01-07
Fix class name collision

luqasn:
Fixes https://github.com/square/kotlinpoet/issues/1031

- [x] CLA signed


Egorand (approved):
Awesome, thank you!

PR #1040, opened 2021-02-25
fix overeager auto-import leading to uncompilable code due to clash with class function name

luqasn:
If I try to reference `kotlin.error` inside a class function to throw an exception, but my class also has a method called `error`, the auto-importing of `kotlin.error` and subsequent shortening of the emitted MemberName to `error` leads to a clash with the class function `error` (with the class member winning over the imported member).

See test for a clearer picture.

I fixed it by introducing tracking of emitted `FunSpec` names for `CodeBlock`s and subsequently in the `CodeWriter`.
This is not an ideal solution because in order to really fix it properly, the CodeWriter would need to have knowledge of scopes, which it does not have currently.
The way I fixed it would mean that a local function (not sure they are even supported?) would lead to blocking a auto import even though there might not be a clash.
But I was not able to come up with a better solution without changing how everything works right now.

Looking forward to your feedback!

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -448,12 +448,25 @@ internal class CodeWriter constructor(
       return memberName.simpleName
     }
 
-    // We'll have to use the fully-qualified name. Mark the member as importable for a future pass.
-    if (!kdoc) {
+    // We'll have to use the fully-qualified name.
+    // Mark the member as importable for a future pass unless the name clashes with
+    // a method in the current context
+    if (!kdoc && !isMethodNameUsedInCurrentContext(memberName.simpleName)) {
       importableMember(memberName)
     }
 
     return memberName.canonicalName
+  }
+
+  // TODO(luqasn): also honor superclass members when resolving names.
+  private fun isMethodNameUsedInCurrentContext(simpleName: String): Boolean {
+    for (it in typeSpecStack.reversed()) {
+      if (it.funSpecs.any { it.name == simpleName })
+        return true
+      if (!it.modifiers.contains(KModifier.INNER))
+        break
+    }
+    return false
   }
 
   private fun importableType(className: ClassName) {
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
@@ -237,6 +237,92 @@ class MemberNameTest {
     )
   }
 
+  @Test fun importedMemberAndSuperClassFunctionNameClashForInnerClass() {
+    val kotlinErrorMember = MemberName("kotlin", "error")
+    val file = FileSpec.builder("com.squareup.tacos", "TacoTest")
+      .addType(
+        TypeSpec.classBuilder("Test")
+          .addFunction(
+            FunSpec
+              .builder("error")
+              .build()
+          )
+          .addType(
+            TypeSpec.classBuilder("TacoTest")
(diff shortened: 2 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -58,7 +58,8 @@ import kotlin.reflect.KClass
  */
 public class CodeBlock private constructor(
   internal val formatParts: List<String>,
-  internal val args: List<Any?>
+  internal val args: List<Any?>,
+  internal val emittedFunNames: List<String> = emptyList()
```
Egorand, line 62: `CodeBlock` is at a lower level of abstraction than `FunSpec`, so I think it's incorrect to have `CodeBlock` know about function names.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -201,6 +201,42 @@ class MemberNameTest {
     )
   }
 
+  @Test fun importedMemberAndClassFunctionNameClash() {
+    val kotlinErrorMember = MemberName("kotlin", "error")
```
Egorand, line 205: I believe the issue you're facing is specific to `MemberName` and seems to not affect `TypeName`, so it might be worth digging into the differences of how imports work for both and fixing the behavior for `MemberName`. I tweaked your test a bit to use `ClassName` instead of `MemberName` and got it to pass with what looks like correct output:

```kotlin
@Test fun importedMemberAndClassFunctionNameClash() {
  val kotlinErrorType = ClassName("kotlin", "Error")
  val file = FileSpec.builder("com.squareup.tacos", "TacoTest")
    .addType(
      TypeSpec.classBuilder("TacoTest")
        .addFunction(
          FunSpec.builder("test")
            .addStatement("%T(%S)", kotlinErrorType, "errorText")
            .build()
        )
        .addType(
          TypeSpec
            .classBuilder("Error")
            .build()
        )
        .build()
    )
    .build()
  assertThat(file.toString()).isEqualTo(
    """
    |package com.squareup.tacos
    |
    |import kotlin.Unit
    |
    |public class TacoTest {
    |  public fun test(): Unit {
    |    kotlin.Error("errorText")
    |  }
    |
    |  public class Error
    |}
    |""".trimMargin()
  )
}
```

luqasn:
You were totally right, I was approaching it from the wrong angle. I now looked at class name resolution and realized that I do have the current type stack to work with. My new approach has the same limitations as the class name resolution (does not check supertypes), but it works for this simple case.

Please check the updated commit.

luqasn:
I also found out that the compiler is intelligent enough that imported members only clash with functions if their signatures match, but checking for that is probably too involved.

Egorand (approved):
Looks awesome, and thanks for addressing the feedback!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -448,14 +448,27 @@ internal class CodeWriter constructor(
       return memberName.simpleName
     }
 
-    // We'll have to use the fully-qualified name. Mark the member as importable for a future pass.
-    if (!kdoc) {
+    // We'll have to use the fully-qualified name.
+    // Mark the member as importable for a future pass unless the name clashes with
+    // a method in the current context
```
JakeWharton, line 453: nit: missing period.

[Comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).

luqasn: Noted for next time :)

PR #1041, opened 2021-02-28
Restructure interop artifactst starting with kotlinx-metadata

ZacSweers:
This restructures the existing metadata artifacts into a new organized `interop` directory with this s structure. This will also allow us to easily add new interop mechanisms (javapoet, KSP) in the future.

New layout:
```
:kotlinpoet
:interop
  :kotlinx-metadata
    :core
    :specs
    :specs-tests
    :classinspectors
      :elements
      :reflect
```


ZacSweers:
We could also look at updating the artifact coordinates with this if we want. Something like `com.squareup.kotlinpoet:interop-kotlinx-metadata-core`

Egorand (approved):
Yep, I like this.

I recall we discussed merging metadata stuff back into the main artifact as current layout felt a bit too overengineered, how would that have looked with the additions of Javapoet and KSP interop?

Egorand:
> We could also look at updating the artifact coordinates with this if we want. Something like com.squareup.kotlinpoet:interop-kotlinx-metadata-core

Yeah, sounds reasonable. Will need to think about how to this in the least disruptive way.

ZacSweers:
I think the least disruptive way would just be to change the coordinates and say updaters need to use the new artifacts.

Re: consolidating - I think it makes sense to consolidate the metadata artifacts into one, but still keep the standalone single artifact. What do you think?

Egorand:
Agree

ZacSweers:
Let me consolidate in a separate PR

PR #1043, opened 2021-03-13
Refactor logic to determine if function body should be omitted

luqasn:
while looking at some related code, I had a hard time understanding the current logic and extracted stuff into (hopefully) self-describing methods.

This also fixes what I think is a bug: `expect` functions would omit their body even if it was not empty. This logic 'flaw' was also present for `abstract` functions, but this code path was never hit due to validation in the constructor of `FunSpec`

Included is also another drive-by: implicitly abstract functions can have their empty body omitted as well now (this is currently not used, but I plan on doing a follow-up PR).


Egorand (approved):
Thanks!

PR #1045, opened 2021-03-19
Make metadata .toMutable() methods completely lossless (fixes #1044)

### LINKED ISSUE #1044: ImmutableKmProperty.toMutable() doesn't copy fieldSignature
Hi, I've noticed that while `KmProperty.toImmutable()` does copy `fieldSignature`, `ImmutableKmProperty.toMutable()` does not.

Sample code:

```kotlin
fun main() {
    val original = KmProperty(0, "test", 0, 0).apply {
        // Required fields
        returnType = KmType(0).apply { classifier = KmClassifier.Class("test/Test") }

        // Set fieldSignature
        fieldSignature = JvmFieldSignature("test", "Ltest/Test;")
    }
    val immutable = original.toImmutable()

    println(original.fieldSignature == immutable.fieldSignature)  // true

    println(original.fieldSignature == immutable.toMutable().fieldSignature)  // false
}
```

Is this intended? At the moment I'm using a wrapper that copies the field in the toMutable result, but it would be nice to have lossless conversions between mutable/immutable.

forsenonlhaimaisentito:
This contribution fixes the weird behavior documented in #1044 for all immutable metadata classes.

For some reason, although JVM extension fields were copied in `toImmutable()`, they weren't in `toMutable()`, causing the following condition to fail:

```kotlin
original.toImmutable().toMutable() == original
```


ZacSweers (approved):
Thanks!

PR #1049, opened 2021-03-20
Update kotlinx-metadata to 0.2.0

ZacSweers:
Notable improvements:
* Support `fun interface`
* Fix primary constructor detection


ZacSweers:
There's a 0.2.1 in the pipeline with full val class support, but not there yet

Egorand (approved):

PR #1050, opened 2021-03-20
Add KModifier.VALUE

ZacSweers:
For the upcoming language support for value classes


kotlinpoet/src/main/java/com/squareup/kotlinpoet/KModifier.kt
```diff
@@ -54,6 +54,7 @@ public enum class KModifier(
 
   ENUM("enum", Target.CLASS),
   ANNOTATION("annotation", Target.CLASS),
+  VAL("val", Target.CLASS),
```
Egorand, line 57: Isn't it `value`? https://kotlinlang.org/docs/inline-classes.html

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -462,7 +464,8 @@ public class TypeSpec private constructor(
     internal val isEnum get() = kind == Kind.CLASS && ENUM in modifiers
     internal val isAnnotation get() = kind == Kind.CLASS && ANNOTATION in modifiers
     internal val isCompanion get() = kind == Kind.OBJECT && COMPANION in modifiers
-    internal val isInlineClass get() = kind == Kind.CLASS && INLINE in modifiers
+    internal val isInlineOrValClass get() = kind == Kind.CLASS &&
```
Egorand, line 467: IIRC they're planning to remove `inline` at some point, but I guess we can rename all occurrences of `inlineOrValue` to `value` then.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/ValTypeSpecTest.kt
```diff
@@ -0,0 +1,236 @@
...
+        arrayOf(true),
+        arrayOf(false)
+      )
+    }
+  }
+
+  private val modifier = if (useVal) KModifier.VAL else INLINE
+  private val modifierString = modifier.keyword
+
+  private fun classBuilder() = if (useVal) {
```
Egorand, line 27: Guess this is more consistent (than `.classBuilder().addModifiers(INLINE)`) with things like `enumBuilder` or `annotationBuilder`.

Egorand (approved):

kotlinpoet/src/main/java/com/squareup/kotlinpoet/KModifier.kt
ZacSweers: D'oh, good catch

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
ZacSweers: Yeah I thought about that. Since it's internal we can always change later, i felt like this was more clear for now. Thoughts?

Egorand: Agree!

ZacSweers:
@egor-n want to get your final sign-off on the `value` keyword escaping just in case

Egorand (approved):
Looks good!

PR #1057, opened 2021-03-29
Omit redundant abstract modifiers on members where possible

ZacSweers:
Partially revert #1013 in favor of moving the other direction


Egorand (approved):
Yeah, this is much better, thanks!

ZacSweers:
rekicked CI as it seems there's some 500s coming from the setup java url

PR #1061, opened 2021-04-02
useIR = true

Egorand:


JakeWharton:
I would not do this, or at least only do it on a GitHub Action that we test with (but not publish). Consumers not using IR will be required to add a special flag to use an IR-based dependency and there is still potential for incompatibility.

Egorand:
Oh, got it, wasn't aware this propagates to consumers. Yeah, can definitely just add it during the CI run.

Egorand:
Closing as obsolete since we've bumped to Kotlin 1.5, which AFAIK makes IR the default

PR #1075, opened 2021-05-02
Adding ability to change initializers and defaultValues after being set

### LINKED ISSUE #1074: Enable method for overriding currently one-set values (initializer, getter, setter, etc)
I'm working on a project to generate Kotlin data classes from JsonSchema definitions. I'm providing fairly lightweight generation of the classes (not even bundling in anything Moshi/Gson/Jackson implementations in the core library) but want to enable an extensible implementation. I was thinking of exposing TypeSpec/ParameterSpec/PropertySpec to Interceptor interfaces of mine so that users of the library could change out property names/class names, add annotations, etc for each. 

I ran into a bit of a speed bump in that once some of the parts of some of these specs are called, they can't be overidden, even with a `.toBuilder()` call after being built. For example, I can't replace the `initializer` on a property, as I'll hit [an exception](https://github.com/square/kotlinpoet/blob/master/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt#L235) about the initializer having already been set.

I realize I could certainly not expose KotlinPoet as the interface for the Interceptors, but I am lazy, and I would much rather not have to rewrite a bunch of functionality that's meant to line up to the codegen anyway.

I'm curious if you'd be open to changes that _at least_ enable this behavior, such as adding `allowOverride = false` to these functions such that they only validate the already-set case when `allowOverride` is `false`?

Interestingly, some parts of the API, such as `TypeSpec.primaryConstructor(...)` don't prevent setting multiple times. I could remove the set-once restriction in places they already exist, but I'm afraid that might break current implementations that may depend on that set-once exception throwing.

dallasgutauckis:
Adds support for 
 - clearing the initializer in PropertySpec
 - clearing the defaultValue in ParameterSpec

Closes #1074

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -143,6 +143,10 @@ public class ParameterSpec private constructor(
       this.defaultValue = codeBlock
     }
 
+    public fun clearDefaultValue(): Builder = apply {
+      this.defaultValue = null
+    }
+
     public fun build(): ParameterSpec = ParameterSpec(this)
   }
 
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterSpec.kt
@@ -138,13 +138,8 @@ public class ParameterSpec private constructor(
     public fun defaultValue(format: String, vararg args: Any?): Builder =
       defaultValue(CodeBlock.of(format, *args))
 
-    public fun defaultValue(codeBlock: CodeBlock): Builder = apply {
-      check(this.defaultValue == null) { "initializer was already set" }
+    public fun defaultValue(codeBlock: CodeBlock?): Builder = apply {
       this.defaultValue = codeBlock
-    }
-
-    public fun clearDefaultValue(): Builder = apply {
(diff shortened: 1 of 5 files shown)


Egorand (approved):
I'm okay with this solution but would also not be opposed to just removing the checks for double initialization. Not too worried about changing the behavior since it's undocumented. The checks were probably added to prevent accidentally setting those values multiple times, which IMO is not a very strong reason, and pretty sure we don't apply this approach consistently across the API.

dallasgutauckis:
@Egorand appreciate the feedback. I'm fine going that route, but if we remove the checks we'd probably also want to make the parameter on at least the CodeBlock forms nullable:

i.e.

```kotlin
public fun defaultValue(codeBlock: CodeBlock): Builder = apply {
      check(this.defaultValue == null) { "initializer was already set" }
      this.defaultValue = codeBlock
    }
```

to become

```
public fun defaultValue(codeBlock: CodeBlock?): Builder = apply {
      this.defaultValue = codeBlock
    }
```

so that it's possible to clear the value from the Spec.

If you're okay with that, I'll make these changes :-)

dallasgutauckis:
@Egorand assuming you're okay with this, I've made the suggested changes to:
 - Make the parameters nullable to support clearing
 - Make the various methods (defaultValue, initializer, etc) callable more than once and updating with the latest value

JakeWharton (approved):

Egorand (approved):
Yep, looks great, thanks @dallasgutauckis!

PR #1088, opened 2021-05-24
Proposal PR: Add DelicateKotlinPoetApi

ZacSweers:
Proposal implementation for https://github.com/square/kotlinpoet/discussions/1087

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -23,7 +23,6 @@ import javax.lang.model.element.TypeElement
 import javax.lang.model.element.VariableElement
 import javax.lang.model.type.TypeMirror
 import javax.lang.model.util.SimpleAnnotationValueVisitor7
-import kotlin.DeprecationLevel.WARNING
 import kotlin.reflect.KClass
 
 /** A generated annotation on a declaration. */
@@ -181,7 +180,13 @@ public class AnnotationSpec private constructor(
   }
 
   public companion object {
-    @JvmStatic @JvmOverloads public fun get(
+    @DelicateKotlinPoetApi(
+      message = "Java reflection APIs don't give complete information on Kotlin types. Consider " +
+        "using the kotlinpoet-metadata APIs instead."
+    )
+    @JvmStatic
+    @JvmOverloads
+    public fun get(
       annotation: Annotation,
       includeDefaultValues: Boolean = false
     ): AnnotationSpec {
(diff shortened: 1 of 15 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/DelicateKotlinPoetApi.kt
```diff
@@ -0,0 +1,17 @@
+package com.squareup.kotlinpoet
+
+/**
+ * Marks declarations in the KotlinPoet that are **delicate** &mdash;
```
ZacSweers, line 4:
```suggestion
 * Marks declarations in the KotlinPoet API that are **delicate** &mdash;
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -181,7 +180,13 @@ public class AnnotationSpec private constructor(
   }
 
   public companion object {
-    @JvmStatic @JvmOverloads public fun get(
+    @DelicateKotlinPoetApi(
+      message = "Java reflection APIs don't give complete information on Kotlin types. Consider " +
```
Egorand, line 184: Nice!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/DelicateKotlinPoetApi.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Marks declarations in the KotlinPoet API that are **delicate** &mdash;
+ * they have limited use-case and shall be used with care in general code.
+ * Any use of a delicate declaration has to be carefully reviewed to make sure it is
+ * properly used and does not create problems like lossy Java -> Kotlin type parsing.
+ * Carefully read documentation and [message] of any declaration marked as `DelicateKotlinPoetApi`.
+ */
+@MustBeDocumented
+@Retention(value = AnnotationRetention.BINARY)
```
Egorand, line 11: Do we need it to be binary?

Egorand (approved):
Makes sense to me, awesome!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/DelicateKotlinPoetApi.kt
ZacSweers: I believe so, otherwise it wouldn't be visible to consumers right?

ZacSweers: Merging as-is but can followup if needed!

Egorand: IIRC you only need an annotation in the bytecode if it's supposed to be queried at runtime through reflection, which isn't the case here, the annotation is only processed by the compiler

ZacSweers: I think you're thinking of RUNTIME retention for that case. BINARY still ensures it's in bytecode, it just can't be queried at runtime. SOURCE is like it sounds - only in source and omitted everywhere else

Egorand: True, sorry!

PR #1091, opened 2021-06-04
Add MemberName.isExtension and update importing logic

### LINKED ISSUE #1089: KotlinPoet wrongly inlines extension functions
KotlinPoet seems to be inlining extensions functions like if they were members and this generates non-compilable code.

Here's a diff of our before/after.

```diff
 import kotlin.Long
 import kotlin.String
 import kotlin.Unit
-import kotlin.hashCode
 import kotlin.jvm.JvmField
 import okio.ByteString

@@ -55,8 +54,8 @@ public class OuterMessage(
     var result = super.hashCode
     if (result == 0) {
       result = unknownFields.hashCode()
-      result = result * 37 + outer_number_before.hashCode()
-      result = result * 37 + embedded_message.hashCode()
+      result = result * 37 + outer_number_before.kotlin.hashCode()
+      result = result * 37 + embedded_message.kotlin.hashCode()
       super.hashCode = result
     }
     return result
```

It breaks Wire when we bump KotlinPoet from `1.7.2` to `1.8.0`.
I think that's a side-effect of https://github.com/square/kotlinpoet/pull/1040 ?

Egorand:
Fixes #1089

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -451,8 +451,11 @@ internal class CodeWriter constructor(
     // We'll have to use the fully-qualified name.
     // Mark the member as importable for a future pass unless the name clashes with
     // a method in the current context
-    if (!kdoc && (memberName.isExtension ||
-          !isMethodNameUsedInCurrentContext(memberName.simpleName))) {
+    if (!kdoc && (
+      memberName.isExtension ||
+        !isMethodNameUsedInCurrentContext(memberName.simpleName)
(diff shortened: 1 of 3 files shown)


oldergod (approved):
I wonder if people would wrongly assume `isExtension` will be set for them

Egorand:
@oldergod as in, thinking it's a computed property vs a constructor parameter? Any updates to the docs you can suggest that would make this more clear?

oldergod:
Is there a way to ever get a member not manually created? If no, then it might be ok.
If yes, a comment on the field would be good.

Egorand:
No, I don't think we're ever inferring `MemberName`s from some other type of input, we expect them to only be created externally.

PR #1092, opened 2021-06-06
Update dependencies

Goooler:
- [kotlin-compile-testing 1.4.2](https://github.com/tschuchortdev/kotlin-compile-testing/releases/tag/1.4.2) Updated kotlin-compiler-embeddable and KSP to 1.5.10
- [AutoCommon 1.0.1](https://github.com/google/auto/releases/tag/auto-common-1.0.1) Added some methods to allow annotation processors to use Streams functionality that is present in mainline Guava but not Android Guava. This can be useful if Android Guava might be on the processor path.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -15,7 +15,7 @@
  */
 object versions {
   const val kotlin = "1.5.10"
-  const val spotless = "5.12.5"
+  const val spotless = "5.13.0"
   const val ktlint = "0.41.0"
   const val mavenPublish = "0.15.1"
   const val dokka = "1.4.32"
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -17,12 +17,12 @@ object versions {
   const val kotlin = "1.5.10"
   const val spotless = "5.12.5"
   const val ktlint = "0.41.0"
-  const val mavenPublish = "0.13.0"
+  const val mavenPublish = "0.15.1"
   const val dokka = "1.4.32"
 }
 
 object deps {
-  const val autoCommon = "com.google.auto:auto-common:1.0"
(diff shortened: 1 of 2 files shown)


Egorand (approved):
In general, we're not striving to pick up every single minor version change, so I would advise against submitting this kind of PRs too often. If there are any specific dependency changes that you think would benefit this library, please mention this in the PR description.

Goooler:
> In general, we're not striving to pick up every single minor version change, so I would advise against submitting this kind of PRs too often. If there are any specific dependency changes that you think would benefit this library, please mention this in the PR description.

Ok

ZacSweers:
descriptions in general are preferable to show awareness of what's in an update. For example, this PR needs more work than what you've submitted because the maven publish update requires changing Gradle properties on our CI.

ZacSweers:
@Goooler can you update the relevant properties in our GitHub actions workflow for the maven publish update? Snapshots will be broken without it

Goooler:
> @Goooler can you update the relevant properties in our GitHub actions workflow for the maven publish update? Snapshots will be broken without it

Should be done now, have a try

Egorand:
Thanks @Goooler! Can we extract all the `mavenPublish`-related changes into a separate commit so it's easier to revert if need be? I suggest rebasing current commits into the following two commits:

```
1. Update dependencies (everything except for maven publish)
2. Update maven-publish-plugin to 0.15.1
```

Once done we should be ready to merge.

Goooler:
> Thanks @Goooler! Can we extract all the `mavenPublish`-related changes into a separate commit so it's easier to revert if need be? I suggest rebasing current commits into the following two commits:
> 
> ```
> 1. Update dependencies (everything except for maven publish)
> 2. Update maven-publish-plugin to 0.15.1
> ```
> 
> Once done we should be ready to merge.

Done

Egorand (approved):
Thanks!

PR #1095, opened 2021-06-15
Emit empty primary constructor

### LINKED ISSUE #1093: Empty primary constructor not generated properly
Hello! 

I am using Kotlinpoet v1.8.0 and I am trying to generate an empty primary constructor for a class `TypeSpec`.

```kotlin
builderSpec.primaryConstructor(FunSpec.constructorBuilder().build())
```

The expected result here would be something like:

```kotlin
class Foo() { ... }
```

whereas the actual result is:

```kotlin
class Foo { ... }
```

This breaks the functionality when we are trying to add a second constructor that uses `callThisConstructor()` since there is no primary empty constructor to delegate to.

pavlospt:
Fixes #1093

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -166,27 +166,24 @@ public class TypeSpec private constructor(
 
         primaryConstructor?.let {
           codeWriter.pushType(this) // avoid name collisions when emitting primary constructor
-          var useKeyword = false
-          var emittedAnnotations = false
+          val emittedAnnotations = it.annotations.isNotEmpty()
+          val useKeyword = it.annotations.isNotEmpty() || it.modifiers.isNotEmpty()
 
           if (it.annotations.isNotEmpty()) {
             codeWriter.emit(" ")
             codeWriter.emitAnnotations(it.annotations, true)
-            useKeyword = true
-            emittedAnnotations = true
           }
 
           if (it.modifiers.isNotEmpty()) {
             if (!emittedAnnotations) codeWriter.emit(" ")
             codeWriter.emitModifiers(it.modifiers)
-            useKeyword = true
           }
 
           if (useKeyword) {
             codeWriter.emit("constructor")
           }
 
-          it.parameters.emit(codeWriter, forceParensOnEmpty = useKeyword, forceNewLines = true) { param ->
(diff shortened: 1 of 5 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -186,17 +186,21 @@ public class TypeSpec private constructor(
...
+                  codeWriter, setOf(PUBLIC), withInitializer = false, inline = true,
+                  inlineAnnotations = false
+                )
+                param.emitDefaultValue(codeWriter)
+              } else {
+                param.emit(codeWriter, emitKdoc = true, inlineAnnotations = false)
+              }
             }
+          } else {
+            codeWriter.emit("()")
```
Egorand, line 203: `parameters.emit()` already encodes this behavior if `forceParensOnEmpty` is true, so I suggest using that. It also looks like this is the only call site that passes a value for `forceParensOnEmpty` so we can explore deleting that parameter altogether and just making the method always emit brackets.

pavlospt: Update with your comments @Egorand 😃

Egorand (approved):
Amazing, thanks!

Egorand:
@pavlospt ready to merge once spotless issues are fixed - just run `spotlessApply` and push!

pavlospt:
> @pavlospt ready to merge once spotless issues are fixed - just run `spotlessApply` and push!

Ready @Egorand !

Egorand:
@pavlospt looks like we need more test updates here

pavlospt:
> @pavlospt looks like we need more test updates here

Yeap I am on it!

pavlospt:
@Egorand hopefully fixed. Locally everything passed !

ZacSweers:
Could we make this a bit more intelligent? I.e. rather than always emit, can we only emit if any constructors have super calls?

pavlospt:
> Could we make this a bit more intelligent? I.e. rather than always emit, can we only emit if any constructors have super calls?

Since this code is currently emitted based on the constructor parameters I can not seem to find a quick and obvious solution on that. Maybe with a bigger refactoring this could be doable!

Egorand:
@ZacSweers we'll only emit when explicitly given a `primaryConstructor(FunSpec.constructorBuilder().build())`, which I think is correct - if the use site doesn't want the empty constructor it should omit passing the `FunSpec`.

PR #1106, opened 2021-07-02
Avoid escaping already escaped string

idanakav:
Fix by avoiding a short circuit and making sure we evaluate !alreadyEscaped

The added test would previously fail.


Egorand (approved):
Looks good, thanks!

PR #1111, opened 2021-07-16
Update dependencies

Goooler:
- [Gradle 7.1.1](https://docs.gradle.org/7.1.1/release-notes.html)
- [gradle-maven-publish-plugin 0.17.0](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md#version-0170-2021-07-04)
- [Auto common 1.1](https://github.com/google/auto/releases/tag/auto-common-1.1)
- ~~Add [dependabot](https://dependabot.com/) for automated dependency updates~~

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "21:00"
+    open-pull-requests-limit: 10
+    target-branch: master
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: gradle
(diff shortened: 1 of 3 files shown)


ZacSweers (commented):
Does dependabot work for the way our dependencies are managed?

Egorand:
TBH I find dependabot too noisy and I wouldn't want us to use it - I can be convinced otherwise if other maintainers strongly support it. This library only really depends on the Kotlin stdlib, which we're pretty diligent about keeping up to date anyway. Bumping all of the test dependencies on a regular basis doesn't feel very useful to me.

ZacSweers:
I agree, it's too limited in anything but the most trivial gradle projects too

Goooler:
> I agree, it's too limited in anything but the most trivial gradle projects too

Ok, I reverted.

Egorand (approved):
Thanks!

PR #1112, opened 2021-07-17
Add JDK 16 to build matrix

Egorand:


.github/workflows/build.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           - 13
           - 14
           - 15
+          - 16
```
ZacSweers, line 24:
```suggestion
          - 16
          - 17-ea
```

Go for gold?

ZacSweers (approved):

.github/workflows/build.yml
Egorand: Sure!

Egorand: Looks like Kotlin is not ready yet though

ZacSweers:
That's weird, I'm using this fine in other kotlin projects.

PR #1116, opened 2021-08-09
made kotlin-reflect a testImplementation dependency

martinbonnin:
`kotlin-reflect` doesn't seem to be required at runtime? I can run all tests locally without `kotlin-reflect`. Let me know if I'm missing something, I'm happy to add a specific test.

### DIFF
--- a/kotlinpoet/build.gradle.kts
+++ b/kotlinpoet/build.gradle.kts
@@ -34,7 +34,7 @@ tasks.withType<KotlinCompile>().named("compileTestKotlin") {
 }
 
 dependencies {
-  implementation(deps.kotlin.reflect)
+  testImplementation(deps.kotlin.reflect)
   testImplementation(deps.kotlin.junit)
   testImplementation(deps.test.truth)
   testImplementation(deps.test.compileTesting)


ZacSweers (commented):
The tests will all still pass because kotlin-reflect is a (test) dependency. Many (most?) would fail workout it. I'd be open to trying to remove this dependency or make it optional, but we'd have to document the APIs (or even move them to a separate artifact) first

martinbonnin:
> The tests will all still pass because kotlin-reflect is a (test) dependency

Arg, right, of course 👍 .

> I'd be open to trying to remove this dependency or make it optional, but we'd have to document the APIs

Looking at the tests, it looks like `kotlin-reflect` is mostly used at the call site (i.e. in the test code itself). For an [example](https://github.com/square/kotlinpoet/blob/e1a8cde615a10df2f3379cb6cf8766922cf1a7a1/kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt#L77) to get instances of `KType` from `returnType` or [using createType](https://github.com/square/kotlinpoet/blob/master/kotlinpoet/src/test/java/com/squareup/kotlinpoet/AbstractTypesTest.kt#L193) directly. 

The only place I could find in KotlinPoet itself is from [KType.asTypeName()](https://github.com/square/kotlinpoet/blob/5395a8e2e0ef0f9e7d0ae044b66c5ef3e06db925/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt#L177) (and by extension `typeNameOf<>`)

Maybe just some doc would be enough, I'll try to suggest something.

martinbonnin:
Changed this PR to be "no-functional change". Only adding a paragraph about `kotlin-reflect` in the doc.

docs/index.md
```diff
@@ -1361,6 +1361,35 @@ fun factories() {
...
+dependencies {
+  implementation("com.squareup:kotlinpoet:1.9.0") {
+    exlude(module = "kotlin-reflect")
+  }
+}
+```
+
+The main APIs that require `kotlin-reflect` are [`KType.asTypeName()`](https://square.github.io/kotlinpoet/1.x/kotlinpoet/kotlinpoet/com.squareup.kotlinpoet/as-type-name.html) and [`typeNameOf<T>()`](https://square.github.io/kotlinpoet/1.x/kotlinpoet/kotlinpoet/com.squareup.kotlinpoet/type-name-of.html). If you're calling one of these without `kotlin-reflect` in the classpath and the type is generic or has annotations you will get a crash.
+
+You can replace it with code that passes type parameters or annotations explicitely and doesn't need `kotlin-reflect`. For example:
```
Egorand, line 1381: nit:
```suggestion
You can replace it with code that passes type parameters or annotations explicitly and doesn't need `kotlin-reflect`. For example:
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ParameterizedTypeName.kt
```diff
@@ -202,7 +202,10 @@ public class ParameterizedTypeName internal constructor(
 public fun ParameterizedType.asParameterizedTypeName(): ParameterizedTypeName =
   ParameterizedTypeName.get(this, mutableMapOf())
 
-/** Returns a class name equivalent to given Kotlin KType.  */
+/**
+ * Returns a [TypeName] equivalent to the given Kotlin KType using reflection, maybe using kotlin-reflect
+ * if required
```
Egorand, line 207: nit:
```suggestion
 * if required.
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -290,5 +290,9 @@ public fun KClass<*>.asTypeName(): ClassName = asClassName()
 @JvmName("get")
 public fun Type.asTypeName(): TypeName = TypeName.get(this, mutableMapOf())
 
+/**
+ * Returns a [TypeName] equivalent of the reified type parameter [T] using reflection, maybe using kotlin-reflect
+ * if required
```
Egorand, line 295: nit:
```suggestion
 * if required.
```

docs/index.md
```diff
@@ -1361,6 +1361,35 @@ fun factories() {
 Top-level classes and members with conflicting names may require aliased imports, as with
 [member names](#m-for-members).
 
+kotlin-reflect
+--------
+
+To generate source code from any [`KType`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-type/), including information that's not accessible to the builtin reflection APIs, KotlinPoet depends on [kotlin-reflect](https://kotlinlang.org/docs/reflection.html#jvm-dependency). `kotlin-reflect` can read the metadata of your classes and access this extra information. KotlinPoet can for an example, read the type parameters and their [variance](https://kotlinlang.org/docs/generics.html#variance) from a generic `KType` and generate appropriate source code.
+
+`kotlin-reflect` is a relatively big dependency though and in some cases it is desirable to remove it from the final executable/app to save some space and/or simplify the proguard/R8 setup. It is possible to do so and still use most of the KotlinPoet APIs:
```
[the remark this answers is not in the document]
martinbonnin: Agreed. My use case is actually in the context of a Gradle plugin. I'll replace app with "Gradle plugin" for more realistic examples.

PR #1119, opened 2021-08-11
Disable publish action in forked repository

Goooler:
I enabled the github actions in my forked repo for build and test, but triggered publish action after jvm job, it's unnecessary to publish, so I think we can disable it in forked repo.

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request]
+on: [ push, pull_request ]
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request]
+on: [ push, pull_request ]
```
ZacSweers, line 3: Let's avoid unnecessary changes

```suggestion
on: [push, pull_request]
```

ZacSweers (approved):

Egorand (approved):

.github/workflows/build.yml
Goooler: Reverted.

PR #901, opened 2020-04-22
Merge all Actions workflows into build.yml

Egorand:

PR #1124, opened 2021-08-17
Merge deploy-website action into build.yml

Goooler:
Like #901 did.
- Bump setup-java action of deploy-website to v2.
- Make deploy-website depends on jvm.


.github/workflows/build.yml
```diff
@@ -59,3 +59,44 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+
+      - name: Configure JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'zulu'
+          java-version: 14
```
JakeWharton, line 80:
```suggestion
```

We can just make this happen in the publish job

JakeWharton (approved):

.editorconfig
```diff
@@ -2,7 +2,6 @@ root = true
 
 [*]
 indent_size = 2
-end_of_line = lf
```
JakeWharton, line 5: Why remove this?

Goooler: Eol should be handled by `.gitattributes`.

JakeWharton: I'm not sure an IDE would read that file, though. Although I'm not sure whether _any_ idea actually uses CRLF by default anymore.

Goooler: I test it in windows, make a new `Test.kt` with CRLF by default.

PR #1126, opened 2021-08-18
Code cleanup

Goooler:


docs/index.md
```diff
@@ -1157,11 +1157,9 @@ This generates a method that contains a class that contains a method:
 
 ```kotlin
 class HelloWorld {
-    fun sortByLength(strings: List<String>) {
-        strings.sortedWith(object : Comparator<String> {
-            override fun compare(a: String, b: String): Int = a.length - b.length
-        })
-    }
+  fun sortByLength(strings: List<String>) {
+    strings.sortedWith { a, b -> a.length - b.length }
```
JakeWharton, line 1161: This code is generated from the above and needs to reflect what it produces.

interop/kotlinx-metadata/specs-tests/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/test/KotlinPoetMetadataSpecsTest.kt
```diff
@@ -1966,7 +1966,7 @@ class KotlinPoetMetadataSpecsTest : MultiClassInspectorTest() {
       // TODO Needless parens below necessary until Kotlin 1.4
       //  Or enable -XXLanguage:+NonParenthesizedAnnotationsOnFunctionalTypes
       //  https://youtrack.jetbrains.com/issue/KT-31734
```
JakeWharton, line 1968:
```suggestion
```

JakeWharton (approved):

Egorand (approved):

PR #1128, opened 2021-08-21
Move dependency declarations into a version catalog

Egorand:


build.gradle.kts
```diff
@@ -18,16 +18,16 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-  kotlin("jvm") version versions.kotlin apply false
-  id("org.jetbrains.dokka") version versions.dokka apply false
-  id("com.diffplug.spotless") version versions.spotless
-  id("com.vanniktech.maven.publish") version versions.mavenPublish apply false
+  kotlin("jvm") version libs.versions.kotlin.get() apply false
```
JakeWharton, line 21: This DSL is truly horrible

JakeWharton (approved):
So much nicer! Death to buildSrc!

ZacSweers (approved):
Nice

build.gradle.kts
Egorand: Indeed, was thinking to look for alternatives

Egorand:
Yeah, I like that there's a 1st class solution now, makes it easier to build tools around it, no need to support all kinds of weird ways people declare dependencies

build.gradle.kts
Goooler: `classpath` thems is a better way ?

Egorand: @Goooler agreed, this is much nicer! #1130

PR #1129, opened 2021-08-21
Configure spotless per subproject

Egorand:
That way runs can be parallelized and this also improves cacheability

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import com.diffplug.gradle.spotless.SpotlessExtension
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -73,4 +65,16 @@ subprojects {
       }
     }
   }
+
+  apply(plugin = "com.diffplug.spotless")
+  pluginManager.withPlugin("com.diffplug.spotless") {
```
ZacSweers, line 70: You don't need this if you're applying it right above

ZacSweers (approved):

build.gradle.kts
Egorand: Nice nice, will change

PR #1130, opened 2021-08-21
Declare plugins using a buildscript block

Egorand:


ZacSweers (commented):
Why not use the new plugins support in gradle 7.2?

```kotlin
plugins {
 alias(libs.plugins.dokka)
}
```

Goooler:
I think we can also merge #1131 into your branch, or master ?

Egorand:
> Why not use the new plugins support in gradle 7.2?

Sure! Wasn't aware it's a thing

PR #1132, opened 2021-08-21
Pull all hardcoded versions into the versions block

Egorand:
Just for consistency


ZacSweers (commented):
Is this actually better? Now it's separating the version from the dep sorta needlessly imo

Egorand:
@ZacSweers can do the other way around - only keep versions that apply to more than one dependency in the `[versions]` block and move everything else into lib declarations - would this be better?

Goooler:
> @ZacSweers can do the other way around - only keep versions that apply to more than one dependency in the `[versions]` block and move everything else into lib declarations - would this be better?

Yeah, only keep the reusable version strings in the `[versions]`.

PR #1133, opened 2021-08-21
Declare plugins via the version catalog

Egorand:


build.gradle.kts
```diff
@@ -19,10 +19,10 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-  kotlin("jvm") version libs.versions.kotlin.get() apply false
-  id("org.jetbrains.dokka") version libs.versions.dokka.get() apply false
-  id("com.diffplug.spotless") version libs.versions.spotless.get() apply false
-  id("com.vanniktech.maven.publish") version libs.versions.mavenPublish.get() apply false
+  alias(libs.plugins.kotlin.jvm)
```
ZacSweers, line 22: Do these need to apply false again?

Egorand: I added them back but I'm not 100% on the effects - the build seemed to work fine with or without. Guess without `apply false` we'd be applying all plugins to `tests` modules?

Goooler:
I think `apply` defalut value is false, no need to apply false again.
If it's true, `tasks.compileTestKotlin {}` extension should be avaliable in root project, you can have a try.

build.gradle.kts
Goooler: I think `apply` defalut value is false, no need to apply false again.
If it's true, `tasks.compileTestKotlin {}` extension should be avaliable in root project, you can have a try.

Egorand: [According to the docs](https://docs.gradle.org/current/javadoc/org/gradle/plugin/use/PluginDependencySpec.html#apply-boolean-), the default value for `apply` is true. That said, I'm not seeing any issues either way. Anyway, I'll merge as is to keep the scope of this PR small, and submit another one to remove `apply false`.

Goooler: I had aslo viewed that doc, just found the `apply` explanation on `id` but not `alias`.

ZacSweers: It applies by default, I was asking about the ones that were previously manually _not_ applied

Egorand: They all had `apply false` before these changes

PR #1140, opened 2021-08-31
Allow copying a ParameterizedTypeName with new type arguments

### LINKED ISSUE #1139: Support copying a ParameterizedTypeName with different type arguments
I need to create a wildcard projected type name and currently there is no way to do so using public APIs. I suggest a `copy` function is added which has all of the parameters of the class.

ansman:
This fixes #1139

### DIFF
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
@@ -112,6 +112,21 @@ class ParameterizedTypeNameTest {
     assertThat(typeName.toString()).isEqualTo("java.util.Map<java.lang.String, java.lang.Integer>")
   }
 
+  @Test fun copyingTypeArguments() {
+    val typeName = java.util.Map::class.java
+      .plusParameter(java.lang.String::class.java)
+      .plusParameter(java.lang.Integer::class.java)
+      .nestedClass(
+        "Entry",
+        listOf(
+          java.lang.String::class.java.asClassName(),
+          java.lang.Integer::class.java.asClassName()
+        )
+      )
+      .copy(typeArguments = listOf(STAR, STAR))
+    assertThat(typeName.toString()).isEqualTo("java.util.Map<java.lang.String, java.lang.Integer>.Entry<*, *>")
+  }
+
   interface Projections {
     val outVariance: KClass<out Annotation>
     val inVariance: KClass<in Test>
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
```diff
@@ -112,6 +112,15 @@ class ParameterizedTypeNameTest {
     assertThat(typeName.toString()).isEqualTo("java.util.Map<java.lang.String, java.lang.Integer>")
   }
 
+  @Test fun copyingTypeArguments() {
+    java.util.Map::class.java.plusParameter(java.lang.String::class.java)
```
Egorand, line 116: Does this line do anything?

Egorand (approved):
Thanks!

kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt
ansman: Damn, I left some debug stuff. But we can pretend I left it there to check if you would notice.

Egorand: I passed!

ZacSweers:
From an API standpoint I wonder if it would be better to have a withParameters instead. These overloaded copy functions are brittle historically

ansman:
I'd be OK with that too. In fact I was probably going to add a `mapTypeArguments` in my project too.

Egorand:
IMO `copy` is easier to discover with it being a Kotlin pattern. What's causing brittleness, the fact that it's hard to change in a backwards-compatible way?

ansman:
If we are keeping copy I think it might make sense to expose the enclosing type and add both that and the raw type to the copy method.

Egorand:
@ansman I'd be reluctant to expose those simply for consistency, without a use case. If there is a use case then we'll consider it.

ansman:
My current use cases are:
* Creating a wildcarded typename.
* Mapping Java types to Kotlin types (`List`, `Set`, `Annotation` etc). 

This isn't possible in all cases without exposing these things.

ZacSweers:
Can you add a test case that wasn't possible before? The one here is possible with the previous API

ansman:
Done

ZacSweers (approved):
This seems ok to me, though that example has me 🤔🤔🤔

ansman:
Yeah, the example isn't perhaps real since you'd access this as `Map.Entry<...>`. I essentially needed a nested class and the enclosing class needs to be parameterized to test the specific case.

PR #1142, opened 2021-09-06
Do not use reflection for kotlin.Unit

martinbonnin:
Explicitely use "kotlin.Unit" instead of using reflection. Allows to use kotlinpoet with a relocated kotlin stdlib if needed. Relocating kotlin.Unit isn't super useful in itself but it can happen and the change makes the codebase more consistent regardless.


ZacSweers (approved):

Egorand:
Nice!

PR #1145, opened 2021-09-12
Rearrange kdoc of MemberName

Goooler:
Looks like the comments of `MemberName` are not showing completely in [doc website](https://square.github.io/kotlinpoet/1.x/kotlinpoet/kotlinpoet/com.squareup.kotlinpoet/-member-name/index.html) as params were placed in middle place instead of the last lines. They should be moved to the last lines, right ?
Filed an issue [here](https://github.com/Kotlin/dokka/issues/2135).

image

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/MemberName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/MemberName.kt
@@ -20,13 +20,6 @@ import kotlin.reflect.KClass
 /**
  * Represents the name of a member (such as a function or a property).
  *
- * @param packageName e.g. `kotlin.collections`
- * @param enclosingClassName e.g. `Map.Entry.Companion`, if the member is declared inside the
- * companion object of the Map.Entry class
- * @param simpleName e.g. `isBlank`, `size`
- * @param isExtension whether the member is an extension property or an extension function. Default
- * is false.
- *
  * If there is a member with the same name as this member in a local scope, the generated code will
  * include this member's fully-qualified name to avoid ambiguity, e.g.:
  *
@@ -66,6 +59,13 @@ import kotlin.reflect.KClass
  *   }
  * }
  * ```
+ *
+ * @param packageName e.g. `kotlin.collections`
+ * @param enclosingClassName e.g. `Map.Entry.Companion`, if the member is declared inside the
+ * companion object of the Map.Entry class
+ * @param simpleName e.g. `isBlank`, `size`
+ * @param isExtension whether the member is an extension property or an extension function. Default
+ * is false.
  */
 public data class MemberName internal constructor(
   public val packageName: String,


ZacSweers:
This looks like a bug in Dokka. Please file an issue over there and share the link here before we'll merge. Also - does this show up correctly when you look at the doc in an actual project vs the website?

Goooler:
Rendering correctly after this commit as I run `gradlew dokka` for a test.

image

ZacSweers:
In a real project, like in an IDE.

Goooler:
Before: image
Now: image

ZacSweers:
Interesting. Does the language spec say they must be last?

Goooler:
> Interesting. Does the language spec say they must be last?

I guess.

ZacSweers:
Can you do better than a guess? Either it's not in the spec and a bug should be filed in Dokka/Kotlin to show these tags, or it's in the spec and about should be filed to warn about these misplaced tags.

Goooler:
Actually I have not found the source about block tags must be the last, and I already filed a bug at dokka.

ZacSweers:
Please link the issue

Goooler:
Can't you see ?

image

ZacSweers:
No, I can't. Please add it to the description.

image

Egorand (commented):
The block starting on line 30 (`If there is a member...`) is actually part of the `@param isExtension` documentation, so it's placed properly, as you can see [here](https://square.github.io/kotlinpoet/1.x/kotlinpoet/kotlinpoet/com.squareup.kotlinpoet/-member-name/index.html). The IDE also seems to render it correctly.

ZacSweers:
I actually completely missed that, good catch. In that case, this doc is correct as-is.

Goooler:
One more question, does this kotlin code block rendered correctly (without hightlights) ?

image

ZacSweers:
Please take questions like that to Dokka

Egorand:
> does this kotlin code block rendered correctly (without hightlights) ?

Not sure whether Dokka supports code highlighting or not, our Markdown seems correct and I'm not sure what else can be done. If you know how to turn highlighting on that'd be great!

Goooler:
> Not sure whether Dokka supports code highlighting or not, our Markdown seems correct and I'm not sure what else can be done. If you know how to turn highlighting on that'd be great!

For now this is an expected behaviour, but will change in our next major release which is quite soon ☺️
_Originally posted by @MarcinAman in https://github.com/Kotlin/dokka/issues/2136#issuecomment-917925281_

Like a good news.

Egorand:
Nice, and pretty sure we'll just get that for free!

Goooler:
After #1192, the website code highlights have been rendered correctly.

image

Egorand:
Oh neat, I like the UI in 1.6.0

PR #1146, opened 2021-09-12
Implement basic KSP interop

ZacSweers:
This starts a basic KSP interop project, with extensions covering common use cases like...
- Basic KS type -> KotlinPoet `TypeName` type conversion
- Support for originating KSFiles to support KSP's incremental processing
- Support for writing `FileSpec` to `CodeGenerator`s
- Annotated with a new `@KotlinPoetKspPreview` annotation while we fine-tune the API.

This is largely upstreamed and cleaned up from the Moshi KSP implementation currently being upstreamed over [here](https://github.com/square/moshi/pull/1393).

What this does _not_ attempt to do: the full "give me a class and I'll give you a TypeSpec of the API" like kotlinx-metadata interop does. I'm open to considering it if we think it's worthwhile, but unlike with metadata we don't need to run everything from a class's `@Metadata` annotation for this to work, so I'm currently erring on the side of keeping the interop lightweight. There could be a good use-case for some existing helpers like `overriding`, `AnnotationSpec.get()`, `ParameterSpec.get()`, etc, but we can see if the community asks for that later.

TODO:
- [ ] Tests. Moshi used integration tests as cover for this. We'll need to figure out a good way to test these here. Open to saving this for a followup PR.

### DIFF
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.squareup.kotlinpoet.ksp
 
 import com.google.devtools.ksp.symbol.KSType
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
@@ -26,12 +26,9 @@ public interface TypeParameterResolver {
      * are known to not have arguments, such as top-level classes.
      */
     public val EMPTY: TypeParameterResolver = object : TypeParameterResolver {
-      override val parametersMap: Map<String, TypeVariableName>
-        get() = emptyMap()
+      override val parametersMap: Map<String, TypeVariableName> = emptyMap()
 
-      override fun get(index: String): TypeVariableName {
-        error("No TypeParameter found for index $index")
-      }
+      override fun get(index: String): TypeVariableName = throw NoSuchElementException("No TypeParameter found for index $index")
     }
   }
 }
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
@@ -1,0 +1,85 @@
+package com.squareup.kotlinpoet.ksp
+
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSTypeParameter
+import com.squareup.kotlinpoet.TypeVariableName
+
+/**
+ * A resolver for enclosing declarations' type parameters. Parent declarations can be anything with
+ * generics that child nodes declare as defined by [KSType.arguments].
+ *
+ * This is important for resolving inherited generics on child declarations, as KSP interop
+ * otherwise can't resolve them.
+ *
+ * In general, you want to retrieve an instance of this via [toTypeParameterResolver].
+ *
+ * @see toTypeParameterResolver
+ */
+@KotlinPoetKspPreview
+public interface TypeParameterResolver {
+  public val parametersMap: Map<String, TypeVariableName>
+  public operator fun get(index: String): TypeVariableName
+
+  public companion object {
+    /**
+     * An empty instance of [TypeParameterResolver], only should be used if enclosing declarations
+     * are known to not have arguments, such as top-level classes.
+     */
+    public val EMPTY: TypeParameterResolver = object : TypeParameterResolver {
+      override val parametersMap: Map<String, TypeVariableName>
+        get() = emptyMap()
+
+      override fun get(index: String): TypeVariableName {
+        error("No TypeParameter found for index $index")
+      }
+    }
+  }
+}
+
+/**
+ * Returns a [TypeParameterResolver] for this list of [KSTypeParameters][KSTypeParameter] for use
+ * with enclosed declarations.
+ *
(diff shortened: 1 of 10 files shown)


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * An empty instance of [TypeParameterResolver], only should be used if enclosing declarations
+     * are known to not have arguments, such as top-level classes.
+     */
+    public val EMPTY: TypeParameterResolver = object : TypeParameterResolver {
+      override val parametersMap: Map<String, TypeVariableName>
+        get() = emptyMap()
+
+      override fun get(index: String): TypeVariableName {
+        error("No TypeParameter found for index $index")
```
Egorand, line 33: nit: should it throw something like `UnsupportedOperationException`?

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Note that if none are specified, the [originatingKSFiles] argument defaults to using
+ * [FileSpec.originatingKSFiles], which will automatically resolve any files added to the
+ * contained declarations.
+ *
+ * See [the docs](https://github.com/google/ksp/blob/main/docs/incremental.md) for more information.
+ *
+ * @see FileSpec.originatingKSFiles
+ * @param codeGenerator the [CodeGenerator] to write to
+ * @param aggregating flag indicating if this is an aggregating
```
Egorand, line 113: nit: "aggregating annotation processor"?

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Note that if no [originatingKSFiles] are specified, the [originatingKSFiles] argument defaults
+ * to using [FileSpec.originatingKSFiles], which will automatically resolve any files added to the
+ * contained declarations.
+ *
+ * See [the docs](https://github.com/google/ksp/blob/main/docs/incremental.md) for more information.
+ *
+ * @see FileSpec.originatingKSFiles
+ * @see FileSpec.writeTo
+ * @param aggregating flag indicating if this is an aggregating
```
Egorand, line 140: Same

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
```diff
@@ -0,0 +1,30 @@
...
+      var count = parameters.size
+      if (receiver != null) {
+        count++
+      }
+      val functionSimpleName = if (count >= 23) {
+        "FunctionN"
+      } else {
+        "Function$count"
+      }
+      ClassName("kotlin.jvm.functions", functionSimpleName)
```
Egorand, line 26: Is there a multiplatform-friendly alternative?

Egorand (approved):
Dope, thanks! nit: need copyright headers where they're missing.

Tests would definitely be great, mostly to provide usage examples and to make it safer to iterate on this code further. A sample project would be nice too.

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
ZacSweers: Not in this case I think, this is an unexpected condition rather than something unsupported or unimplemented

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
ZacSweers: Good catch, though I think we want to say "symbol processor"

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
ZacSweers: I'm not sure to be honest. The [spec](https://github.com/JetBrains/kotlin/blob/master/spec-docs/function-types.md) focuses mostly on JVM and agnostic impls, but doesn't explicitly call out how it works on other platforms. Seeing as we do this here and in metadata, mind if we save this for a later day?

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
Egorand: Called out as based on the message it may seem that this API will return something for a different `index`, when in fact it will always throw.

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
Egorand: Agree

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
Egorand: For sure, especially since I believe KSP only targets JVM with 1.0? Or it will support MPP?

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
ZacSweers: ahhh I didn't realize where you were commenting (was looking on mobile). Good point

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
ZacSweers: KSP will support MPP yeah (insofar as it will be able to run code gen for MPP projects)

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/TypeParameterResolver.kt
ZacSweers: Went with NSEE to match what we do in metadata for this 74e323a

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
ZacSweers: cb6f2c5

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/originatingKSFiles.kt
ZacSweers: cb6f2c5

ZacSweers:
Talked offline, will work at some basic test infra for this in a followup PR. Will likely entail writing a sample processor that parses a class and then generates code that shows it read it correctly.

PR #1147, opened 2021-09-12
Implement basic JavaPoet interop module

### LINKED ISSUE #417: JavaPoet Interop
This is more of a question/feature suggestion. I'm working on a project where I'm using both JavaPoet and KotlinPoet (JavaPoet for some code and then Kotlin extensions that make the API sweeter for Kotlin users). 

One thing that I really wished for would be mapping between JavaPoet and KotlinPoet using extension functions. Possible use-cases would be to map between JavaPoet's `ClassName` and KotlinPoet's `ClassName` with something like `com.squareup.javapoet.ClassName.toKClassName()` which will give you the corresponding KotlinPoet `ClassName`. This would be similar for parameters, methods etc and give library developers a very easy path to write Kotlin extensions for their annotation processing libraries with minimal change to their existing codebase.

ZacSweers:
Resolves #417 and is based on the implementation I wrote in the past that @shaishavgandhi shared in the issue via https://gist.github.com/ShaishavGandhi/097033cc528ae25741186973e4d36ce4

This'll be pretty helpful for processors like Room that support generating both Java and Kotlin. I know we weren't sure a few years ago when the issue was first created, but I think KSP's dual support will create new avenues for this where this becomes helpful. We also have a bit of a stronger interop artifacts story now too.

TODO:
- [x] Tests (see below)
- [x] `WildcardTypeName` interop
- [x] Do we want to use an experimenta/preview annotation for these?
- [x] What about other Kotlin intrinsics? Should we cover the set contained in KotlinPoet itself (beyond just `LIST`, `SET`, etc)?

### DIFF
--- a/interop/javapoet/gradle.properties
+++ b/interop/javapoet/gradle.properties
@@ -1,0 +1,4 @@
+POM_ARTIFACT_ID=kotlinpoet-ksp
+POM_NAME=KotlinPoet (KSP Interop)
+POM_DESCRIPTION=Extensions for interop with KSP (Kotlin Symbol Processing).
+POM_PACKAGING=jar
--- a/interop/javapoet/gradle.properties
+++ b/interop/javapoet/gradle.properties
@@ -1,4 +1,4 @@
-POM_ARTIFACT_ID=kotlinpoet-ksp
-POM_NAME=KotlinPoet (KSP Interop)
-POM_DESCRIPTION=Extensions for interop with KSP (Kotlin Symbol Processing).
+POM_ARTIFACT_ID=kotlinpoet-javapoet
+POM_NAME=KotlinPoet (JavaPoet Interop)
+POM_DESCRIPTION=Extensions for interop with JavaPoet.
 POM_PACKAGING=jar
--- a/interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/k2jInterop.kt
+++ b/interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/k2jInterop.kt
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.javapoet
+
+import com.squareup.javapoet.ArrayTypeName
+import com.squareup.kotlinpoet.ANY
+import com.squareup.kotlinpoet.ARRAY
+import com.squareup.kotlinpoet.BOOLEAN
+import com.squareup.kotlinpoet.BYTE
+import com.squareup.kotlinpoet.CHAR
+import com.squareup.kotlinpoet.DOUBLE
+import com.squareup.kotlinpoet.Dynamic
+import com.squareup.kotlinpoet.FLOAT
+import com.squareup.kotlinpoet.INT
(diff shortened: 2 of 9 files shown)


ZacSweers:
@ShaishavGandhi could you share the tests that were associated with these interop extensions?

interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/typeAliases.kt
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.javapoet
+
+/*
+ * Useful typealiases for colliding names
+ */
```
Egorand, line 20: nit (I believe as is this would be attached as doc for `KTypeName` which seems wrong):
```suggestion
// Useful typealiases for colliding names
```

interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/k2jInterop.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.squareup.kotlinpoet.INT
+import com.squareup.kotlinpoet.LIST
+import com.squareup.kotlinpoet.LONG
+import com.squareup.kotlinpoet.LambdaTypeName
+import com.squareup.kotlinpoet.MAP
+import com.squareup.kotlinpoet.SET
+import com.squareup.kotlinpoet.SHORT
+import com.squareup.kotlinpoet.STRING
+
+public fun KClassName.toJClassName(shouldBox: Boolean = false): JTypeName {
```
Egorand, line 36: nit: is `boxIfPrimitive` a better name than `shouldBox`?

interop/javapoet/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=kotlinpoet-ksp
+POM_NAME=KotlinPoet (KSP Interop)
```
Egorand, line 2: Copy pasta!

Egorand (approved):
Sweet! Yep, tests would be great.

interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/typeAliases.kt
ZacSweers: It actually won't be attached to any elements if it's just a single star statement like this. It only becomes a full kdoc if you have two stars (`/**`)

interop/javapoet/gradle.properties
ZacSweers: 🙈

interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/typeAliases.kt
Egorand: Ah, true, nvm then!

interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/k2jInterop.kt
ZacSweers: 5d15587

interop/javapoet/gradle.properties
ZacSweers: cf35d5a

ZacSweers:
@Egorand thoughts on the last two TODO items in the desciption?

Egorand:
>  Do we want to use an experimenta/preview annotation for these?

I'd say yes, will give us more flexibility to change the API if needed.

>  What about other Kotlin intrinsics? Should we cover the set contained in KotlinPoet itself (beyond just LIST, SET, etc)?

Can definitely do this in followup PRs, and postpone until there are use cases for it. Another idea would be to open an issue and accept external contributions for these features.

ShaishavGandhi (approved):
LGTM! Updated the gist with the tests.

ZacSweers:
Added a bunch of tests (and fixed some bugs they revealed!), the preview annotation, and more intrinsics coverage. I think this is a good enough start!

PR #1149, opened 2021-09-15
Add KSP docs

ZacSweers:


docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
+KotlinPoet-ksp
+==============
+
+`interop:ksp` is an interop API for converting
+[Kotlin Symbol Processing](https://github.com/google/ksp) (KSP) types to KotlinPoet types and
```
Egorand, line 5: nit: I like the style of moving all of the referenced links to the bottom of the document and using short refs, makes markdown easier to read:
```
[Kotlin Symbol Processing][ksp]
...

 [ksp]: https://github.com/google/ksp
```

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+writing to KSP `CodeGenerator`.
+
+Note that this API is currently in preview and subject to API changes. Usage of them requires opting
```
Egorand, line 8: nit:
```suggestion
Note that this API is currently in preview and subject to API changes. Usage of it requires opting
```

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+in to the `@KotlinPoetKspPreview` annotation.
+
+### Examples
+
+Examples are based on reading the following property as a `KSProperty`
```
Egorand, line 13: nit:
```suggestion
Examples are based on reading the following property as a `KSProperty`:
```

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+Type parameters can be declared on classes, functions, and typealiases. These parameters are then
+available to all of its enclosed elements. In order for these elements to resolve these in KSP, you
+must be able to reference these type parameters by their _index_.
+
+In `kotlinpoet-ksp` this is orchestrated by the `TypeParameterResolver` API, which can be passed
+into most `toTypeName()` (or similar) functions to give them access to enclosing type parameters
+that they may reference.
+
+The canonical way to create an instance of this is to call `toTypeParameterResolver()` on a
+`List<toTypeParameterResolver>`.
```
Egorand, line 62:
```suggestion
`List<KSTypeParameter>`.
```

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+abstract class Taco<T> {
+  abstract val seasoning: T
+}
+```
+
+To properly resolve the type of `seasoning`, we need to pass the class `TypeParameterResolver` to
+`toTypeName()` so that it can properly resolve it.
+
+```kotlin
+val classTypeParams = ksClassDeclaration.typeParameters.toTypeParameterResolver()
```
Egorand, line 76: Not clear in this example where `ksClassDeclaration` and `seasoningsKsProperty` come from, maybe extend the example to include to code for obtaining those?

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+```
+
+To resolve its parameters, we need to create a `TypeParameterResolver` from the function's
+`typeParameters` and _compose_ it with the enclosing class's type parameters as a `parent`.
+
+```kotlin
+val classTypeParams = ksClassDeclaration.typeParameters.toTypeParameterResolver()
+val functionTypeParams = ksFunction.typeParameters.toTypeParameterResolver(parent = classTypeParams)
```
Egorand, line 99: I like this API!

docs/interop-ksp.md
```diff
@@ -0,0 +1,126 @@
...
+// returns `[E, T]`
+val seasoningType = ksFunction.parameterTypes.map { it.toTypeName(functionTypeParams) }
+```
+
+### Incremental Processing
+
+KSP supports [incremental processing](https://github.com/google/ksp/blob/main/docs/incremental.md) as
+long as symbol processors properly indicate originating files in generated new files and whether or
+not they are `aggregating`. `kotlinpoet-ksp` supports this via `OriginatingKSFiles`, which is a simple
+API that sits atop KotlinPoet's `Taggable` API. To use this, simple add relevant originating files to
```
Egorand, line 109: nit:
```suggestion
API that sits atop KotlinPoet's `Taggable` API. To use this, simply add relevant originating files to
```

Egorand (approved):
Awesome!

docs/interop-ksp.md
ZacSweers: Does that still work in mkdocs?

docs/interop-ksp.md
ZacSweers: it can come from many places, I intentionally made it agnostic. I think trying to show where they come from would be something for KSP docs that we already link. Making a fuller example wouldn't really be grokkable at a glance

docs/interop-ksp.md
Egorand: Yes!

docs/interop-ksp.md
Egorand: Gotcha

docs/interop-ksp.md
ZacSweers: 4ec59a9

docs/interop-ksp.md
ZacSweers: yeah it's worked well. We use basically the same thing in metadata, so I wonder if it would be worth trying to share somewhere.

PR #1151, opened 2021-09-18
Consolidate and simplify kotlinx-metadata interop

ZacSweers:
This rehashes a previous attempt to simplify the kotlinx-metadata artifacts into a single artifact.

The commits here are intended to be focused and easily digestible.

The final structure is we now have a single `interop:kotlinx-metadata` project and analogous single artifact. The previous artifacts (core, specs, classinspectors) are all merged into this project now. The previous `specs-tests` are also now merged into this project's tests.

API changes:
- Immutable implementations of Km types are now removed. They didn't offer much value, their equality/hashcode semantics were unstable due to the underlying types not always implementing them, and in general just unnecessary. Everything runs through the main `Km` types
- One change as a result of the previous bullet is we had to implement a custom comparator for Maps with `Km` types, since the raw ones didn't implement equals. The custom comparators just use names, signatures, and (as a last resort) a simple computed key.

Most of the previous public APIs are the same otherwise! Considering most usages just converted to the immutable type and then to the typespec, adoption of this should be fairly simple.


docs/interop-kotlinx-metadata.md
```diff
@@ -35,21 +35,50 @@ println("Is class? ${kmClass.isClass}")
...
+represented sources can be incomplete without this information available. Reflective and javax
+`Elements` implementations are available under the `kotlinpoet-classinspector-*` artifacts.
+
+Generated sources are solely _stub_ implementations, meaning implementation details of elements
+like functions, property getters, and delegated properties are simply stubbed with `TODO()`
+placeholders.
+
+### Known limitations
+
+- Only `KotlinClassMetadata.Class` supported for now. No support for `FileFacade`, `SyntheticClass`, `MultiFileClassFacade`, or `MultiFileClassPart`
```
ZacSweers, line 76:
```suggestion
- Only `KotlinClassMetadata.Class` and `KotlinClassMetadata.FileFacade` are supported for now. No support for `SyntheticClass`, `MultiFileClassFacade`, or `MultiFileClassPart`
```

Out of date

docs/interop-kotlinx-metadata.md
```diff
@@ -35,21 +35,50 @@ println("Is class? ${kmClass.isClass}")
...
+- `@file:` annotations are not supported yet.
```
ZacSweers, line 77:
```suggestion
```

No longer true!

docs/interop-kotlinx-metadata.md
```diff
@@ -35,21 +35,50 @@ println("Is class? ${kmClass.isClass}")
...
+- `@JvmOverloads` annotations are only supported with `ElementsClassInspector` and not reflection.
+- Non-const literal values are only supported with `ReflectiveClassInspector` and not reflection.
```
ZacSweers, line 79:
```suggestion
- Non-const literal values are only supported with `ElementsClassInspector` and not reflection.
```

This is incorrect

ZacSweers (commented):
Noticed some stale docs

docs/interop-kotlinx-metadata.md
```diff
@@ -35,21 +35,49 @@ println("Is class? ${kmClass.isClass}")
...
+
+The generated representations are a _best effort_ representation of the underlying source code.
+This means that synthetic elements will be excluded from generation. Kotlin-specific language
+features like lambdas or delegation will be coerced to their idiomatic source form.
+
+To aid with this, `toTypeSpec()` and `toFileSpec()` accept optional `ClassInspector` instances
+to assist in parsing/understanding the underlying JVM code. This is important for things like
+annotations, companion objects, certain JVM modifiers, overrides, and more. While it is optional,
+represented sources can be incomplete without this information available. Reflective and javax
+`Elements` implementations are available under the `kotlinpoet-classinspector-*` artifacts.
```
Egorand, line 68: All class inspectors are in this artifact now, need to reword this statement?

interop/kotlinx-metadata/build.gradle.kts
```diff
@@ -16,14 +16,24 @@
...
+
+tasks.compileTestKotlin {
+  kotlinOptions {
+    freeCompilerArgs = listOf("-Xjvm-default=all")
   }
 }
 
 dependencies {
+  api(libs.autoCommon)
+  api(libs.guava)
```
Egorand, line 31: Any way we can get guava off the classpath?

Egorand (commented):
Nice!

Egorand (approved):

interop/kotlinx-metadata/build.gradle.kts
ZacSweers: Not easily but we can at least make them implementation deps

image

docs/interop-kotlinx-metadata.md
ZacSweers: f4b001b

interop/kotlinx-metadata/build.gradle.kts
ZacSweers: f4b001b

Egorand: Yeah, that's better, assuming the types don't appear in the ABI

PR #1152, opened 2021-09-20
Unify `TypeAliasTag` API between metadata and KSP

ZacSweers:
This aligns the KSP implementation of this with metadata by just storing the abbreviated type in a shared `TypeAliasTag`. This also reuses moshi's implementation of unwrapping them for its test. I kind of think this would be worth considering moving to KotlinPoet's main API as it could be reusable as a companion to `TypeAliasTag`.


ZacSweers:
Note that this will conflict with #1151 but I'll rebase appropriately whichever is merged first.

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
```diff
@@ -68,16 +66,17 @@ public fun KSType.toTypeName(
...
+
+      val abbreviatedType = decl.type.resolve()
+        .toTypeName(extraResolver)
+        .copy(nullable = isMarkedNullable)
+        .rawType()
+        .withTypeArguments(mappedArgs)
+
+      decl.toClassNameInternal()
+        .withTypeArguments(mappedArgs)
+        .copy(tags = mapOf(TypeAliasTag::class to TypeAliasTag(abbreviatedType)))
```
Egorand, line 79: So we'll unwrap a typealias by default now and attach a `TypeAliasTag` to indicate that the type was unwrapped? If the use case needs typealises to not be unwrapped, would it be easy to achieve this by using the tag? Or is there a different use case here?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/tags/TypeAliasTag.kt
```diff
@@ -0,0 +1,10 @@
+package com.squareup.kotlinpoet.tags
+
+import com.squareup.kotlinpoet.TypeName
+
+/**
+ * This tag indicates that this [TypeName] represents a `typealias` type.
+ *
+ * @property [abbreviatedType] the underlying type for this alias.
+ */
+public class TypeAliasTag(public val abbreviatedType: TypeName)
```
Egorand, line 10: Would it make sense to make this a `value` class?

Egorand (approved):

kotlinpoet/src/main/java/com/squareup/kotlinpoet/tags/TypeAliasTag.kt
ZacSweers: Seeing as we have to let it by the class, I don't really know if there would be any value to it

Egorand: Ah, true

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
Egorand: Got it!

PR #1161, opened 2021-09-20
Remove Guava dependency

Egorand:
Doesn't look like it's used?


ZacSweers:
It's being used here: https://github.com/square/kotlinpoet/blob/master/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ElementsClassInspector.kt#L21-L22

Only reason PR doesn't fail is because autocommon still transitively includes it

Egorand:
Ah, I see, all good then

PR #1162, opened 2021-09-20
Correct generation of typealiases with type args

evant:
If you had `typealias Foo = List<String>` then `toTypeName()` would
incorrectly generate `Foo<String>` instead of `Foo`. This is because
`KSType.arguments`, returns the arguments of the underlying type, _not_
the alias. In order to get the alias args you need to use
`KSTypeReference.element.typeArguments` instead.

### DIFF
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
@@ -53,7 +53,8 @@ public fun KSType.toClassName(): ClassName {
  */
 @KotlinPoetKspPreview
 public fun KSType.toTypeName(
-  typeParamResolver: TypeParameterResolver = TypeParameterResolver.EMPTY
+  typeParamResolver: TypeParameterResolver = TypeParameterResolver.EMPTY,
+  typeArguments: List<KSTypeArgument> = emptyList(),
 ): TypeName {
   val type = when (val decl = declaration) {
     is KSClassDeclaration -> {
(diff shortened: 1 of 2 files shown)


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
```diff
@@ -53,7 +53,8 @@ public fun KSType.toClassName(): ClassName {
  */
 @KotlinPoetKspPreview
 public fun KSType.toTypeName(
-  typeParamResolver: TypeParameterResolver = TypeParameterResolver.EMPTY
+  typeParamResolver: TypeParameterResolver = TypeParameterResolver.EMPTY,
+  typeArguments: List<KSTypeArgument> = emptyList(),
```
evant, line 57: Not sure how to handle this api-wise. I just added it as an extra argument to get this info but there may be a better way.

ZacSweers: Let's keep the original public signature but then move the full impl with this overload to an internal separate function. We can open it up later if there's a need for it

ZacSweers (commented):
Thanks for this, I just ran into the same issue in testing this out in another repo

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt
ZacSweers, line 57:
```suggestion
  typeParamResolver: TypeParameterResolver = TypeParameterResolver.EMPTY
): TypeName = toTypeName(typeParamResolver, emptyList())

@KotlinPoetKspPreview
internal fun KSType.toTypeName(
  typeParamResolver: TypeParameterResolver,
  typeArguments: List<KSTypeArgument>,
): TypeName {
```

ZacSweers (approved):

PR #1164, opened 2021-09-22
Bump ksp to 1.5.31-1.0.0

Goooler:
[1.5.31-1.0.0](https://github.com/google/ksp/releases/tag/1.5.31-1.0.0).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,7 +15,7 @@
 [versions]
 kotlin = "1.5.31"
 kct = "1.4.4"
-ksp = "1.5.30-1.0.0"
+ksp = "1.5.31-1.0.0"
 ktlint = "0.42.1"
 
 [plugins]


ZacSweers:
Does this update fix a specific issue that you encountered in a project consuming KotlinPoet?

Goooler:
Nope, just follow the version.

Egorand:
There's usually not a lot of value in minor version bumps, and there's always risk of introducing breaking changes. We decided to not integrate dependabot because it's more noise than value, but manual version bumps create the same amount of noise. I suggest ignoring minor version increments unless we have a proper use case - e.g. when a version bump addresses an issue KotlinPoet faces.

That said, I really appreciate your contributions to this repo and by no means want to detract you from contributing in the future! We've got a fair number of bugs we haven't managed to address yet, please feel free to pick any issue labelled with "bug" and send us PRs - that'd be super helpful.

Egorand (approved):

PR #1167, opened 2021-09-26
Ignore java deprecated annotations on synthetic methods for annotations

### LINKED ISSUE #1046: @Deprecated annotation added on a property
Im generating a class from list of property specs. The list has primitive and String elements filtered from the original class. 
Generated class is supposed to only have props from the list.

Original class:

```
@Serializable
data class Person(
    @SerialName("mName") val name: String,
    val surname: String?,
    val age: Int,
    val dog: Dog,
    val listOfDogs: List<Dog>,
    val listOfInts: List<Int>
)
```

Using metadata I successfully filtered properties and made them into a `List<PropertySpec>`. Im leaving the code for that too, in case it is useful to solve the issue. Otherwise, just skip next two blocks.

```
        val metadata = element.getAnnotation(Metadata::class.java)
        val typeSpec = metadata.toImmutableKmClass().toTypeSpec(
            ElementsClassInspector.create(processingEnv.elementUtils, processingEnv.typeUtils)
        )
        val membersSeparator = PropertyTypesSeparator(typeSpec)
        val primitives = membersSeparator.getPrimitiveProperties()
```

```
internal class PropertyTypesSeparator(private val classType: TypeSpec) {

    private val primitiveFields = mutableListOf<PropertySpec>()
    private val compositeFields = mutableListOf<PropertySpec>()

    init {
        processClassParameters()
    }

    fun getPrimitiveProperties(): List<PropertySpec> {
        return primitiveFields.toList()
    }

    fun getCompositeProperties(): List<PropertySpec> {
        return compositeFields.toList()
    }

    private fun processClassParameters() {
        classType.propertySpecs.forEach { property ->
            if (property.type is ParameterizedTypeName) {
                if ((property.type as ParameterizedTypeName).isCollection()) {
                    primitiveFields.add(property)
                } else {
                    compositeFields.add(property)
                }
            } else {
                if (property.type.isPrimitiveOrString()) {
                    primitiveFields.add(property)
                } else {
                    compositeFields.add(property)
                }
            }
        }
    }

    private fun TypeName.isPrimitiveOrString(): Boolean {
        return when (this) {
            STRING,
            BOOLEAN,
            BYTE,
            SHORT,
            INT,
            LONG,
            CHAR,
            FLOAT,
            DOUBLE,
            CHAR_SEQUENCE,
            NUMBER,
            BOOLEAN_ARRAY,
            BYTE_ARRAY,
            SHORT_ARRAY,
            INT_ARRAY,
            LONG_ARRAY,
            CHAR_ARRAY,
            FLOAT_ARRAY,
            DOUBLE_ARRAY,
            U_BYTE,
            U_SHORT,
            U_INT,
            U_LONG,
            U_BYTE_ARRAY,
            U_SHORT_ARRAY,
            U_INT_ARRAY,
            U_LONG_ARRAY,
            STRING.copy(nullable = true),
            BOOLEAN.copy(nullable = true),
            BYTE.copy(nullable = true),
            SHORT.copy(nullable = true),
            INT.copy(nullable = true),
            LONG.copy(nullable = true),
            CHAR.copy(nullable = true),
            FLOAT.copy(nullable = true),
            DOUBLE.copy(nullable = true),
            CHAR_SEQUENCE.copy(nullable = true),
            NUMBER.copy(nullable = true),
            BOOLEAN_ARRAY.copy(nullable = true),
            BYTE_ARRAY.copy(nullable = true),
            SHORT_ARRAY.copy(nullable = true),
            INT_ARRAY.copy(nullable = true),
            LONG_ARRAY.copy(nullable = true),
            CHAR_ARRAY.copy(nullable = true),
            FLOAT_ARRAY.copy(nullable = true),
            DOUBLE_ARRAY.copy(nullable = true),
            U_BYTE.copy(nullable = true),
            U_SHORT.copy(nullable = true),
            U_INT.copy(nullable = true),
            U_LONG.copy(nullable = true),
            U_BYTE_ARRAY.copy(nullable = true),
            U_SHORT_ARRAY.copy(nullable = true),
            U_INT_ARRAY.copy(nullable = true),
            U_LONG_ARRAY.copy(nullable = true) -> true
            else -> false
        }
    }

    private fun ParameterizedTypeName.isCollection(): Boolean {

        return this.typeArguments.all { it.isPrimitiveOrString() } && when (rawType) {
            ARRAY,
            ITERABLE,
            COLLECTION,
            LIST,
            SET,
            MAP,
            MAP_ENTRY,
            MUTABLE_ITERABLE,
            MUTABLE_COLLECTION,
            MUTABLE_LIST,
            MUTABLE_SET,
            MUTABLE_MAP,
            MUTABLE_MAP_ENTRY,
            ARRAY.copy(nullable = true),
            ITERABLE.copy(nullable = true),
            COLLECTION.copy(nullable = true),
            LIST.copy(nullable = true),
            SET.copy(nullable = true),
            MAP.copy(nullable = true),
            MAP_ENTRY.copy(nullable = true),
            MUTABLE_ITERABLE.copy(nullable = true),
            MUTABLE_COLLECTION.copy(nullable = true),
            MUTABLE_LIST.copy(nullable = true),
            MUTABLE_SET.copy(nullable = true),
            MUTABLE_MAP.copy(nullable = true),
            MUTABLE_MAP_ENTRY.copy(nullable = true) -> true
            else -> false
        }
    }
}
```

This is code that I use for generation:

```
    fun build(
        attributes: List<PropertySpec>,
        packageName: String,
        className: String
    ): FileSpec {
        val parameterSpecs = attributes.map {
            ParameterSpec.builder(it.name, it.type).build()
        }

        return FileSpec.builder(packageName, className)
            .addType(
                TypeSpec.classBuilder(generatedName)
                    .addSuperinterface(AttributesModel::class)
                    .addAnnotation(Serializable::class.asClassName())
                    .primaryConstructor(
                        FunSpec.constructorBuilder()
                            .addParameters(parameterSpecs)
                            .build()
                    )
                    .addProperties(attributes)
                    .build()
            )
            .build()
    }
```

This is the output class that I get:

```
@Serializable
public data class AttributesModel_Person(
  public val age: Int,
  public val listOfInts: List<Int>,
  @property:Deprecated
  @property:SerialName(value = "mName")
  public val name: String,
  public val surname: String?
) : AttributesModel
```

I can't figure out why Im getting the `@property:Deprecated` above `public val name: String`.
Without that annotation everything works fine.
`@SerialName` has the `AnnotationTarget.PROPERTY` so it should be fine. Also I tried playing around with adding the annotations to constructor parameter specs instead. Got the same result.

ZacSweers:
Resolves #1046. We can't properly test it because it's a kept-specific issue and our tests don't actually use kapt. Filed an issue on kapt separately here https://youtrack.jetbrains.com/issue/KT-48922

### DIFF
--- a/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ClassInspectorUtil.kt
+++ b/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ClassInspectorUtil.kt
@@ -54,6 +54,7 @@ internal object ClassInspectorUtil {
   internal val JVM_FIELD_SPEC = AnnotationSpec.builder(JVM_FIELD).build()
   internal val JVM_SYNTHETIC = JvmSynthetic::class.asClassName()
   internal val JVM_SYNTHETIC_SPEC = AnnotationSpec.builder(JVM_SYNTHETIC).build()
+  internal val JAVA_DEPRECATED = java.lang.Deprecated::class.asClassName()
   private val JVM_TRANSIENT = Transient::class.asClassName()
   private val JVM_VOLATILE = Volatile::class.asClassName()
   private val IMPLICIT_FIELD_ANNOTATIONS = setOf(
--- a/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ElementsClassInspector.kt
+++ b/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ElementsClassInspector.kt
@@ -28,6 +28,7 @@ import com.squareup.kotlinpoet.TypeName
 import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
 import com.squareup.kotlinpoet.metadata.KotlinPoetMetadataPreview
+import com.squareup.kotlinpoet.metadata.classinspectors.ClassInspectorUtil.JAVA_DEPRECATED
 import com.squareup.kotlinpoet.metadata.classinspectors.ClassInspectorUtil.JVM_NAME
 import com.squareup.kotlinpoet.metadata.classinspectors.ClassInspectorUtil.filterOutNullabilityAnnotations
 import com.squareup.kotlinpoet.metadata.hasAnnotations
@@ -421,6 +422,8 @@ public class ElementsClassInspector private constructor(
             val method = typeElement.lookupMethod(annotationsHolderSignature, ElementFilter::methodsIn)
               ?: return@let MethodData.SYNTHETIC
             annotations += method.annotationSpecs()
+              // Cover for https://github.com/square/kotlinpoet/issues/1046
+              .filterNot { it.typeName == JAVA_DEPRECATED }
           }
         }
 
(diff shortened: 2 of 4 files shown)


ZacSweers:
I think this makes a compelling case for trying to do this test in a way that exercises kapt too, like how the ksp interop tests work. Biggest issue is that some of what we generate technically doesn't compile, though we could explore making it compile. All for something later though!

PR #1170, opened 2021-10-08
Fix #1169: Handle long import with aliased name.

### LINKED ISSUE #1169: Import aliases must not be on separate lines
Aliasing imports with very long package names causes the alias to be put on a new line, which is a syntax error.

The problem looks like this:

```kotlin
import this.is.a.very.long.imported.package.with.a.function.i.need.to.rename.from.function as
  importedFunction123
```

But the generated code needs to look like this:

```kotlin
import this.is.a.very.long.imported.package.with.a.function.i.need.to.rename.from.function as importedFunction123
```

glureau:
https://github.com/square/kotlinpoet/issues/1169#issuecomment-931360858


glureau:
I think I did it, not sure how to check.

PR #1184, opened 2021-11-05
Disallow enum constants and properties called "name" and "ordinal"

### LINKED ISSUE #1183: name and ordinal enum constant names generates non-compiling code
```kotlin
    val byeNat = TypeSpec.enumBuilder("ByeNat")
      .addEnumConstant("name")
      .addEnumConstant("ordinal")
      .build()
```
This will fail because Kotlin enums already have an `name` and an `ordinal` fields.

Egorand:
Since it's a well-known limitation, perhaps it does make sense to validate it in the library.

Fixes #1183

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -651,7 +651,7 @@ public class TypeSpec private constructor(
       typeSpec: TypeSpec = anonymousClassBuilder().build()
     ): Builder = apply {
       require(name != "name" && name != "ordinal") {
-        "constant with name \"$name\" conflicts with a property with the same name"
+        "constant with name \"$name\" conflicts with a supertype member with the same name"
       }
       enumConstants[name] = typeSpec
     }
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -5015,6 +5016,42 @@ class TypeSpecTest {
     )
   }
 
+  // https://github.com/square/kotlinpoet/issues/1183
+  @Test fun `forbidden enum constant names`() {
+    var exception = assertFailsWith<IllegalArgumentException> {
+      TypeSpec.enumBuilder("Topping")
+        .addEnumConstant("name")
+    }
```
oldergod, line 5024: Interesting, the validation doesn't happen on `build` ?

oldergod (commented):
I would use the _final supertype member_ in both cases, I think it's the clearest.

oldergod (approved):

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -650,6 +650,9 @@ public class TypeSpec private constructor(
       name: String,
       typeSpec: TypeSpec = anonymousClassBuilder().build()
     ): Builder = apply {
+      require(name != "name" && name != "ordinal") {
+        "constant with name \"$name\" conflicts with a property with the same name"
```
Egorand, line 654:
```suggestion
        "constant with name \"$name\" conflicts with a supertype member with the same name"
```

PR #1189, opened 2021-11-17
Add `Unit` and `CharSequence` convert

liujingxing:
KotlinPoet Version: 1.10.2
kotlinpoet-javapoet Version: 1.10.2
Ksp Version: 1.5.31-1.0.0

```java
class Taco {
    public void add(CharSequence seasoning) { }
}
```

returnType
```kotlin
// Returns a KotlinPoet `ClassName` of value `kotlin.Unit`
val kTypeName = addKsFunction.returnType?.toTypeName()

// Returns a JavaPoet `ClassName` of value `kotlin.Unit`
val jTypeName = kTypeName?.toJTypeName()

```

paramType

```kotlin
// Returns a KotlinPoet `ClassName` of value `kotlin.CharSequence`
val kTypeName = seasoningKsValue.type.toTypeName()

// Returns a JavaPoet `ClassName` of value `kotlin.CharSequence`
val jTypeName = kTypeName.toJTypeName()
```

I expect get `void` and `java.lang.CharSequence`

### DIFF
--- a/interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/PoetInterop.kt
+++ b/interop/javapoet/src/main/kotlin/com/squareup/kotlinpoet/javapoet/PoetInterop.kt
@@ -18,6 +18,7 @@ package com.squareup.kotlinpoet.javapoet
 /** Various JavaPoet and KotlinPoet representations of some common types. */
 @OptIn(KotlinPoetJavaPoetPreview::class)
 internal object PoetInterop {
+  internal val CN_JAVA_CHAR_SEQUENCE = JClassName.get("java.lang", "CharSequence")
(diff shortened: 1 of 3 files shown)


ZacSweers (changes requested):
I agree with the charsequence interop but not void to unit. Java can use both void and unit so it's not safe to just assume they're convertible.

liujingxing:
Ok, void to unit, I can do that myself depending on the usage scenario

For example

```
fun KTypeName.toJavaTypeName(boxIfPrimitive: Boolean = false) =
    toJTypeName(boxIfPrimitive).let {
        when {
            it.toString() == "kotlin.Unit" -> JTypeName.VOID
            else -> it
        }
    }
```

ZacSweers:
Just note we can't proceed with this or until you remove the void handling

ZacSweers (approved):

PR #1191, opened 2021-11-23
Update kotlinx-metadata to 0.4.0

ZacSweers:
Supports up to kotlin 1.7

https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ChangeLog.md#040


Egorand (approved):
- Should we have any checks for supported kotlinx-metadata in the code to fail faster? E.g. if someone's running this artifact on metadata version 1.8 we'll just crash and say "sorry, not supported yet".
- Should we include kotlinx-metadata and javapoet versions in the version names of interop artifacts to make it more clear which versions of those artifacts they track? E.g. `kotlinx-metadata:1.10.2-0.4.0`? I believe KSP does this. Or will we run into issues with semver?

ZacSweers:
Maybe, I'm neutral-to-ambivelant on a different error message. For versioning no - Ksp had issues with this and is actually moving away from it

Issue #960, opened 2020-08-02
Exclude default imports from FileSpec


nielsvanvelzen:
A lot of classes inside of the `kotlin.*` package are imported automatically by Kotlin. (See [Packages and Imports](https://kotlinlang.org/docs/reference/packages.html#default-imports)).

KotlinPoet still adds import statements for these packages. This is unnecessary in most (or all?) scenarios. To reduce the generated code size the default imports could be ignored in the FileSpec class when generating the code.

Egorand:
See #361, #763

nielsvanvelzen:
Thanks for the reply. I didn't find those issues as I only searched for open issues (my bad!)

Issue #763, opened 2019-08-27
Remove explicit imports of implicitly imported stdlib packages (e.g. kotlin.String)


LouisCAD:
It seems kotlinpoet current add unneeded/noisy `kotlin.String` imports, and potentially more that are redundant in Kotlin files because they are from the stdlib in packages that are imported by default.

See the full reference here: https://kotlinlang.org/docs/reference/packages.html#default-imports

And here's an issue in an affected project: https://github.com/jmfayard/buildSrcVersions/issues/54

LouisCAD:
Isn't there a way to allow opt-out of this behavior?

Egorand:
Are there practical reasons for not wanting those imports to be emitted, or is it just the noise that bothers you?

JakeWharton:
Kotlin shouldn't have included default imports in the language.

Change. My. Mind.

Issue #361, opened 2018-04-25
Don't add default imports to generated files


gabrielhuff:
Not sure if this was already tackled, but currently any generated code will include [default imports](https://kotlinlang.org/docs/reference/packages.html#default-imports), for instance:

```kotlin
...
import kotlin.Boolean
import kotlin.Int
import kotlin.collections.Map
...
```

Can we remove them?

JakeWharton:
No. They disambiguate types of the same name which might be in scope. And since they don't hurt anything by being there, there's really no point in removing them.

lsuski:
Maybe they don't hurt much but imagine this scenario
1. Optimize imports on commit enabled
2. Generate some classes
3. Commit - default imports are removed by Intellij
4. Generate again - it seems that something changed but not - you need to manually optimize imports on root generated package to see if something really changed

I think that there could be option to disable or choose what default imports will be included

Egorand:
Normally you will not be putting generated code under source control.

JakeWharton:
And if you were that behavior seems fine because they're not being regenerated on every build so you wouldn't get constantly dirty files/

PR #1194, opened 2021-11-30
Add basic support for default imports

ZacSweers:
This implements support for specifying custom default imports in `FileSpec`. The goal here is mostly targeting script generation, but I've also added a helper that allows for opt-in inclusion of kotlin's default imports to this (i.e. they would then be _excluded_ during writing).

Open to suggestions on other tests to add!

Ref #960, #361, #763

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -167,7 +167,6 @@ public class FileSpec private constructor(
       .filterNot(defaultImportsFilter)
       .toSortedSet()
       .plus(aliasedImports.map { it.toString() }.toSortedSet())
-      .toList()
 
     if (imports.isNotEmpty()) {
       for (import in imports) {
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
@@ -52,6 +52,7 @@ public class FileSpec private constructor(
   public val packageName: String = builder.packageName
   public val name: String = builder.name
   public val members: List<Any> = builder.members.toList()
+  public val defaultImports: Set<String> = builder.defaultImports.toSet()
   public val body: CodeBlock = builder.body.build()
   public val isScript: Boolean = builder.isScript
   private val memberImports = builder.memberImports.associateBy(Import::qualifiedName)
@@ -65,7 +66,7 @@ public class FileSpec private constructor(
       NullAppendable, indent, memberImports,
       columnLimit = Integer.MAX_VALUE
     )
-    emit(importsCollector)
+    emit(importsCollector, collectingImports = true)
     val suggestedTypeImports = importsCollector.suggestedTypeImports()
     val suggestedMemberImports = importsCollector.suggestedMemberImports()
     importsCollector.close()
@@ -75,7 +76,7 @@ public class FileSpec private constructor(
       out, indent, memberImports, suggestedTypeImports,
       suggestedMemberImports
     )
-    emit(codeWriter)
+    emit(codeWriter, collectingImports = false)
     codeWriter.close()
   }
 
@@ -126,7 +127,7 @@ public class FileSpec private constructor(
     }
   }
 
-  private fun emit(codeWriter: CodeWriter) {
+  private fun emit(codeWriter: CodeWriter, collectingImports: Boolean) {
     if (comment.isNotEmpty()) {
       codeWriter.emitComment(comment)
     }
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -398,6 +413,28 @@ public class FileSpec private constructor(
...
+    public fun addDefaultPackageImport(packageName: String): Builder = apply {
+      defaultImports += packageName
+    }
+
+    /**
+     * Adds Kotlin's standard default package imports as described
+     * [here](https://kotlinlang.org/docs/packages.html#default-imports).
+     */
+    public fun addKotlinDefaultImports(
+      includeJvm: Boolean = true,
```
Egorand, line 426: Should we use "java" instead of "jvm" in naming? "jvm" might suggest it refers to "kotlin.jvm" imports.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -398,6 +413,28 @@ public class FileSpec private constructor(
       memberImports += Import(memberName.canonicalName, `as`)
     }
 
+    /** Adds a default import for the given [packageName]. */
```
Egorand, line 416: Should we include examples of supported inputs, i.e. is `kotlin.*` supported, or should it be `kotlin`? Also, should we support imports for individual types?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -147,14 +148,26 @@ public class FileSpec private constructor(
...
     val imports = (importedTypeNames + importedMemberNames)
+      .asSequence()
       .filterNot { it in memberImports.keys }
       .map { it.escapeSegmentsIfNecessary() }
-      .plus(nonAliasedImports.map { it.toString() })
+      .plus(nonAliasedImports.asSequence().map { it.toString() })
+      .filterNot(defaultImportsFilter)
       .toSortedSet()
       .plus(aliasedImports.map { it.toString() }.toSortedSet())
+      .toList()
```
Egorand, line 170: nit: Is `toList()` needed here, given that there's a `toSortedSet()` above?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -147,14 +148,26 @@ public class FileSpec private constructor(
 
     val importedTypeNames = codeWriter.importedTypes.values.map { it.canonicalName }
     val importedMemberNames = codeWriter.importedMembers.values.map { it.canonicalName }
+
+    // If we don't have default imports or are collecting them, we don't need to filter
+    var defaultImportsFilter: (String) -> Boolean = { false }
```
Egorand, line 153: nit: maybe call it `isDefaultImport`? `filterNot(isDefaultImport)` then becomes slightly easier to read.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -1131,4 +1133,27 @@ class FileSpecTest {
...
+      .addDefaultPackageImport("java.util.function")
+      .build()
+    assertThat(spec.toString()).isEqualTo(
+      """
+      |import java.util.concurrent.Callable
+      |
+      |val prop0: String? = null
+      |val prop1: Int? = null
+      |val prop2: Map<String, Any>? = null
+      |val prop3: @FunctionalInterface Callable<String>? = null
```
Egorand, line 1154: Oh wow, where do these annotations come from?

Egorand (approved):
Nice 👍

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
ZacSweers: It does, technically. These are the imports listed under the jvm subsection linked above. It includes java.lang and kotlin.jvm

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
ZacSweers: Good point, I'll check. This was automatically added by IJ when converting to a sequence

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
ZacSweers: Seeeemingly from our typeNameOf. I guess these are included on the resolved KType and we should prolly ignore them

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
ZacSweers: I thought about individual types but opted against it, I couldn't find any such examples in Kotlin, Java, or Gradle. We can add it if folks come up with one?

ZacSweers: 7159948

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
ZacSweers: 1434cac

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
ZacSweers: e2bfd77

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
Egorand: Nice, thanks, the examples are great. Agree let's keep the scope small until there's a need to support single types.

Egorand (approved):
:shipit:

PR #1195, opened 2021-12-01
Add 'yield' to the list of reserved keywords

BoD:
This is not documented, but `yield` is actually reserved too ([source](https://github.com/JetBrains/kotlin/blob/v1.6.0/compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ReservedChecking.kt#L27)).

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Util.kt
@@ -247,6 +247,9 @@ private val KEYWORDS = setOf(
   "tailrec",
   "value",
   "vararg",
+
+  // Other reserved keywords
+  "yield",
 )
 
 private const val ALLOWED_CHARACTER = '$'


Egorand (approved):
Interesting, nice find. Thanks!

Is it documented anywhere that yield is a reserved keyword? If not - would it be worth filing an issue against Kotlin?

BoD:
> Is it documented anywhere that yield is a reserved keyword? If not - would it be worth filing an issue against Kotlin?

I haven't found anything so far, but I will dig 😊.  Good idea about filing an issue, will do if nothing comes up.

Also, it's a bit odd, as you can see in [this example](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS4zLjMwIiwiY29kZSI6ImNsYXNzIHlpZWxkIHtcbiAgICBcbn1cblxuY2xhc3MgRm9vQmFyIHtcbiAgICB2YWwgeWllbGQgPSAxXG59XG5cbmVudW0gY2xhc3MgTXlFbnVtIHtcbiAgICB5aWVsZFxufVxuXG5cbmZ1biBtYWluKCkge1xuICAgIGZ1biB5aWVsZCgpID0gMVxuICAgIFxuXHR2YWwgeWllbGQgPSAyXG5cbiAgICB2YWwgZm9vID0geWllbGQoKVxuICAgIFxuICAgIHByaW50bG4oYHlpZWxkYClcbiAgICBwcmludGxuKHlpZWxkKVxufSIsInBsYXRmb3JtIjoiamF2YSIsImFyZ3MiOiIifQ==), `yield` is valid when declaring it, but invalid when referencing it. Also doesn't show up in red in IntelliJ.

BoD:
I've opened an issue [here](https://youtrack.jetbrains.com/issue/KT-50012)

Egorand:
Perfect, thank you!

PR #1196, opened 2021-12-06
Add support for `KSAnnotation.toAnnotationSpec()`

ZacSweers:
Ported from Moshi at @swankjesse's suggestion - https://github.com/square/moshi/pull/1393#discussion_r762614313

### DIFF
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/annotations.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/annotations.kt
@@ -26,12 +26,16 @@ import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.AnnotationSpec.UseSiteTarget
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.CodeBlock
+import com.squareup.kotlinpoet.ParameterizedTypeName
 
 /** Returns an [AnnotationSpec] representation of this [KSAnnotation] instance. */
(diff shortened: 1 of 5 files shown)


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/annotations.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.CodeBlock
+import com.squareup.kotlinpoet.ParameterizedTypeName
+
+/** Returns an [AnnotationSpec] representation of this [KSAnnotation] instance. */
+@KotlinPoetKspPreview
+public fun KSAnnotation.toAnnotationSpec(): AnnotationSpec {
+  val builder = when (val type = annotationType.resolve().unwrapTypeAlias().toTypeName()) {
+    is ClassName -> AnnotationSpec.builder(type)
+    is ParameterizedTypeName -> AnnotationSpec.builder(type)
```
Egorand, line 36: Would Kotlin allow merging these two branches and still get correct typecasts?

Egorand (approved):
Great tests!

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/annotations.kt
ZacSweers: Unfortunately not, we provide two overloads for each of those but no option for just the typename

PR #1214, opened 2022-02-15
Fix misleading documentation

### LINKED ISSUE #1213: Improve documentation about explicit library mode behavior
According to the documentation, this:
```
val greeterClass = ClassName("", "Greeter")
val file = FileSpec.builder("", "HelloWorld")
    .addType(TypeSpec.classBuilder("Greeter")
            .build())
        .addProperty(PropertySpec.builder("name", String::class)
            .initializer("name")
            .build())
        .addFunction(FunSpec.builder("greet")
            .addStatement("println(%P)", "Hello, \$name")
            .build())
        .build())
        .addParameter("args", String::class, VARARG)
        .addStatement("%T(args[0]).greet()", greeterClass)
        .build())
    .build()

file.writeTo(System.out)
```

should generate this:
```
class Greeter(val name: String) {
  fun greet() {
    println("""Hello, $name""")
  }
}

fun main(vararg args: String) {
  Greeter(args[0]).greet()
}
```

when in fact it generates this:
```
public class Greeter(
  public val name: String
) {
  public fun greet(): Unit {
    println("""Hello, $name""")
  }
}

public fun main(vararg args: String): Unit {
  Greeter(args[0]).greet()
}
```

so no surprise people [ask on StackOverflow](https://stackoverflow.com/questions/64914615/kotlinpoet-ommitting-redundant-public-modifier-from-generated-types-and-prope) and [create issues](https://github.com/square/kotlinpoet/issues/1001) about redundant `public` modifier [again](https://github.com/square/kotlinpoet/issues/1098) and [again](https://github.com/square/kotlinpoet/issues/1205).

micheljung:
At least for the very first example. There might be others.

Fixes #1213

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,13 +13,15 @@ the need to write boilerplate while also keeping a single source of truth for th
 Here's a `HelloWorld` file:
 
 ```kotlin
-class Greeter(val name: String) {
-  fun greet() {
+public class Greeter(
+  public val name: String
+) {
+  public fun greet(): Unit {
     println("""Hello, $name""")
   }
 }
 
-fun main(vararg args: String) {
+public fun main(vararg args: String): Unit {
   Greeter(args[0]).greet()
 }
 ```
@@ -50,6 +52,8 @@ file.writeTo(System.out)
 ```
 
 The [KDoc][kdoc] catalogs the complete KotlinPoet API, which is inspired by [JavaPoet][javapoet].
+
+Note that KotlinPoet generates code which supports explicit API mode: https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors. Therefore, redundant `public` modifiers are always generated.
 
 ### Code & Control Flow
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -53,7 +53,7 @@ file.writeTo(System.out)
 
 The [KDoc][kdoc] catalogs the complete KotlinPoet API, which is inspired by [JavaPoet][javapoet].
 
-Note that KotlinPoet generates code which supports explicit API mode: https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors. Therefore, redundant `public` modifiers are always generated.
+Note that KotlinPoet generates code which supports [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors). Therefore, redundant `public` modifiers are always generated.
 
 ### Code & Control Flow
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -51,6 +53,8 @@ file.writeTo(System.out)
 
 The [KDoc][kdoc] catalogs the complete KotlinPoet API, which is inspired by [JavaPoet][javapoet].
 
+Note that KotlinPoet generates code which supports explicit API mode: https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors. Therefore, redundant `public` modifiers are always generated.
```
micheljung, line 56:
```suggestion
Note that KotlinPoet generates code which supports [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors): . Therefore, redundant `public` modifiers are always generated.
```

micheljung:
I can squash this later. Can't use Git right now

docs/index.md
```diff
@@ -51,6 +53,8 @@ file.writeTo(System.out)
 
 The [KDoc][kdoc] catalogs the complete KotlinPoet API, which is inspired by [JavaPoet][javapoet].
 
+Note that KotlinPoet generates code which supports [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors). Therefore, redundant `public` modifiers are always generated.
```
ZacSweers, line 56:
```suggestion
Note that KotlinPoet generates code which supports [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors). Therefore, explicit `public` modifiers are always generated.
```

ZacSweers (approved):
One nit

ZacSweers:
There are actually multiple others. Perhaps it's better to just add a note about explicit API mode rather than have inconsistent usage in the examples.

micheljung:
So would you rather have examples that are inconsistent with what kotlinpoet actually does? Even if we don't catch them all now, I'd rather work towards a correct documentation

Egorand:
It'd be nice to have a solution to store these snippets in source files, execute them at every CI build and include the outputs into the docs.

ZacSweers:
Knit is supposed to be useful for that

ZacSweers:
We're going to explore Knit separately. We'll also better document the explicit library mode behavior. If you're open to feedback - in the future, PRs and issues like this would be better received if every comment isn't punctuated with a complaint or needling observation. It's ok to disagree with a design feature and want improved documentation, but please keep it constructive when offering feedback. We are all just trying to build something good :)

PR #1222, opened 2022-03-01
Fail compilation if you only pass one string to ClassName

### LINKED ISSUE #1221: Consider error-deprecated `ClassName(String)` overload
Sometimes I accidentally forget to split the simple name and write `ClassName("com.example.Foo")`. This fails at runtime.
```
Caused by: java.lang.IllegalArgumentException: simpleNames must not be empty
        at com.squareup.kotlinpoet.ClassName.<init>(TypeName.kt:323)
```

It would be nice if it failed at compile-time. This can be done with something like
```kotlin
@Deprecated("Simple names must not be empty", level = ERROR)
fun ClassName(value: String): Nothing = throw IAE(..)
```

JakeWharton:
Tried to test, but even with `@Suppress("DEPRECATION_ERROR")` the compiler spit out the error.

Closes #1221

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ClassName.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/ClassName.kt
@@ -40,6 +40,13 @@ public class ClassName internal constructor(
   public constructor(packageName: String, simpleName: String, vararg simpleNames: String) :
     this(listOf(packageName, simpleName, *simpleNames))
 
+  @Deprecated(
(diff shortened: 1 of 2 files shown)


JakeWharton:
Oh it's an error on some other test. Hey it's working!

ZacSweers (approved):

Egorand:
This is Kotlin source-incompatible, right?

JakeWharton:
Yeah kinda. Your source code will still properly resolve to the same thing, but your compilation will now fail.

Issue #1234, opened 2022-04-14
Enum constants are not taken into account when resolving conflicts


martinbonnin:
The following code:

```kotlin
    val sb = StringBuilder()
    FileSpec.builder("com.example", "test")
        .addType(
            TypeSpec.enumBuilder("Foo")
                .addProperty(
                    PropertySpec.builder("rawValue", String::class)
                        .initializer("%S", "")
                        .build()
                )
                // adding a nested class resolves the conflict correctly
//                .addType(
//                    TypeSpec.classBuilder("String")
//                        .build()
//                )
                // but not with enum constants
                .addEnumConstant("String")
                .build()
        )
        .build()
        .writeTo(sb)

    println(sb.toString())
```

produces:

```kotlin
public enum class Foo {
  String,
  ;

  public val rawValue: String = ""
}
```

Which doesn't compile because `rawValue` should be a `kotlin.String` instead.

PR #1235, opened 2022-04-14
Resolve enum constants when emitting types

martinbonnin:
Closes https://github.com/square/kotlinpoet/issues/1234

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -403,7 +403,7 @@ internal class CodeWriter constructor(
     while (c != null) {
       val alias = memberImports[c.canonicalName]?.alias
       val simpleName = alias ?: c.simpleName
-      val resolved = resolve(simpleName)
+      val resolved = resolveClass(simpleName) ?: resolveEnumConstant(simpleName)
       nameResolved = resolved != null
 
       // We don't care about nullability and type annotations here, as it's irrelevant for imports.
@@ -509,7 +509,7 @@ internal class CodeWriter constructor(
    * imports.
    */
   // TODO(jwilson): also honor superclass members when resolving names.
-  private fun resolve(simpleName: String): ClassName? {
+  private fun resolveClass(simpleName: String): ClassName? {
     // Match a child of the current (potentially nested) class.
     for (i in typeSpecStack.indices.reversed()) {
       val typeSpec = typeSpecStack[i]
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -531,6 +531,20 @@ internal class CodeWriter constructor(
     return null
   }
 
+  /**
+   * Returns the qualified name of the `simpleName` enum constant if it's in the current scope
+   */
+  private fun resolveEnumConstant(simpleName: String): String? {
```
martinbonnin, line 537: I made this into a separate function because I wasn't sure if a `ClassName` could represent an enum constant.

Egorand: It's odd that this method returns a different type, hence the check on line 410 can now fail because of the type check, and not logical inequality. We don't represent enum constants as `ClassName`s, so not sure what the outcome will be if we tried.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -531,6 +531,20 @@ internal class CodeWriter constructor(
...
+    if (typeSpecStack.size > 0) {
+      val typeSpec = typeSpecStack[0]
+      if (typeSpec.isEnum && typeSpec.enumConstants.keys.contains(simpleName)) {
+        return "$packageName.${typeSpec.name}.$simpleName"
```
Egorand, line 541: This is technically compatible with the structure of `ClassName`, would there be any undesired side effects (e.g. unwanted imports) if we returned `ClassName` here?

martinbonnin: I think that'd work. Especially since `resolve()` is very local to this class the risk of side effect seems pretty limited. I'll try this

martinbonnin: Code updated. Let me know what you think

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -518,9 +518,17 @@ internal class CodeWriter constructor(
...
-      return ClassName(packageName, simpleName)
+    if (typeSpecStack.size > 0) {
+      val typeSpec = typeSpecStack[0]
+      if (typeSpec.name == simpleName) {
+        // Match the top-level class.
+        return ClassName(packageName, simpleName)
+      }
+      if (typeSpec.isEnum && typeSpec.enumConstants.keys.contains(simpleName)) {
+        // Match a top level enum value
+        // Enum values are not proper classes but can still be modeled using ClassName
```
Egorand, line 529: supernits:
```suggestion
        // Match a top level enum value.
        // Enum values are not proper classes but can still be modeled using ClassName.
```

martinbonnin:
> Any other test cases you can think of that might lead to name clashes?

TBH I would have never thought of that without [this issue](https://github.com/apollographql/apollo-kotlin/issues/4023). I'll make sure to add new tests if some other cases pop up :)

Egorand:
Sounds good, thank you!

PR #1237, opened 2022-04-14
Add Java 18 to CI build matrix

Egorand:
Is it supported?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,7 @@ jobs:
           - 15
           - 16
           - 17
+          - 18
 
     steps:
       - name: Checkout


JakeWharton:
Note that this is the wrong way to test multiple versions and we should migrate to building on a single version and merely testing on multiple: https://github.com/square/retrofit/pull/3713.

JakeWharton (approved):

Egorand:
Makes sense, can fix it up in a followup!

Egorand:
@JakeWharton the downside of this solution is that I can't really run `gw clean build` locally anymore as it'll start pulling tons of JDKs. Is there a workaround?

JakeWharton:
We could remove them from the check task and instead use a custom aggregating task that's only run on CI or explicitly locally.

JakeWharton:
Also worth noting those JDKs that are downloaded are only done once (and only when missing on the system). So once you pay the cost it'll be back to being extremely fast.

Egorand:
I'm getting toolchain download errors, seemingly cause it can't find JDKs of specific versions for M1. I think I'd prefer a separate task for the CI, that can also be called explicitly on a dev machine if needed - I'll send a PR.

Issue #1244, opened 2022-04-28
Consider adding support for context receivers in PropertySpec


seriouslyhypersonic:
I opened [this](https://github.com/square/kotlinpoet/issues/1243) regarding support for context receivers yesterday and received the reply that they got supported with [#1232](https://github.com/square/kotlinpoet/issues/1232). I checked out `1.12.0-SNAPSHOT` but it seems there is currently no support for context receivers on `PropertySpec`, which is exactly the use case I have for the library I am writing 😅. Please consider adding support for this use case.

ZacSweers:
The spec says

> Property getters and setters

Which are implemented as `FunSpec`s on `PropertySpec`.

seriouslyhypersonic:
Makes sense of course! I was just too naïve and thought they would be implemented as another property of the property spec builder. Silly me.

seriouslyhypersonic:
Sorry to reopen, but I tried adding context receivers via getters/setters and it doesn't seem to work properly. The context is declared on the actual getter/setter and not on the property. Example:

```kotlin
PropertySpec
    .builder(name = "loudGreeting", String::class.asTypeName())
    .getter(
        FunSpec
            .getterBuilder()
            .contextReceivers(String::class.asTypeName())
            .addStatement("return \"Hello, \${uppercase()}\"")
            .build()
    )
    .build()
```

Yields:
```kotlin
public val loudGreeting: String
      context(String) // Error
      get() = "Hello, ${uppercase()}"
```

While I believe it should yield this instead:
```kotlin
context(String)
public val loudGreeting: String
      get() = "Hello, ${uppercase()}"
```

This is a very simple example which could surely have been simply solved as an extension property on `String`. My real use case is more complex because the property I am declaring is already an extension, but that extension must only be available in a specific context (with more than one receiver).

In fact, I believe the context declaration (if present) must be the first statement on the property declaration (including annotations). For example, lets assume we have:

```kotlin
annotation class MyFancyAnnotation
```

This:
```kotlin
PropertySpec
    .builder(name = "loudGreeting", String::class.asTypeName())
    .addAnnotation(MyFancyAnnotation::class.asTypeName())
    .getter(/*...*/)
    .build
```

Should yield:
```kotlin
context(String)
@MyFancyAnnotation
public val loudGreeting: String
    get() = "Hello, ${uppercase()}"
```

seriouslyhypersonic:
[...]

PR #1247, opened 2022-05-07
Add support for context receivers @PropertySpec and fix issues with annotations

seriouslyhypersonic:
# Summary
As discussed in #1244, here is a PR to add support for context receivers @`PropertySpec`.

# Proposed Changes
1. Make `contextReceivers` available in `PropertySpec` builder
2. Throw exceptions when: property is mutable or immutable without a getter
3. Update `FunSpec` to throw exception when a context receiver is added to an accessor
4. Fix order in which context receivers and annotations are emitted: context receivers must come first

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -60,7 +60,7 @@ public class PropertySpec private constructor(
     }
     if (contextReceiverTypes.isNotEmpty()) {
       check(!mutable) { "mutable properties cannot have context receivers" }
-      require(getter != null){ "immutable properties with context receivers require a $GETTER" }
+      require(getter != null) { "immutable properties with context receivers require a $GETTER" }
     }
   }
 
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -59,11 +59,10 @@ public class PropertySpec private constructor(
       "only a mutable property can have a setter"
     }
     if (contextReceiverTypes.isNotEmpty()) {
+      requireNotNull(getter) { "properties with context receivers require a $GETTER" }
       if (mutable) {
-        requireNotNull(getter) { "mutable properties with context receivers require a $GETTER" }
         requireNotNull(setter) { "mutable properties with context receivers require a $SETTER" }
       }
-      requireNotNull(getter) { "immutable properties with context receivers require a $GETTER" }
     }
   }
 
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -59,8 +59,11 @@ public class PropertySpec private constructor(
       "only a mutable property can have a setter"
     }
     if (contextReceiverTypes.isNotEmpty()) {
-      check(!mutable) { "mutable properties cannot have context receivers" }
-      require(getter != null) { "immutable properties with context receivers require a $GETTER" }
+      if (mutable) {
+        requireNotNull(getter) { "mutable properties with context receivers require a $GETTER" }
+        requireNotNull(setter) { "mutable properties with context receivers require a $SETTER" }
+      }
+      requireNotNull(getter) { "immutable properties with context receivers require a $GETTER" }
     }
   }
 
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -54,6 +58,10 @@ public class PropertySpec private constructor(
     require(mutable || setter == null) {
       "only a mutable property can have a setter"
     }
+    if (contextReceiverTypes.isNotEmpty()) {
+      check(!mutable) { "mutable properties cannot have context receivers" }
```
Egorand, line 62: Mutable properties _can_ have context receivers, the snippet below compiles:

```kotlin
context(LoggingContext)
var foo: String
  get() = ""
  set(value) {}
```

The rule here seems to be:

- If the property is immutable - it has to have a custom getter
- If the property is mutable - it has to have custom accessors

Also, let's use `require` everywhere for consistency (unless there's a specific reason to use `check`?).

kotlinpoet/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -559,4 +563,67 @@ class PropertySpecTest {
       """.trimIndent()
     )
   }
+
+  @Test fun varWithcontextReceiver() {
```
Egorand, line 567: nit:
```suggestion
  @Test fun varWithContextReceiver() {
```

kotlinpoet/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
```diff
@@ -559,4 +563,67 @@ class PropertySpecTest {
...
+    assertThrows<IllegalStateException> {
+      PropertySpec.builder("foo", STRING)
+        .mutable()
+        .contextReceivers(INT)
+        .build()
+    }.hasMessageThat().isEqualTo("mutable properties cannot have context receivers")
+  }
+
+  @Test fun valWithoutGetterWithcontextReceiver() {
```
Egorand, line 576: nit:
```suggestion
  @Test fun valWithoutGetterWithContextReceiver() {
```

Egorand (changes requested):

kotlinpoet/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
seriouslyhypersonic: Fixed

kotlinpoet/src/test/java/com/squareup/kotlinpoet/PropertySpecTest.kt
seriouslyhypersonic: Fixed

kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -54,6 +58,13 @@ public class PropertySpec private constructor(
     require(mutable || setter == null) {
       "only a mutable property can have a setter"
     }
+    if (contextReceiverTypes.isNotEmpty()) {
+      if (mutable) {
+        requireNotNull(getter) { "mutable properties with context receivers require a $GETTER" }
+        requireNotNull(setter) { "mutable properties with context receivers require a $SETTER" }
+      }
+      requireNotNull(getter) { "immutable properties with context receivers require a $GETTER" }
```
Egorand, line 66: This feels a bit cleaner and avoids duplication:
```suggestion
      requireNotNull(getter) { "properties with context receivers require a $GETTER" }
      if (mutable) {
        requireNotNull(setter) { "mutable properties with context receivers require a $SETTER" }
      }
```

Egorand (approved):
Looks good! Added a single suggestion, otherwise good to merge - thanks!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
Egorand: Note that this change will require tests to be updated as well.

seriouslyhypersonic:
@Egorand will the snapshot repository be updated with this? I really need this fix to publish my library ☺️

Egorand:
Yes! As soon as it's merged CI will publish a fresh snapshot.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
seriouslyhypersonic: > Note that this change will require tests to be updated as well.

Which is obviously something I forgot. Updated them just now.

seriouslyhypersonic:
@Egorand Please recheck as I just accepted your suggestion earlier today and had to update the tests right now. Thank you!

Egorand:
Merging, thank you!

PR #1249, opened 2022-05-13
use %N instead of %L for annotation arg names so keywords are handled

dkilmer:


ZacSweers (approved):

ZacSweers:
Thanks!

Egorand:
Is it possible to add a test for this change? That'd be great.

Issue #1253, opened 2022-05-20
incorrect syntax  for generate enum


dark0ghost:
Kotlin poet (**version: 1.9.0**) generate incorrect enum (code for generte from https://square.github.io/kotlinpoet/#enums)
```kotlin
val helloWorld = TypeSpec.enumBuilder("Roshambo")
  .primaryConstructor(
    FunSpec.constructorBuilder()
      .addParameter("handsign", String::class)
      .build()
  )
  .addEnumConstant(
    "ROCK", TypeSpec.anonymousClassBuilder()
      .addSuperclassConstructorParameter("%S", "fist")
      .addFunction(
        FunSpec.builder("toString")
          .addModifiers(KModifier.OVERRIDE)
          .addStatement("return %S", "avalanche!")
          .returns(String::class)
          .build()
      )
      .build()
  )
  .addEnumConstant(
    "SCISSORS", TypeSpec.anonymousClassBuilder()
      .addSuperclassConstructorParameter("%S", "peace")
      .build()
  )
  .addEnumConstant(
    "PAPER", TypeSpec.anonymousClassBuilder()
      .addSuperclassConstructorParameter("%S", "flat")
      .build()
  )
  .addProperty(
    PropertySpec.builder("handsign", String::class, KModifier.PRIVATE)
      .initializer("handsign")
      .build()
  )
  .build()
```

## Expect:
```kotlin
public enum class Roshambo(
  private val handsign: kotlin.String
) {
  ROCK("fist") {
    public override fun toString(): kotlin.String = "avalanche!"
  },
  SCISSORS("peace"),
  PAPER("flat");
}
```

## Actual:

```kotlin
enum class Roshambo(private val handsign: String) {
  ROCK("fist") {
    override fun toString(): String = "avalanche!"
  },

  SCISSORS("peace"),

  PAPER("flat");
}
```

JakeWharton:
I'm not seeing what's incorrect. Looks fine.

dark0ghost:
```kotlin
PAPER("flat"),
  ;
``` 
When we except ```PAPER("flat");```

JakeWharton:
This style is intentional and ensures that when a new enum entry is added or removed that only a single line of code changes (if you happen to be committing generated source or viewing generated source diffs).

PR #1254, opened 2022-05-21
refactoring doc with new algorithm generation enum class

dark0ghost:
#1253 

This change needs to be added because the old version of the documentation is misleading to developers when write test for generated code and expect one code, but lib generate other code different from the one shown in the example

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1188,7 +1188,8 @@ enum class Roshambo(private val handsign: String) {
 
   SCISSORS("peace"),
 
-  PAPER("flat");
+  PAPER("flat"),
+  ;
 }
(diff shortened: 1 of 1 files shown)


JakeWharton:
The documentation reflects the spirit of what is generated. We also don't bother showing the public modifiers or exact line wrapping. The format of the generated code can and will change over time and relying on a small example for the canonical format is unwise.

I would prefer the documentation stay as terse as possible to convey the intent of what happens rather than being 100% accurate. If you are writing test cases, I would suggest copying what is generated to get your golden version. If you're curious if something is intentional or not, check our test cases or start a discussion.

PR #1256, opened 2022-05-29
Fix a bug caused by too long return expressions

### LINKED ISSUE #1252: kotlinpoet will fail if it returns an expression that is too long
When the generated `return someExpression()` expression is too long (over 100), it will cause kotlinpoet to automatically insert a newline and the IDE check will pop up a `[RETURN_TYPE_MISMATCH]` error and the compilation will fail.

And when I tried to change generate `columnLimit`, I found it very difficult to change it because `CodeWriter` is an internal class.

image

image

My SymbolProcessor and FileSpec generation code is as follows, which simply generates a long return expression:

```kotlin
@AutoService(SymbolProcessorProvider::class)
class SimpleKspProcessor : SymbolProcessorProvider {
    override fun create(environment: SymbolProcessorEnvironment): SymbolProcessor {
        return object : SymbolProcessor {
            override fun process(resolver: Resolver): List<KSAnnotated> {
                val allSeq = resolver.getSymbolsWithAnnotation(Anno::class.java.canonicalName)
                val annotated = allSeq.firstOrNull() ?: return emptyList()
                annotated.accept(object : KSVisitorVoid() {
                    override fun visitFunctionDeclaration(function: KSFunctionDeclaration, data: Unit) {
                        val fileSpec = longReturn(function.packageName.asString())
                        runCatching {
                            val file = environment.codeGenerator.createNewFile(Dependencies.ALL_FILES, fileSpec.packageName, fileSpec.name)
                            OutputStreamWriter(file, StandardCharsets.UTF_8).use(fileSpec::writeTo)
                        }
                    }
                }, Unit)
                return listOf(annotated)
            }
        }
    }
}

fun longReturn(packageName: String): FileSpec {
    val myLongClass = "Qwertyuiopasdfghjklzxcvbnmqwertyuiopasdfghjklzxcvbnmqwertyuiopasdfghjklzxcvbnmqwertyuiopasdfghjklzxcvbnm"
    val simpleClass = TypeSpec.classBuilder("KspGen")
    simpleClass.addFunction(
        FunSpec.builder("ccc")
            .addModifiers(KModifier.PUBLIC)
            .returns(ClassName.bestGuess("org.example.$myLongClass"))
            .addStatement("val a = 888")
            .addStatement("return $myLongClass()")
            .build()
    )
    return FileSpec.builder(packageName, "KspGen")
        .addType(simpleClass.build())
        .build()
}
```

---
versions:
- kotlin: 1.6.20
- kotlinpoet: 1.11.0
- ksp: 1.6.20-1.0.5

My test code is here: [zsqw123/KotlinPoetIssue](https://github.com/zsqw123/KotlinPoetIssue)

zsqw123:
Fix #1252 @Egorand

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -458,8 +458,6 @@ public class CodeBlock private constructor(
     private const val TYPE_NAME = 2
     private val NO_ARG_PLACEHOLDERS = setOf("⇥", "⇤", "«", "»")
     internal val EMPTY = CodeBlock(emptyList(), emptyList())
-    internal const val RETURN_WITH_SPACE = "return "
-    internal const val RETURN_WITH_NBSP = "return·"
 
     @JvmStatic public fun of(format: String, vararg args: Any?): CodeBlock =
       Builder().add(format, *args).build()
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -270,11 +270,7 @@ public class CodeBlock private constructor(
         if (format[p] != '%') {
           var nextP = format.nextPotentialPlaceholderPosition(startIndex = p + 1)
           if (nextP == -1) nextP = format.length
-          var formatPart = format.substring(p, nextP)
-          if (formatPart.startsWith(RETURN_WITH_SPACE)) {
-            formatPart = formatPart.replaceRange(RETURN_WITH_SPACE.indices, RETURN_WITH_NBSP)
-          }
-          formatParts += formatPart
+          formatParts += format.substring(p, nextP)
           p = nextP
           continue
         }
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -270,7 +270,11 @@ public class CodeBlock private constructor(
         if (format[p] != '%') {
           var nextP = format.nextPotentialPlaceholderPosition(startIndex = p + 1)
           if (nextP == -1) nextP = format.length
-          formatParts += format.substring(p, nextP)
+          var formatPart = format.substring(p, nextP)
+          if (formatPart.startsWith(RETURN_WITH_SPACE)) {
+            formatPart = formatPart.replaceRange(RETURN_WITH_SPACE.indices, RETURN_WITH_NBSP)
+          }
+          formatParts += formatPart
           p = nextP
           continue
         }
@@ -458,6 +462,8 @@ public class CodeBlock private constructor(
     private const val TYPE_NAME = 2
     private val NO_ARG_PLACEHOLDERS = setOf("⇥", "⇤", "«", "»")
     internal val EMPTY = CodeBlock(emptyList(), emptyList())
+    internal const val RETURN_WITH_SPACE = "return "
+    internal const val RETURN_WITH_NBSP = "return·"
 
     @JvmStatic public fun of(format: String, vararg args: Any?): CodeBlock =
       Builder().add(format, *args).build()
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -247,19 +247,13 @@ public class FunSpec private constructor(
 
   private fun CodeBlock.returnsWithoutLinebreak(): CodeBlock {
     val originCodeBlockBuilder = toBuilder()
-    val returnWithSpace = CodeBlock.RETURN_WITH_SPACE
+    val returnWithSpace = "return "
+    val returnWithNbsp = "return·"
     originCodeBlockBuilder.formatParts.clear()
     formatParts.mapTo(originCodeBlockBuilder.formatParts) { formatPart ->
-      if (formatPart.isEmpty()) return@mapTo formatPart
-      var startReturnIndex = 0
-      while (startReturnIndex < formatPart.length && formatPart[startReturnIndex] == ' ') startReturnIndex++
-      var endIndex = startReturnIndex + returnWithSpace.length
-      if (endIndex > formatPart.length) return@mapTo formatPart
-      if (formatPart.substring(startReturnIndex, endIndex) != returnWithSpace) {
-        return@mapTo formatPart
-      }
-      while (endIndex < formatPart.length && formatPart[endIndex] == ' ') endIndex++
-      CodeBlock.RETURN_WITH_NBSP + formatPart.substring(endIndex, formatPart.length)
+      if (formatPart.startsWith(returnWithSpace)) {
+        formatPart.replaceFirst(returnWithSpace, returnWithNbsp)
+      } else formatPart
     }
     return originCodeBlockBuilder.build()
   }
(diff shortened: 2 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -270,7 +270,11 @@ public class CodeBlock private constructor(
         if (format[p] != '%') {
           var nextP = format.nextPotentialPlaceholderPosition(startIndex = p + 1)
           if (nextP == -1) nextP = format.length
-          formatParts += format.substring(p, nextP)
+          var formatPart = format.substring(p, nextP)
```
Egorand, line 273: I think this is too generic, we should probably only fixup function bodies. I believe this code belongs somewhere in `FunSpec`, likely `emit()`.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -458,6 +458,8 @@ public class CodeBlock private constructor(
     private const val TYPE_NAME = 2
     private val NO_ARG_PLACEHOLDERS = setOf("⇥", "⇤", "«", "»")
     internal val EMPTY = CodeBlock(emptyList(), emptyList())
+    internal const val RETURN_WITH_SPACE = "return "
```
Egorand, line 461: Revert this change?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -206,6 +206,62 @@ class FunSpecTest {
...
+      """
+      |public fun foo(): kotlin.String {
+      |  val placeholder = 1
+      |  return "Loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong"
+      |}
+      |""".trimMargin()
+    )
+  }
+
+  @Test fun returnsLongExpressions() {
```
Egorand, line 234: This test does not seem to add any value over the previous one, there should be existing tests that would catch a regression for a case where the substitution is not needed. I suggest deleting this test.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -245,6 +245,25 @@ public class FunSpec private constructor(
     return null
   }
 
+  private fun CodeBlock.returnsWithoutLinebreak(): CodeBlock {
+    val originCodeBlockBuilder = toBuilder()
+    val returnWithSpace = CodeBlock.RETURN_WITH_SPACE
+    originCodeBlockBuilder.formatParts.clear()
+    formatParts.mapTo(originCodeBlockBuilder.formatParts) { formatPart ->
```
Egorand, line 252: TBH I have a hard time understanding what this block does, can it be simplified? Essentially what we want to do is (in pseudocode):

```
if (body.startsWith(RETURN_EXPRESSION_BODY_PREFIX_SPACE)) {
  body = body.replace(RETURN_EXPRESSION_BODY_PREFIX_SPACE, RETURN_EXPRESSION_BODY_PREFIX_NBSP)
}
```

Helper methods inside `CodeBlock`, such as `withoutPrefix()`, `trim()`, `replaceAll()` can help implement that.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -206,6 +206,62 @@ class FunSpecTest {
...
   }
 
+  @Test fun returnsLongExpression() {
+    val funSpec = FunSpec.builder("foo")
+      .returns(String::class)
+      .addStatement("val placeholder = 1")
+      .addStatement("return     \"Loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong\"")
+      .build()
+    val sb = StringBuilder()
+    CodeWriter(sb).use {
```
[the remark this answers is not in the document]
zsqw123: The `toString` metho's columnLimit is Integer.MAX_VALUE, This does not reproduce my case

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
JakeWharton: We don't need to support trimming. If the person is generating code which has leading spaces or multiple spaces between the actual return expression then they can fix that.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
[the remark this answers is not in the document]
Egorand: Ah, good point. Might be worth adding a short comment to the code explaining why `CodeWriter` is used.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -245,6 +245,19 @@ public class FunSpec private constructor(
     return null
   }
 
+  private fun CodeBlock.returnsWithoutLinebreak(): CodeBlock {
+    val originCodeBlockBuilder = toBuilder()
+    val returnWithSpace = "return "
+    val returnWithNbsp = "return·"
```
Egorand, line 251:
```suggestion
    val returnWithSpace = RETURN_EXPRESSION_BODY_PREFIX_SPACE
    val returnWithNbsp = RETURN_EXPRESSION_BODY_PREFIX_NBSP
```

Egorand: And I would actually inline these variables.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -245,6 +245,19 @@ public class FunSpec private constructor(
     return null
   }
 
+  private fun CodeBlock.returnsWithoutLinebreak(): CodeBlock {
+    val originCodeBlockBuilder = toBuilder()
```
Egorand, line 249: Is there a way to optimize the code to only call `toBuilder()` if there is something to replace?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -206,6 +206,31 @@ class FunSpecTest {
     )
   }
 
+  @Test fun returnsLongExpression() {
+    val funSpec = FunSpec.builder("foo")
+      .returns(String::class)
+      .addStatement("val placeholder = 1")
+      .addStatement("return \"Loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong\"")
```
Egorand, line 213: This is the preferred form, should not affect the output:
```suggestion
      .addStatement("return %S", "Loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong")
```

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -245,6 +245,19 @@ public class FunSpec private constructor(
...
+    val returnWithSpace = "return "
+    val returnWithNbsp = "return·"
+    originCodeBlockBuilder.formatParts.clear()
+    formatParts.mapTo(originCodeBlockBuilder.formatParts) { formatPart ->
+      if (formatPart.startsWith(returnWithSpace)) {
+        formatPart.replaceFirst(returnWithSpace, returnWithNbsp)
+      } else formatPart
```
Egorand, line 256: nit: would be nice to use braces with the `else` statement too

Egorand (approved):
Looks great, thanks! I just added a few suggestions on the style.

Also, please sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) before I can merge your PR.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
zsqw123: I think so, but I'm ready to go to bed now. I'll change it tomorrow morning (I'm at UTC+8 time zone).

zsqw123: I've changed it so that the `toBuilder` is now called on demand

zsqw123: However, many functions actually have a return method, this optimization may not have a significant effect

Egorand: Yeah, good point, the code looks good though, so let's leave it as is!

Egorand:
Merging, thanks again!

PR #1262, opened 2022-06-12
Update to Kotlin 1.7 + other deps

ZacSweers:
KCT is inexplicably built against a dev build of kotlin and snapshot of KSP, so we have to force newer versions to work around both. We should probably consider alternatives in the future as it's not well maintained


gradle.properties
```diff
@@ -14,3 +14,6 @@ POM_LICENCE_DIST=repo
 
 POM_DEVELOPER_ID=square
 POM_DEVELOPER_NAME=Square, Inc.
+
+SONATYPE_HOST=DEFAULT
```
Egorand, line 18:
```suggestion
SONATYPE_HOST=S01
```

Egorand (approved):

PR #1270, opened 2022-06-17
Optimize holder interfaces

ZacSweers:
Started with #1269, but a new feature of Kotlin 1.7 is we can inline these holder classes to make them not allocate. https://kotlinlang.org/docs/whatsnew17.html#allow-implementation-by-delegation-to-an-inlined-value-of-an-inline-class

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Taggable.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/Taggable.kt
@@ -27,7 +27,10 @@ public interface Taggable {
   public fun <T : Any> tag(type: Class<T>): T? = tag(type.kotlin)
 
   /** Returns the tag attached with [type] as a key, or null if no tag is attached with that key. */
-  public fun <T : Any> tag(type: KClass<T>): T?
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/Taggable.kt
```diff
@@ -145,13 +148,11 @@ public inline fun <reified T : Any> TypeAliasSpec.Builder.tag(tag: T?): TypeAlia
 public inline fun <reified T : Any> TypeSpec.Builder.tag(tag: T?): TypeSpec.Builder =
   tag(T::class, tag)
 
-internal fun Taggable.Builder<*>.buildTagMap(): TagMap = TagMap(LinkedHashMap(tags)) // Defensive copy
+internal fun Taggable.Builder<*>.buildTagMap(): TagMap = TagMap(tags.toImmutableMap())
```
Egorand, line 151: Not need since the factory method already does it?
```suggestion
internal fun Taggable.Builder<*>.buildTagMap(): TagMap = TagMap(tags)
```

Egorand (approved):

PR #1306, opened 2022-07-03
Fix transitive aliases

bitPogo:
As the test case describes - transitive Aliases are break (since KSP 1.7.0-1.6.0, but not in prior versions). This aims to resolve the issue by cascading the `extraResolver` for Aliases and use it also for arguments of Aliases.


ZacSweers (approved):
Thanks for adding a thorough test!

Egorand:
Yeah, that's a requirement, sorry.

PR #1307, opened 2022-07-03
Add possibility to use kdoc for file comment header

sokomishalov:
Sometimes it's better to use kdoc-like comment for file headers (i.e. license headers)


JakeWharton:
I don't see a reason to ever use kdoc style. Multi-line? Sure. But I'm not sure why you would ever explicitly need it.

JakeWharton:
In any case: we decided we don't want this already. https://github.com/square/kotlinpoet/pull/533

ZacSweers:
`/** */` block comments also aren't valid as file headers as they would be considered dangling/floating kdocs that are not attached to any declaration. If it were ever supported, it would have to be `/* */` style

PR #1308, opened 2022-07-03
Add possibility to specify parameters size for line wrap

sokomishalov:


ZacSweers:
We're not looking to make this configurable. If you want more granular formatting, consider post-processing kotlinpoet outputs with the formatter of your choice (ktlint, ktfmt, etc). Also, in future PRs, please abide by our codebase's formatting rather than deviating (star imports, etc)

PR #1310, opened 2022-07-03
Remove LambdaTypeName from TypeName#findRawType

bitPogo:
Currently LambdaTypeName is referenced in TypeName#findRawType (which is only invoked for Aliases).
However  this is dead code which you can see if you run [this test](https://github.com/square/kotlinpoet/blob/81679e2643caa042adcc9cd7d2db39386977379b/interop/ksp/test-processor/src/test/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessorTest.kt#L395) with coverage. Additionally it references Java instead of Kotlin, which can break KMP cases. Therefore it can be removed.

### DIFF
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/utils.kt
@@ -18,7 +18,6 @@ package com.squareup.kotlinpoet.ksp
 import com.google.devtools.ksp.isLocal
 import com.google.devtools.ksp.symbol.KSDeclaration
 import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.LambdaTypeName
 import com.squareup.kotlinpoet.ParameterizedTypeName
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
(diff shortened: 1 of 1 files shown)


ZacSweers:
The fact that test code today doesn't cover it doesn't mean it's dead code. The real indicator would be if we every create one in other KSP interop APIs

bitPogo:
I actually ported the interop layer myself to adapt to my usecase which you can find [here](https://sonarcloud.io/code?id=kmock&selected=kmock%3Akmock-processor%2Fsrc%2Fmain%2Fkotlin%2Ftech%2Fantibytes%2Fkmock%2Fprocessor%2Fkotlinpoet%2FShared.kt). I actually have not a small number of test but I was never able to trigger it, since there currently no case which transforms it into LambdaTypeName (this would need to happen [here](https://github.com/square/kotlinpoet/blob/81679e2643caa042adcc9cd7d2db39386977379b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt#L57)), I think. I can port 'my' version to here where I already made the change we discussed on Slack or  we sticking with removing it until the usecase emerges - what do you think?

Egorand:
Agree with Zac that we shouldn't be removing this code, I'll close the PR.

PR #1318, opened 2022-07-12
fix #1273 

### LINKED ISSUE #1273: Supporting TypeMirror breaks multiplatform KSP usage
```
buildCodeBlock {
        add("return %T(", Whatever::class)
}
```

Even though this uses KClass, I get this error when running e.g. `kspKotlinIosArm64`
```
        java.lang.NoClassDefFoundError: javax/lang/model/type/TypeMirror
        at com.squareup.kotlinpoet.CodeBlock$Builder.argToType(CodeBlock.kt:377)
        at com.squareup.kotlinpoet.CodeBlock$Builder.addArgument(CodeBlock.kt:346)
        at com.squareup.kotlinpoet.CodeBlock$Builder.add(CodeBlock.kt:318)
```

This [PR](https://github.com/square/kotlinpoet/pull/919) deprecated support for TypeMirror in 2020, maybe it's time to drop it? If not, maybe this piece of code can be refactored so that `TypeName` isn't evaluated if not necessary?

```
    private fun argToType(o: Any?) = when (o) {
      is TypeName -> o
      is TypeMirror -> {
        logDeprecationWarning(o)
        o.asTypeName()
      }
      is Element -> {
        logDeprecationWarning(o)
        o.asType().asTypeName()
      }
      is Type -> o.asTypeName()
      is KClass<*> -> o.asTypeName()
      else -> throw IllegalArgumentException("expected type but was $o")
    }
```

If you think it's a good idea, I could contribute.

micHar:
by only resolving legacy javax.models TypeMirror and Element as a last resort

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -374,6 +374,8 @@ public class CodeBlock private constructor(
 
     private fun argToType(o: Any?) = when (o) {
       is TypeName -> o
+      is Type -> o.asTypeName()
+      is KClass<*> -> o.asTypeName()
       is TypeMirror -> {
         logDeprecationWarning(o)
         o.asTypeName()
@@ -382,8 +384,6 @@ public class CodeBlock private constructor(
         logDeprecationWarning(o)
         o.asType().asTypeName()
       }
-      is Type -> o.asTypeName()
-      is KClass<*> -> o.asTypeName()
       else -> throw IllegalArgumentException("expected type but was $o")
     }
 


ZacSweers:
This isn't really a solution, we should do something better to skip the javax API checks entirely if they're not available

ZacSweers:
It would just fail later in other cases, that's not a fix. I'm not sure what the right approach is here. We could try/catch loading a javax class, but also I'm still not convinced this isn't an environment issue on your side. I haven't had a chance to look at a repro yet

Egorand:
I agree with Zac that this doesn't seem like a good strategy, next time a different instruction fails in a different place in the library and we'll need to move code around again. If this is indeed not a configuration issue, we'll need to discuss more general solutions to the problem of depending on JDK types.

Issue #1304, opened 2022-07-01
KSType.toTypeName incorrectly resolves typealias type argument


micHar:
For `typealias LeAlias = Map<Int, String>`, if I use `KSTypeReference.toTypeName` I get excessively parameterized typealias if the typealias itself is a type param. Complicated 😅, so here's an example:

```
original KSTypeReference -> TypeName
---
LeAlias -> LeAlias (OK)
Flow<LeAlias> -> Flow<LeAlias<Int, String>> (wrong, unexpected type params for LeAlias)
```

kotlinpoet version: 1.12.0

micHar:
What happens is that when `Flow` type is encountered, we fall into [this line](https://github.com/square/kotlinpoet/blob/8f6c9a56bc554fddd32d527bc67b7679fac223a7/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt#L66). If I understand it correctly, `arguments` is resolved to TypeName without regard to wether it's a typealias or not. Relevant info seems to be provided by `KSTypeArgument`, just not used in `KsTypeArgument.toTypeName` like it is in `KsType.toTypeName`.

<img width="660" alt="Zrzut ekranu 2022-07-1 o 14 26 22" src="https://user-images.githubusercontent.com/1609815/176894162-945fa69f-8efc-4991-bfd5-f33a9bc54f1a.png">

micHar:
This still doesn't work in my case, although I can see that it's supposed to be fixed and there are even tests for it.

I used 1.14.2 and 1.15.0-SNAPSHOT.
I matched the kotlin and ksp version with the main branch.
```
kotlin = "1.8.22
ksp = "1.8.22-1.0.11"
```

However I still get this:
```
List<Alias997<Int, String>>
```
from this:
```
typealias Alias997 = Map<Int, String>
List<Alias997>
```

I narrowed it down to the fact, that in `internal fun KSType.toTypeName`
```
    is KSClassDeclaration -> {
      decl.toClassName().withTypeArguments(arguments.map { it.toTypeName(typeParamResolver) })
    }
```
the `arguments` in my case is `Map<Int, String>` and in the kotlinpoet tests it's `Alias997`. `arguments` is part of KSP API, but I cannot understand why in my case it's a resolved typealias and in your case still the unresolved one. @ZacSweers @Egorand do you have any idea what might be the cause?

micHar:
[...]

PR #1320, opened 2022-07-12
add failing test for #1304 - wrong resolution of typealias as type argument

micHar:
Just a failing test, not fix. Following discussion with @oldergod on kotlinlang slack.


micHar:
Seems that @bitPogo [fixed the issue and used this very test](https://github.com/square/kotlinpoet/pull/1321), so I don't think this PR is needed anymore. Closing

PR #1335, opened 2022-08-05
Avoid redundant require for varargs in addImport function

InsanusMokrassar:
I have added this PR as and improvement for addImport functions which will prohibit to use addImport without any added name


JakeWharton:
This is a binary breaking change so we cannot accept it as-is. It also potentially breaks someone using array expansion to fill the varargs, but I'm less concerned about that case.

InsanusMokrassar:
I understand these points. First of all, I wished to show the way to avoid redundant check in runtime and make more clear API for users. I will think about how to make it backward compatible with previous signatures, but currently can't promise anything on this field

PR #1342, opened 2022-08-09
Make shouldEmitPublicModifier false if public isn't implicit

### LINKED ISSUE #1341: Duplicate public modifier when adding Modifier.PUBLIC to a Property's getter or setter
I've pushed a failing test here: https://github.com/square/kotlinpoet/compare/master...sjudd:kotlinpoet:fix_duplicate_modifiers_properties.

The test output is:

```
expected:
    var foo: kotlin.String
      public get() = "_foo"
      public set(`value`) {
      }
    
but was:
    var foo: kotlin.String
      public public get() = "_foo"
      public public set(`value`) {
      }
```

I'll take a look at fixing this as well.

sjudd:
Fixes #1341.

Previously we'd return true even if public was not in the implicit
modifier list, which would cause the calling method to double emit
public. This change makes the behavior of the method match the kdoc,
specifically that it will only return true if Public is in the implicit
modifier list.

Arguably a more robust code change would be to also change the calling
method to explicitly check for and avoid double emitting public. I've
chosen not to do that for now because I'm mildly worried it would hide
future bugs and the existing design seems to work if this method
behaves.


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -628,18 +628,18 @@ internal class CodeWriter constructor(
    * there are no other opposing modifiers (like [KModifier.PROTECTED] etc.) supplied by the
    * consumer.
    */
-  private fun shouldEmitPublicModifier(
+  private fun shouldEmitImplicitPublicModifier(
```
Egorand, line 631: Not sure I agree with the name change, it looks like the KDoc is incorrect - this method does not return true when `public` is in the set of implicit modifiers. My suggestion would be to revert the name change, and to delete the KDoc entirely - it's not needed for a private method, and the code is readable enough to not warrant extra documentation.

sjudd: Just to make sure we're on the same page - the method definitely can return true if public is in the set of implicit modifiers. In fact after my change it will only return true if public is in implicit modifiers (and no other opposing modifiers...).

With that in mind, if you think the original name is clearer and it's better to delete the kdoc, that's fine with me.

It's a bit confusing because implicitModifiers is mostly used to skip explicit modifiers, except in the case of public where it's used to add public even if it's not explicit.

Egorand: Yeah, you're right - just re-read the method. It is indeed confusing.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -628,18 +628,18 @@ internal class CodeWriter constructor(
...
   ): Boolean {
-    if (modifiers.contains(KModifier.PUBLIC)) {
-      return true
-    }
-
     if (!implicitModifiers.contains(KModifier.PUBLIC)) {
       return false
     }
 
+    if (modifiers.contains(KModifier.PUBLIC)) {
```
Egorand, line 639: In fact, is this check even needed? We return false if `public` is an implicit modifier, or if a different visibility modifier is provided; whether or not `public` is in `modifiers` - the method should return true.

sjudd: Yeah in this implementation, it wasn't.

However, I think this approach is kind of confusing, so I went back and just explicitly filtered public in the calling method and then reverted the name/functionality change here. 

Now shouldEmitPublicModifier is the only way the public modifier will be emitted. If checks both explicit and implicit modifiers. If it returns false, public won't be emitted.

I think this is a bit easier to read and understand.

Egorand: Looks great, thanks for taking the time to make it more readable!

Egorand (approved):

Issue #1343, opened 2022-08-10
ParameterSpec disallows private vals in constructors


sjudd:
The assertion in ParameterSpec applies to both function parameters and constructor parameters. While function parameters cannot be visible, constructor parameters can be. E.g. this is perfectly valid Kotlin:

```
class MyClass(private val number: Int)
```

Failing test case: https://github.com/square/kotlinpoet/compare/master...sjudd:kotlinpoet:typespec_disallowed_private_constructor_parameters

Introduced here: 28f977b5f7890555da5966a03734426927525a2d

sjudd:
Ok actually I think this is WAI.

If you define the modifier on the property it works just fine:

```
PropertySpec.builder("number", Int::class).initializer("number").addModifiers(PRIVATE).build())
```

You happen to get the same output if you add modifier to the parameter and the assertion is removed:

```
FunSpec.constructorBuilder().addParameter("number", Int::class, PRIVATE).build(),
```

But that doesn't mean it's correct. There's a bunch of older code lying around that did this, but that's not the library's problem :)

PR #1344, opened 2022-08-10
Add tests for private constructor parameters.

sjudd:
Related to #1343.

These tests actually pass just fine. However I don't see any other tests
specifically for constructor parameters so I thought they might be
helpful to add anyway.


Egorand:
Thank you!

PR #1355, opened 2022-08-25
Auto-generate import aliases for types and members

### LINKED ISSUE #938: Auto-suggested import aliases
This feature proposal stems from #924, where in case if a type cannot be imported due to a simple name clash, generating a reference to that type's constructor would produce broken code:

```kotlin
val squareTaco = ClassName("com.squareup.tacos", "Taco")
val twitterTaco = ClassName("com.twitter.tacos", "Taco")
val file = FileSpec.builder("com.squareup.example", "Tacos")
        .addStatement("val squareTacosCreator = %L", squareTaco.constructorReference())
        .addStatement("val twitterTacosCreator = %L", twitterTaco.constructorReference())
        .build())
    .build()
println(file)

// Output:
//
// ...
// val squareTacosCreator = ::Taco
// val twitterTacosCreator = ::com.twitter.tacos.Taco // Fails to compile!
// ...
```

A similar problem occurs whenever a `MemberName` that cannot be imported due to a name clash is used as an extension function:

```kotlin
val kotlinIsNullOrEmpty = MemberName("kotlin.text", "isNullOrEmpty")
val cashIsNullOrEmpty = MemberName("com.squareup.cash.util", "isNullOrEmpty")
val file = FileSpec.builder("com.squareup.example", "Tacos")
        .addStatement("println(%S.%M())", "Taco", kotlinIsNullOrEmpty)
        .addStatement("println(%S.%M())", "Taco", cashIsNullOrEmpty)
        .build())
    .build()
println(file)

// Output:
//
// ...
// println("Taco".isNullOrEmpty())
// println("Taco".com.squareup.cash.util.isNullOrEmpty())  // Fails to compile!
// ...
```

The workaround for this issue is to use [`FileSpec.Builder.addAliasedImport()`](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/-file-spec/-builder/add-aliased-import/). For instance, in the first example, `addAliasedImport(twitterTaco, "TwitterTaco")` will turn the output into:

```kotlin
...
val squareTacosCreator = ::Taco
val twitterTacosCreator = ::TwitterTaco
...
```

However, managing import aliases by hand is not ideal:

- First of all, this is not an intuitive solution to the problem. If valid import aliases are not provided KotlinPoet will silently generate broken code and we'll get bug reports.
- If the generated code changes, proposed import aliases may start clashing with newly introduced type names, which will again fail silently.
- If the simple name clash disappears due to changes in the generated code, proposed import aliases will become unnecessary, and there will be no clear indication that they can be removed.

### Proposal

- Update `CodeWriter` to keep track of types and members that could not be imported due to a name clash.
- For those types and members, generate unique aliases and use them to create `Import` instances.
- At the end of the import collection pass, `FileSpec` should retrieve the `Import`s from the `CodeWriter` and pass them into the new instance of `CodeWriter`, along with the `memberImports`.
- During the second pass, `CodeWriter` should treat generated import aliases in the same way as if they were added manually via `addAliasedImport()`.

#### Alias generation

Since import aliases are local to the generated file and will not be seen by the consumers of the API defined in this file, we should prefer correctness over readability, still making an effort to generate meaningful aliases. For types, we can prefix simple names with capitalized letters of each of the package name segments, e.g.:

```
com.squareup.cash.Taco -> CSCTaco
```

For members, we can use the same letters in lowercase and capitalize the first letter of the member name, e.g.:

```
com.squareup.cash.util.isNullOrEmpty -> cscuIsNullOrEmpty
```

The algorithm should be implemented by the `AliasAllocator`, which would function similarly to `NameAllocator`, holding on to the global pool of allocated aliases to avoid clashes. `AliasAllocator` can be made public and used as a utility when generating a valid alias is required outside of KotlinPoet. Having `AliasAllocator` would also allow us to introduce overloads to methods such as [`FileSpec.Builder.addAliasedImport()`](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/-file-spec/-builder/add-aliased-import/) or [`TypeAlias.builder()`](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/-type-alias-spec/builder/) which do not require passing the actual alias.

Looking for feedback on this feature and volunteers to build it.

Egorand:
Closes #938

### DIFF
--- a/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
+++ b/kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
@@ -360,12 +360,13 @@ class FileSpecTest {
       """
         |package com.squareup.tacos
         |
-        |import java.util.Date
+        |import java.sql.Date as SqlDate
+        |import java.util.Date as UtilDate
         |
         |public class Taco {
-        |  public val madeFreshDate: Date
+        |  public val madeFreshDate: UtilDate
         |
-        |  public val madeFreshDatabaseDate: java.sql.Date
+        |  public val madeFreshDatabaseDate: SqlDate
         |}
         |
       """.trimMargin(),
@@ -385,12 +386,13 @@ class FileSpecTest {
       """
         |package com.squareup.tacos
         |
-        |import com.squareup.`is`.`fun`.`in`.Date
+        |import com.squareup.`do`.`val`.`var`.Date as VarDate
+        |import com.squareup.`is`.`fun`.`in`.Date as InDate
         |
         |public class Taco {
-        |  public val madeFreshDate1: Date
+        |  public val madeFreshDate1: InDate
         |
-        |  public val madeFreshDate2: com.squareup.`do`.`val`.`var`.Date
+        |  public val madeFreshDate2: VarDate
         |}
         |
       """.trimMargin(),
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeWriter.kt
@@ -16,6 +16,7 @@
 package com.squareup.kotlinpoet
 
 import java.io.Closeable
+import kotlin.math.min
 
 /** Sentinel value that indicates that no user-provided package has been set.  */
 private val NO_PACKAGE = String()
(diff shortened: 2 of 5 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/FileSpecTest.kt
```diff
@@ -360,12 +360,13 @@ class FileSpecTest {
...
         |
-        |import java.util.Date
+        |import java.sql.Date as SqlDate
+        |import java.util.Date as UtilDate
         |
         |public class Taco {
-        |  public val madeFreshDate: Date
+        |  public val madeFreshDate: UtilDate
         |
-        |  public val madeFreshDatabaseDate: java.sql.Date
```
adrw, line 368: Nice!

adrw (approved):

PR #1384, opened 2022-10-11
Feat: Add number formatting

### LINKED ISSUE #1381: Consider inserting underscores into large decimal literals
```kotlin
CodeBlock.of("%L", 1000000)
```
should probably produce
```kotlin
1_000_000
```

Sironheart:
Fixes #1381


kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -361,9 +363,23 @@ public class CodeBlock private constructor(
...
-    private fun argToString(o: Any?) = o?.toString()
+    private fun argToString(o: Any?) = if (o is Number) formatNumericValue(o) else o?.toString()
+
+    private fun formatNumericValue(o: Number): Any? {
+      val format = DecimalFormatSymbols().apply {
+        decimalSeparator = '.'
+        groupingSeparator = '_'
+      }
+
+      val formattedValue = DecimalFormat("###,###.####", format).format(o)
```
Egorand, line 376: Not really familiar with `DecimalFormat`'s pattern rules, I see in the tests that this pattern handles numbers with more than 6 digits, will it also properly handle numbers with more than 4 decimal digits? Wanna add a test case for this?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -18,6 +18,8 @@
 package com.squareup.kotlinpoet
 
 import java.lang.reflect.Type
+import java.text.DecimalFormat
```
Egorand, line 21: Wondering if there are multiplatform-friendly Kotlin alternatives, while KotlinPoet currently won't work without a JVM, in future we'd potentially want it to be JDK-free.

Egorand (approved):
Looks great!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Sironheart: I think there currently is no real alternative, except string-foo... I'd guess that the `DecimalFormat` option is the best for the moment, since string-foo could cause more issues until the switch to multiplatform

Sironheart:
Last but not least: I needed to remove the Long/ Float notation, since it caused to some errors in the other unit tests (for example some tests contained `0.123f` and similar)
I hope this is still the best solution for the issue

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
JakeWharton: I don't think the string-foo would be a terrible approach. We can always switch to it when needed.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -361,9 +363,25 @@ public class CodeBlock private constructor(
       else -> throw IllegalArgumentException("expected name but was $o")
     }
 
-    private fun argToLiteral(o: Any?) = o
+    private fun argToLiteral(o: Any?) = if (o is Number) formatNumericValue(o) else o
 
-    private fun argToString(o: Any?) = o?.toString()
+    private fun argToString(o: Any?) = if (o is Number) formatNumericValue(o) else o?.toString()
```
JakeWharton, line 368: Please revert. We should not alter number formatting flowing into string literals.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -604,4 +604,27 @@ class CodeBlockTest {
...
+  @Test fun useUnderscoresOnLargeDecimalLiterals() {
+    assertThat(CodeBlock.of("%L", 10000).toString()).isEqualTo("10_000")
+    assertThat(CodeBlock.of("%L", 100000L).toString()).isEqualTo("100_000")
+    assertThat(CodeBlock.of("%L", Int.MIN_VALUE).toString()).isEqualTo("-2_147_483_648")
+    assertThat(CodeBlock.of("%L", Int.MAX_VALUE).toString()).isEqualTo("2_147_483_647")
+    assertThat(CodeBlock.of("%L", Long.MIN_VALUE).toString()).isEqualTo("-9_223_372_036_854_775_808")
+    assertThat(CodeBlock.of("%L", 10000.123).toString()).isEqualTo("10_000.123")
+    assertThat(CodeBlock.of("%L", 3.0).toString()).isEqualTo("3.0")
+    assertThat(CodeBlock.of("%L", 10000.123f).toString()).isEqualTo("10_000.123")
+    assertThat(CodeBlock.of("%L", 10000.123456789012).toString()).isEqualTo("10_000.123456789011")
```
JakeWharton, line 618: Need tests for shorts.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Sironheart: Thats what I thought, yet I think, for the moment, it's just a safer bet to use the java given features. Do you think like this too?

kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
Sironheart: I reverted this string formatting

kotlinpoet/src/test/java/com/squareup/kotlinpoet/CodeBlockTest.kt
Sironheart: I added one test for shorts (i.e. `1281.toShort()`) Do you think there are more cases that should be properly tested, that I haven't thought of?

Sironheart:
@Egorand I did sign the CLA.

Egorand:
Thank you!

PR #1398, opened 2022-10-26
Feature/omit non private modifiers on private types

### LINKED ISSUE #1301: Omit non-private modifiers on private types
Thankfully it's not a warning, but the IDE still soft warns about it:

Screen Shot 2022-06-29 at 2 18 28 PM

polarene:
When generating the string representation of functions inside a private type, all non-private modifiers (PUBLIC, INTERNAL) are omitted.

Fixes #1301

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -30,6 +30,7 @@ import com.squareup.kotlinpoet.KModifier.PUBLIC
 import com.squareup.kotlinpoet.KModifier.SEALED
 import com.squareup.kotlinpoet.KModifier.VALUE
 import java.lang.reflect.Type
+import java.util.EnumSet
 import javax.lang.model.element.Element
 import kotlin.reflect.KClass
 
@@ -299,10 +300,10 @@ public class TypeSpec private constructor(
       }
 
       // Functions.
-      for (funSpec in funSpecs) {
+      for (funSpec in funSpecsOmittingVisibility()) {
         if (funSpec.isConstructor) continue
         if (!firstMember) codeWriter.emit("\n")
-        funSpec.emit(codeWriter, name, kind.implicitFunctionModifiers(modifiers + implicitModifiers), true)
+        funSpec.emit(codeWriter, name, implicitFunctionModifiers(implicitModifiers), true)
         firstMember = false
       }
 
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -299,10 +300,10 @@ public class TypeSpec private constructor(
       }
 
       // Functions.
-      for (funSpec in funSpecs) {
+      for (funSpec in funSpecsOmittingVisibility()) {
```
Egorand, line 303: Not sure it makes sense to filter out the modifiers both here and when passing the implicit modifiers to `funSpec.emit()` - maybe we should simply add unnecessary visibility modifiers to `implicitModifiers` if the type is private? Any reason why that wouldn't work?

JakeWharton: Hopefully that also adds support for properties which I suspect exhibit the same behavior.

polarene: Honestly, I don't know the library this deep... I tried to reverse engineer how the mechanism works and found out that filtering here before emit() caused the least amount of code changes. What do you suggest would be a better approach? Also, I limited the scope to functions only, following the issue description, but if you want we can extend it to properties as well, need to double-check.

polarene: I think when I did the initial implementation, I tried to add visibility modifiers to `implicitModifiers` but that didn't work for PUBLIC, since it undergoes a contrived logic I couldn't fully grasp, so I turned to this approach.

polarene: @Egorand so what should we do with this implementation, do you have any advice?

Egorand: I think the least amount of changes should not be the priority here, and your PR IMO adds complexity rather than reducing it. I believe `implicitModifiers` is the right tool for the job, and it'd be great to remove the obstacles preventing it from working correctly - e.g. the custom handling of `PUBLIC` should probably be removed in favour of achieving the same result with `implicitModifiers`. This will likely require deeper knowledge of the library's inner workings and understand if you don't have the time to invest, but I don't think current approach is the best one.

polarene: Ok, thanks for your feedback. You're right, this solution increases complexity and it was a quick workaround for  `implicitModifiers` not working as expected. I'll try to dedicate some time to this issue and rework the solution following your guidelines.

Egorand: Thank you!

PR #1439, opened 2023-01-09
Remove future from mkdocs-requirements.txt

Egorand:
Attempt to address https://github.com/square/kotlinpoet/security/dependabot/1. Not sure if there's a better way to check if this package is actually used, so trying to remove it to see if the build still passes. The package has not been updated since June 2020, so not sure there will be a release that fixes the vulnerability: https://github.com/PythonCharmers/python-future.

### DIFF
--- a/.github/workflows/mkdocs-requirements.txt
+++ b/.github/workflows/mkdocs-requirements.txt
@@ -1,5 +1,4 @@
 click==8.1.3
-future==0.18.2
 Jinja2==3.1.2
 livereload==2.6.3
 lunr==0.6.2


JakeWharton:
The python is not deployed anywhere so you can ignore all security problems. It's only used locally as a tool.

JakeWharton (approved):

Egorand:
Ah, so technically we might run into issues running `mkdocs` commands locally if this package is not installed? Maybe I should dismiss the alert instead and leave this dependency in?

Egorand:
Superseded by #1443

PR #1440, opened 2023-01-09
Don't test on JDK 10

Egorand:
Not sure what caused CI to start failing tests on JDK 10 and not sure what exactly is the issue, but JDK 10 is probably old enough to stop caring about it. Thoughts?


JakeWharton (approved):
This is okay. I'm fine with it as is.

A better fix would likely be to use https://github.com/gradle/foojay-toolchains so that Gradle provisions Azul Zulu which has all versions for all platforms and won't break. But we don't need to worry about it right now.

Issue #941, opened 2020-07-15
Consecutive blank lines with top level property


btrautmann:
I'm noticing that when creating a top-level property in a file, two blank lines are output below the property, causing detekt to fail with a `NoConsecutiveBlankLines` violation.

Sample `FileSpec`:
```
import com.squareup.kotlinpoet.CodeBlock
import com.squareup.kotlinpoet.FileSpec
import com.squareup.kotlinpoet.PropertySpec

val spec = FileSpec.builder("com.example.foo", "Test")
  .addProperty(
    PropertySpec.builder("MyProperty", String::class.java)
      .initializer(
        CodeBlock.builder()
            .beginControlFlow("if (1 + 1 == 2)")
            .addStatement("Expected")
            .endControlFlow()
            .beginControlFlow("else")
            .addStatement("Unexpected")
            .endControlFlow()
            .build()
      ).build()
  ).build()
```

Results in:
```
val equals = """
    |package com.example.foo
    |
    |import java.lang.String
    |
    |val MyProperty: String = if (1 + 1 == 2) {
    |  Expected
    |}
    |else {
    |  Unexpected
    |}
    |
    | <-- Extra line. Removing this causes `equals` to be `false`
""".trimMargin() == spec.toString()
println(equals) <-- Prints `true`
```

I don't think we should ever have consecutive blank lines.

Egorand:
Thanks for reporting, please feel free to send us a PR.

That said, we usually discourage people from running code quality tools over KotlinPoet's output. Generated code is semantically different from source code, so you shouldn't expect it to match the levels of quality that you would expect from code written by a programmer. It should either live in `build/generated` or be added to the code quality tool's excludes.

btrautmann:
@Egorand Thanks, and yeah, in most cases I do agree with you. In our particular case, we are using KotlinPoet to generate some actual source to help alleviate the pain of some boilerplate code needing to be written for each new "mini-feature-module". Totally understand that it's an unorthodox usage of generated code. If I have some time to look into the reason for the extra line, I'll give a contribution a shot!

JakeWharton:
In that case you might want to run ktfmt or ktlint's formatter on the output as part of the generation process.

BTW you can replace `endControlFlow().beginControlFlow()` calls with `nextControlFlow()`

btrautmann:
[...]

Issue #1300, opened 2022-06-29
Expression body indent wrong


JakeWharton:
Today:

```kotlin
  public override val descriptor: SerialDescriptor = buildString {
      }
```
```kotlin
  public override val descriptor: SerialDescriptor =
      buildClassSerialDescriptor("RowVerticalAlignment") {
      }
```

Should be:

```kotlin
  public override val descriptor: SerialDescriptor = buildString {
  }
```
```kotlin
  public override val descriptor: SerialDescriptor =
    buildClassSerialDescriptor("RowVerticalAlignment") {
    }
```

We need to remove the function indent when it collapses so that wraps are +2 and begin/end control flows can have their curly's flush with the function's visibility modifier indent level.

JakeWharton:
I made the second example a non-empty block and things got weirder:
```kotlin
  public override val descriptor: SerialDescriptor =
      buildClassSerialDescriptor("RowVerticalAlignment") {
    element<Int>("direction")
  }
```

Here's the code:
```kotlin
addProperty(PropertySpec.builder("descriptor", SerialDescriptor)
  .addModifiers(OVERRIDE)
  .initializer(CodeBlock.builder()
    .beginControlFlow("%M(%S)", buildClassSerialDescriptor, layoutModifierType.simpleName)
    .apply {
      for (property in layoutModifier.traits) {
        addStatement("%M<%T>(%S)", element, property.type.asTypeName(), property.name)
      }
    }
    .endControlFlow()
    .build())
  .build())
```

JakeWharton:
```kotlin
  public override fun equals(other: Any?): Boolean = other is AccessibilityDescription
  && other.`value` == `value`
```
Beginning to think expression body support was a mistake because it seems to be wrong more often than not.

Egorand:
I dug a bit into how we emit `PropertySpec.initializer`, and the logic is that we'll wrap the entire `CodeBlock` into a statement (`«%L»`) unless it already includes statements. Emitting the `CodeBlock` as a single statement seems to produce expected results. Maybe we should just do a better job processing the `CodeBlock` before emitting it?

PR #1460, opened 2023-01-28
Fix extra newline when a property has an initializer with statement (#941)

drawers:
Fix extra newline when a property has an initializer with a statement in it (#941)


drawers:
I have signed the CLA via the Google Form

drawers:
Hmm, I just saw https://github.com/square/kotlinpoet/issues/1300#issuecomment-1173844669 and I think pre-processing the initializer code block is a better solution. So I'm no longer sure if this current PR is a good idea.

Egorand:
Thanks for your PR! I have not explored the idea of processing the `CodeBlock` any further, if you're interested in doing it that'd be appreciated. But also, if your code works for all test cases outlined in #941 and #1300 I'm happy to merge it (provided that unit tests for those test cases are added to the PR).

PR #1477, opened 2023-02-14
Fix TypeName equals #1470

### LINKED ISSUE #1470: `TypeName.equals` is broken
I don't have time to fix it right now, but
```kotlin
  @Test fun equalsDifferentiatesPackagesFromSimpleNames() {
    assertThat(ClassName("com.example.Foo", "Bar"))
      .isNotEqualTo(ClassName("com.example", "Foo", "Bar"))
  }
```
this should not pass and it does today.

The `TypeName` hierarchy is sealed. There's no reason to to a `toString()` comparison anymore and we can do proper comparison of things.

drawers:
Not sure if you are interested in a patch for #1470 where it is suggested to fix equality for the `TypeName` hierarchy.


interop/javapoet/src/test/kotlin/com/squareup/kotlinpoet/javapoet/PoetInteropTest.kt
```diff
@@ -139,23 +139,23 @@ class PoetInteropTest {
       JWildcardTypeName.supertypeOf(String::class.java),
     )
     assertThat(inKType.toJTypeName()).isEqualTo(superJType)
-    assertThat(superJType.toKTypeName()).isEqualTo(inKType)
+    assertThat(superJType.toKTypeName().toString()).isEqualTo(inKType.toString())
```
drawers, line 142: This test was only passing because `TypeName` equality was previously equality of `toString()` representation. 

In actuality, the `inKType` created through `typeNameOf<GenericType<in String>>()` gets an `enclosedType` while the version obtained through `JParameterizedTypeName` does not.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/LambdaTypeNameTest.kt
```diff
@@ -192,4 +192,40 @@ class LambdaTypeNameTest {
     assertThat(typeName.toString())
       .isEqualTo("((kotlin.Int) -> kotlin.Int) -> kotlin.Unit")
   }
+
+  @Test fun equalsAndHashCode() {
+    val lambdaTypeName1 = LambdaTypeName.get(
+      parameters = arrayOf(Int::class.asTypeName()),
```
Egorand, line 198: Can use `INT` and `ANY`, declared inside `TypeName`.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -340,7 +340,8 @@ public class TypeSpec private constructor(
       val property = propertySpecs[propertyIndex]
       if (property.getter != null || property.setter != null) continue
       val parameter = primaryConstructor.parameter(property.name) ?: continue
-      if (parameter.type != property.type) continue
+      // compare TypeName using `toString()` to account for type aliases
```
Egorand, line 343: Is there a failing test case for this condition? Feels a bit off, as a `TypeAliasSpec` would still point to the real `TypeName` through its `type` property, which I guess is what the `ParameterSpec` here is referencing?

Egorand (commented):
Added comments: would like to clarify a couple of details before I can approve. Looks great though!

interop/javapoet/src/test/kotlin/com/squareup/kotlinpoet/javapoet/PoetInteropTest.kt
drawers: Maybe a way to highlight this to consumers would be:

`equals()` and `hashCode()` of `TypeName` are now more strict. The previous behavior can be retained by comparing `toString()` representations.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
```diff
@@ -95,12 +95,23 @@ public sealed class TypeName constructor(
 
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
-    if (other == null) return false
-    if (javaClass != other.javaClass) return false
-    return toString() == other.toString()
+    if (javaClass != other?.javaClass) return false
```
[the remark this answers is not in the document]
drawers: Thanks, so I looked it up and it seems the check uses the runtime class. 

https://github.com/JetBrains/kotlin/blob/30788566012c571aa1d3590912468d1ebe59983d/libraries/stdlib/jvm/runtime/kotlin/jvm/JvmClassMapping.kt#L85

So when comparing two `ClassName`, the equals in `ClassName` will call `super.equals()` which will invoke `TypeName.equals()`. Because the runtime class of both is `ClassName`, the `super.equals()` will return true if nullable, annotations, and tags are the same.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -175,12 +180,32 @@ public class ClassName internal constructor(
    * com.example.Robot.Motor
    * com.example.RoboticVacuum
    * ```
+   *
+   * Comparison is consistent with equals()
```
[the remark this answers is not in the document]
drawers: Thank you, this is a really good catch. 

Now that I have another look (70f18ae23) I am not sure about the whole enterprise of trying to implement a `compareTo` that is consistent with equals for `ClassName`. This is because there are lots of edge cases and subtleties, especially with tags where the entry values can be of type `Any`. 

Maybe it is better to just accept that the `compareTo` will never be consistent with `equals()` and just warn consumers of this fact? When I looked it up (https://blog.joda.org/2012/11/pitfalls-of-consistent-with-equals.html) seems there are a few famous classes that do this (like `BigDecimal`). For the purposes of this PR, it would mean deleting all of the tests related to `compareTo` and reverting back to the old implementation that just used string comparison.

What do you think?

kotlinpoet/src/test/java/com/squareup/kotlinpoet/LambdaTypeNameTest.kt
drawers: Thanks, done as 53c75f56

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
drawers: Thanks, the failing test is:
https://github.com/square/kotlinpoet/blob/master/interop/kotlinx-metadata/src/test/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecsTest.kt#L200

I spent more time looking into the underlying cause of the comparison failure, and it is actually because the tagMap are not equal because values of `TypeAliasTag` are not equal. I tried to fix this in 0619ef2b8

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt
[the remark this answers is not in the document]
Egorand: Ah, that makes sense, thank you for clarifying!

interop/javapoet/src/test/kotlin/com/squareup/kotlinpoet/javapoet/PoetInteropTest.kt
Egorand: SGTM!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ClassName.kt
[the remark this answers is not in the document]
Egorand: Yeah, I think striving to preserve consistency with equals() is a lot of complexity for no apparent gain, my vote is for sticking with the existing behaviour and documenting the fact that compareTo() is not consistent with equals(). @JakeWharton @ZacSweers would love to hear your thoughts on this.

In any case though, I think the right course of action would be to revert changes related to compareTo() in this PR and open a separate one if needed, as these seem like separate bits of work.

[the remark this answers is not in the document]
JakeWharton: I would not include tags in anything. They're side-channel data. Does that make `compareTo` easier? If not, falling back to `toString()` for it seems fine, as that's like the lexicographical sorting of the `import` list in the IDE anyway.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
Egorand: Looks good, thanks!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/ClassName.kt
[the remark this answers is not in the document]
drawers: Thank you everyone, I have removed tag comparison from `equals()` and `hashCode()` and noted that they are side-channel data in a comment. 

I also removed the adjustments to `compareTo` for now. I think that should be reviewed separately as a different piece of work.

Egorand (approved):

Egorand:
@JakeWharton the PR looks good to me and I'm planning to merge it late next week, let us know if you're planning to take another look.

Egorand:
Merging in preparation for a new release. Thanks @drawers!

drawers:
> Merging in preparation for a new release. Thanks @drawers!

Thank you @Egorand :pray: - I followed up with fixing `compareTo` so you don't end up with a broken release in this PR:

https://github.com/square/kotlinpoet/pull/1506

PR #1492, opened 2023-03-13
Added `useAsParameter` flag to AnnotationSpec

lev-zakaryan:
Greetings! 

I was looking for a way to solve an issue with moshi (https://github.com/square/moshi/issues/1581), but as far as I can tell it cannot be fixed in moshi only.

Moshi uses kotlinpoet to emit annotations. Kotlinpoet has a mechanism to emit annotation as a parameter (dropping initial '@' in the emission); this mechanism is started by the parent annotation, but there is no way to emit parent annotation itself as a parameter. And that is something that has to be done from moshi (afaiu) in order to properly generate a class adapter with nested annotations.

If this PR is approved, it would be possible to call this code from [DelegateKey](https://github.com/square/moshi/blob/1af40a433f2125421d9ced0f45e358cda1283c3f/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/DelegateKey.kt#L79) class in moshi:

```
private fun AnnotationSpec.asInstantiationExpression(): CodeBlock {
  return CodeBlock.of(
    "%L",
    this.toBuilder().useAsParameter().build()
  )
}
```

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -38,10 +38,11 @@ public class AnnotationSpec private constructor(
     get() = typeName as? ClassName ?: error("ClassName is not available. Call typeName instead.")
   public val typeName: TypeName = builder.typeName
   public val members: List<CodeBlock> = builder.members.toImmutableList()
+  public val useAsParameter: Boolean = builder.useAsParameter
   public val useSiteTarget: UseSiteTarget? = builder.useSiteTarget
 
   internal fun emit(codeWriter: CodeWriter, inline: Boolean, asParameter: Boolean = false) {
-    if (!asParameter) {
+    if (!asParameter && !useAsParameter) {
       codeWriter.emit("@")
     }
     if (useSiteTarget != null) {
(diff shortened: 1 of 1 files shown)


Egorand:
In the Moshi issue, you mention that "Moshi 1.12.0 + kotlin 1.5.0 + kapt generates proper adapter", does it mean that the functionality regressed at some point? Is it clear what caused the regression? Is there a way to fix this issue without introducing new API?

lev-zakaryan:
> In the Moshi issue, you mention that "Moshi 1.12.0 + kotlin 1.5.0 + kapt generates proper adapter", does it mean that the functionality regressed at some point? Is it clear what caused the regression? Is there a way to fix this issue without introducing new API?

Hello @Egorand, thank you for looking into this.

For some reason, I was under the wrong impression that moshi 12.0 did not use kotlinpoet, to be honest. After reading your comment I have searched git history and found [this ](https://github.com/square/moshi/commit/73553286fb482040990c9efadd829faf849b676e) commit that changed the way [DelegateKey](https://github.com/square/moshi/commit/73553286fb482040990c9efadd829faf849b676e#diff-38d9cd4fce3da293198ab776387c064107fd9da2b3ded024807af16cea91a12a) works.

It does not look like it was changed by mistake. 

I have also tried putting back this old version of `initializerString` into the latest code:
```
val (initializerString, args) = when {
      jsonQualifiers.isEmpty() -> ", %M()" to arrayOf(MemberName("kotlin.collections", "emptySet"))
      else -> {
        ", %T.getFieldJsonQualifierAnnotations(javaClass, " +
          "%S)" to arrayOf(Types::class.asTypeName(), adapterName)
      }
    }
```
It does work, but several tests start to fail.

Egorand:
Could be a regression in Moshi then? If you could fix up the tests and submit a PR against Moshi that'd be much appreciated, I don't see any activity on that ticket so your contribution would help move it forward.

Let me close your PR for now as I believe the problem should be addressed on the Moshi side, but please let me know if you think differently and I can reopen it. Thank you!

PR #1503, opened 2023-03-28
add FunSpec.Builder.endControlFlowWithComma

mscheong01:
When generating code that includes constructor invocation, we come across the need to end the control flow with `},`, not just `}`
ex)
```kotlin
val some = SomeClass(
    name = if ( // some condition ) {
        "a"
    } else {
        "b"
    },
    age = 10,
)
```
we could do 
```kotlin
funSpecBuilder
    ~
    .endControlFlow()
    .addStatement(",")
    ~
    .build()
```
but this would end up as:
```kotlin
}
,
````
which is't ideal.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/CodeBlock.kt
@@ -442,6 +442,11 @@ public class CodeBlock private constructor(
       add("}\n")
     }
 
+    public fun endControlFlowWithComma(): Builder = apply {
+      unindent()
+      add("},\n")
+    }
+
     public fun addStatement(format: String, vararg args: Any?): Builder = apply {
       add("«")
       add(format, *args)
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -550,6 +550,10 @@ public class FunSpec private constructor(
       body.endControlFlow()
     }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by Egorand — COMMENTED** [review]
I don't think we should introduce new API to work around this limitation. Would we need to introduce more similar methods if, e.g., we'll want to allow semicolons after the control flow? 

I think the right way to address this is to remove the implicit `\n` from `endControlFlow()`, see what breaks and how to deal with it.

**Comment by JakeWharton** [comment]
This is #1330 and I'm not sure adding this API is a great way to solve it. In particular, the whole control flow API was built around statements in JavaPoet and Kotlin is a more expression-oriented language. I think I'd prefer if we solved the problem generally.

**Comment by mscheong01** [comment]
I understand 🙂 What other implementations would there be?
> I think the right way to address this is to remove the implicit \n from endControlFlow(), see what breaks and how to deal with it.

I'm ok with this, but my concern is that it's a breaking change and some code would break as a result
ex)
```kotlin 
if (...) {
    return a
}
return b
```
this code generated with control endControlFlow() would become
```kotlin
if (...) {
    return a
}return b // break
```
maybe I should close this pr and create an issue where we could further discuss about a general solution?

JakeWharton:
Yeah I'm opposed to that simple change too for that very reason.

I think maybe we could do a %E which was like %L but which trimmed trailing newlines and that would allow you to use the otherwise statement-based control flow helpers (among others) in a CodeBlock and then embed it as an expression using %E.

mscheong01:
I'll send a issue for this general issue with having to add prefixes to statements or control flows before the newline 😄 
By the way, thank's for the speedy support 🙇‍♂️

PR #1506, opened 2023-04-03
Make equals consistent with compareTo for ClassName

drawers:
Follow-up of https://github.com/square/kotlinpoet/pull/1477#discussion_r1108320463

To make compareTo consistent with equals we have to ensure:

> e1.compareTo(e2) == 0 has the same boolean value as e1.equals(e2) for every e1 and e2 of class C. 

https://blog.joda.org/2012/11/pitfalls-of-consistent-with-equals.html

In other words, we cannot get a `compareTo == 0` result for two unequal ClassName.

For ClassName, we are comparing by name parts, by nullability, then by annotations. 

Comparing by name parts and nullability are safe because equality of ClassName is defined by equality of name (in ClassName.kt) and nullability (through the super TypeName).

In general, comparing by using `joinToString()` on some list property included in equals checks is not safe when you have two elements in the list x and y where `listOf(x, y).joinToString()` is the same as `listOf(x + y).joinToString()`. 

I made a Kotlin Playground to show how this could break in general:
https://pl.kotl.in/oCxOv0Sjh

For `AnnotationSpec`, it is safe because we can never have a single AnnotationSpec in a list that gives the same `joinToString()` as two separate AnnotationSpec in a list and because equality of AnnotationSpec is equality of string representation.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
@@ -40,6 +40,13 @@ public class AnnotationSpec private constructor(
   public val members: List<CodeBlock> = builder.members.toImmutableList()
   public val useSiteTarget: UseSiteTarget? = builder.useSiteTarget
 
+  /** Lazily-initialized toString of this type name.  */
+  private val cachedString by lazy {
+    buildCodeString {
+      emit(this, inline = true, asParameter = false)
+    }
+  }
+
   internal fun emit(codeWriter: CodeWriter, inline: Boolean, asParameter: Boolean = false) {
     if (!asParameter) {
       codeWriter.emit("@")
@@ -96,9 +103,7 @@ public class AnnotationSpec private constructor(
 
   override fun hashCode(): Int = toString().hashCode()
 
-  override fun toString(): String = buildCodeString {
-    emit(this, inline = true, asParameter = false)
-  }
+  override fun toString(): String = cachedString
 
   public enum class UseSiteTarget(internal val keyword: String) {
     FILE("file"),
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
```diff
@@ -40,6 +40,13 @@ public class AnnotationSpec private constructor(
   public val members: List<CodeBlock> = builder.members.toImmutableList()
   public val useSiteTarget: UseSiteTarget? = builder.useSiteTarget
 
+  /** Lazily-initialized toString of this type name.  */
+  private val cachedString by lazy {
```
drawers, line 44: I copied the pattern in 

https://github.com/square/kotlinpoet/blob/master/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt#L73

Egorand (approved):
Nice, the tests are very thorough - thanks so much!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
ZacSweers: nit but I think it's fine to make this an unsafe lazy -> `lazy(NONE)`

ZacSweers (approved):
One small nit!

kotlinpoet/src/main/java/com/squareup/kotlinpoet/AnnotationSpec.kt
drawers: Thanks! done [a39dbb7](https://github.com/square/kotlinpoet/pull/1506/commits/a39dbb7bd0107a27432b4aaa8c36bd175f0b0def)

ZacSweers (approved):

PR #1508, opened 2023-04-05
Rename submodules to avoid GAV coordinate conflict with JavaPoet

### LINKED ISSUE #1507: Can't assemble `:interop:javapoet` when `VERSION_NAME` in gradle.properties is set to "1.13.0" or higher
**To Reproduce**

1. Clone the repo.
2. Run `./gradlew :interop:javapoet:assemble --rerun-tasks` and ensure the build passes.
3. Change the value of the `VERSION_NAME` property inside `gradle.properties` to `1.13.0` and re-execute the command above - the build will fail reporting compilation errors.
4. Try the same changing the version to anything higher than `1.13.0`.

Egorand:
Ran into what looks like https://github.com/gradle/gradle/issues/847 while trying to release 1.13.0. Since the version of JavaPoet we depend on is also 1.13.0, the GAV coordinates of JavaPoet and the :interop:javapoet modules are exactly the same -
com.squareup:javapoet:1.13.0, which Gradle can't disambiguate.

Renaming the modules should help avoid this problem. A cleaner solution would've been to change the group name to a unique value, e.g. com.squareup.kotlinpoet, but this will have to wait until 2.0.

Fixes #1507


Egorand:
Worth noting that this doesn't affect the Maven coordinates of the modules when they're published, as those come from a different source.

oldergod (approved):
👍

JakeWharton:
I would prefer we rename/move the files to match if only as a reduction in build customization but also because it's very confusing when the Gradle path does not line up with the folder path.

ZacSweers:
Can you set the group? The maven publish plugin should set them for you and avoid the need for this

Egorand:
> I would prefer we rename/move the files to match if only as a reduction in build customization but also because it's very confusing when the Gradle path does not line up with the folder path.

Sure, I can do that.

> Can you set the group? The maven publish plugin should set them for you and avoid the need for this

Do you mean setting `allprojects { group = "com.squareup.kotlinpoet" }` but passing the current value to the publish plugin, to have the KP-specific group for Gradle and avoid publishing artifacts with a new group name?

Egorand:
There does not seem to be a way to set a different group on the [`MavenPom`](https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPom.html), it's tied to the `group` property of the `Project`, so it looks like our only option is renaming modules.

Egorand:
I'm gonna merge as is for now to unblock my release - @ZacSweers LMK if I misinterpreted your suggestion. Happy to revert these changes if there's a cleaner way to set the group, otherwise I'll follow up with moving files around to correspond to the new module structure.

PR #1510, opened 2023-04-06
Automatically publish releases on CI

Egorand:


.github/workflows/release.yml
```diff
@@ -1,33 +1,21 @@
-name: Build
+name: Release
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '[0-9]+.[0-9]+*'
```
JakeWharton, line 8: Our tags have three numbers, not two. Also the dot needs escaped since it matches any character, not just a dot. However, I would recommend just removing this completely and instead just use a full wildcard `**` like we do in other projects: https://github.com/cashapp/molecule/blob/trunk/.github/workflows/release.yaml#L3-L6

gradle.properties
```diff
@@ -1,7 +1,6 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 
 GROUP=com.squareup
-VERSION_NAME=1.14.0-SNAPSHOT
```
JakeWharton, line 4: Doesn't this break the ability to do Maven local deploys for testing?

.github/workflows/release.yml
```diff
@@ -39,34 +27,13 @@ jobs:
           distribution: 'zulu'
           java-version: 19
 
-      - name: Prep docs
-        run: ./gradlew dokkaHtml
+      - name: Set release version
+        if: startsWith(github.ref, 'refs/tags/')
+        run: echo "VERSION_NAME=${GITHUB_REF#refs/tags/}" >> gradle.properties
```
JakeWharton, line 32: We had a setup like this in RxAndroid that was mostly forced on us from the RxJava Travis CI infrastructure. Personally, I was glad to see it go away when we moved to GitHub Actions and I greatly prefer the explicit version. In other projects I'm editing the README and CHANGELOG when I release to indicate the next version so I'm not sure whether there's a ton of value in automating only one location.

If you want to go down the automating route I would like to see a workflow that's manually dispatched where the version is in the input. It would automatically update the README, migrate the CHANGELOG entries from an "Unreleased" section to a versioned section, and do the `gradle.properties` update/tag/update dance.

.github/workflows/release.yml
Egorand: This is based on the discussion @swankjesse & I had - we wanted to have basic tag format validation, that would accept valid versions, even ones we don't normally use (e.g., "1.0-SQUARE"), and reject obviously incorrect ones (e.g., "hello").

gradle.properties
Egorand: It does, yep. This likely breaks more things, and will generally be non-ergonomic for local development.

.github/workflows/release.yml
Egorand: That makes sense.

Taking a step back, the main quality of life improvements I'd love to get from this change is not having to build the project locally (and risking using a wrong JDK or incorrect environment setup in general) for release, and not having to host signing keys on my machine - these can be achieved even without these extra workflow improvements. Given that this setup likely needs much more work to be pleasant to use, I feel like sticking with the basic requirements, and leaving the rest for later. Let me know if you have thoughts!

swankjesse: I don’t like having to do 2 commits per release. Changing the property for the current release and again for the next snapshot is a bit clumsy.

I also don’t like how the tag and the property duplicate information.

I’d like to discuss our options on a VC, maybe next week?

PR #1514, opened 2023-04-09
Return the generated path from FileSpec.writeTo().

### LINKED ISSUE #1719: Add FileSpec output path
**Is your feature request related to a problem? Please describe.**
I'm writing a gradle plugin to generate code from some source files, using gradle's `InputChanges` API. 

* When a source file is added, I create the FileSpec for that file and call `writeTo(outputDirectory)`. 
* When a source file is removed, I want to delete the corresponding generated file. However, I have to calculate the generated file's relative path myself, because `writeTo` hides the details of that. 

**Describe the solution you'd like**
An API like `FileSpec.relativePath` that gives the same relative path that `writeTo(File)` adds to the base directory.
Then I can call something like `outputDirectory.resolve(fileSpec.relativePath).delete()`

**Describe alternatives you've considered**
I can write my own `FileSpec` extension.

**Additional context**
The details aren't too complex to copy myself, I'm just worried that it could potentially get out of sync.

fejesjoco:
Hi,

I have a use case where it would be nice to know the path that is generated in the writeTo method.

This is similar to how JavaPoet's JavaFile.writeToPath() and JavaFile.writeToFile() work.

Note: I haven't updated all methods in FileWritingTest because I wanted to get a round of review first. I'll be happy to do the same to the rest of the file for consistency.

Closes #1719

### DIFF
--- a/kotlinpoet/api/kotlinpoet.api
+++ b/kotlinpoet/api/kotlinpoet.api
@@ -207,9 +207,11 @@ public final class com/squareup/kotlinpoet/FileSpec : com/squareup/kotlinpoet/An
 	public static synthetic fun toBuilder$default (Lcom/squareup/kotlinpoet/FileSpec;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/FileSpec$Builder;
 	public final fun toJavaFileObject ()Ljavax/tools/JavaFileObject;
 	public fun toString ()Ljava/lang/String;
-	public final fun writeTo (Ljava/io/File;)V
+	public final fun writeTo (Ljava/io/File;)Ljava/io/File;
+	public final synthetic fun writeTo (Ljava/io/File;)V
 	public final fun writeTo (Ljava/lang/Appendable;)V
-	public final fun writeTo (Ljava/nio/file/Path;)V
+	public final fun writeTo (Ljava/nio/file/Path;)Ljava/nio/file/Path;
+	public final synthetic fun writeTo (Ljava/nio/file/Path;)V
 	public final fun writeTo (Ljavax/annotation/processing/Filer;)V
 }
 
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
@@ -30,6 +30,7 @@ import javax.tools.JavaFileObject
 import javax.tools.JavaFileObject.Kind
 import javax.tools.SimpleJavaFileObject
 import javax.tools.StandardLocation
+import kotlin.DeprecationLevel.HIDDEN
 import kotlin.reflect.KClass
 
 /**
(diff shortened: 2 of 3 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -88,11 +91,15 @@ public class FileSpec private constructor(
...
   }
 
-  /** Writes this to `directory` as UTF-8 using the standard directory structure.  */
+  /**
+   * Writes this to `directory` as UTF-8 using the standard directory structure and returns the path
+   * of the output file.
+   */
   @Throws(IOException::class)
-  public fun writeTo(directory: File): Unit = writeTo(directory.toPath())
+  public fun writeTo(directory: File): File = writeTo(directory.toPath()).toFile()
```
fejesjoco, line 102: For reasons I don't understand, Path.toFile() may throw UnsupportedOperationException so technically this could be a regression even if nobody wants to know the return value. But JavaPoet does the same [here](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/JavaFile.java#L159-L160), and since we already had a File to begin with, I hope it should be fine.

JakeWharton:
This is a binary breaking change and this must be done as a new method with a JvmName that differs from the existing one and the existing one marked as hidden. That will retain binary compatibility and source compatibility for Kotlin (probably) but break source compatibility with Java callers, if there are any.

fejesjoco:
How about doing exactly what JavaPoet did: introduce new writeToPath/writeToFile methods which are the same as writeTo but with a return type. That way it should be source and binary compatible.

JakeWharton:
Note that we had no choice in JavaPoet. Kotlin affords us the choice of using hidden methods. I think the risk is probably worth it.

fejesjoco:
This doesn't seem to work: https://pl.kotl.in/7_gQxvl0X . Also I would find it strange if the Java and Kotlin API surfaces differ only be the name of a method.

lukellmann:
> This doesn't seem to work: https://pl.kotl.in/7_gQxvl0X . Also I would find it strange if the Java and Kotlin API surfaces differ only be the name of a method.

It is possible with something like this:

```kotlin
@Throws(IOException::class)
@Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
@JvmName("writeTo") // preserve the old name in the bytecode for binary compatibility
public fun oldWriteTo(directory: Path) { // change the Kotlin name so the compiler won't complain about conflicting overloads
    writeTo(directory) // resolves to non-hidden variant
}

@Throws(IOException::class)
public fun writeTo(directory: Path): Path {
    // ...
}
```

That way both Kotlin and Java callers (after recompilation) would call into the variant with return value while still preserving binary compatibility and without needing a name change.

To verify this works you can look at the API dump of the [Binary compatibility validator](https://github.com/Kotlin/binary-compatibility-validator):
```diff
-public final fun writeTo (Ljava/nio/file/Path;)V
+public final fun writeTo (Ljava/nio/file/Path;)Ljava/nio/file/Path;
+public final synthetic fun writeTo (Ljava/nio/file/Path;)V
```
The old variant was only changed to have the [`ACC_SYNTHETIC`](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.6) flag set (not a binary-incompatible change) and will therefore be inaccessible for newly compiled Java sources (see [`@JvmSynthetic`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-synthetic/)).

ebraminio:
I think #1730 is probably related to this.

PR #1529, opened 2023-04-23
Fix arguments handling in KSType.toTypeName()

### LINKED ISSUE #1513: KSType.toTypeName function returns ClassName for alias type instead of ParameterizedTypeName
**Describe the bug**
I expect calling `KSType.toTypeName()` on a type like `ArrayList<String>` to return `ParameterizedTypeName` with `rawType=kotlin.collections.List` and `typeArguments=[kotlin.String]`, but I get just raw `ClassName(kotlin.collections.List)` instead.

**To Reproduce**
Just change [this](https://github.com/square/kotlinpoet/blob/master/interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessor.kt#L133) line to `val parameterType = parameter.type.resolve().toTypeName(functionTypeParams).let {` and `smokeTest` will fail with something like this:
```
value of:
    readText$default(...)
diff (-expected +actual):
    @@ -103,7 +103,7 @@
         nullableSetListMapArrayNullableIntWithDefault: Set<List<Map<String, Array<IntArray?>>>>?,
         aliasedName: TypeAliasName,
         genericAlias: GenericTypeAlias,
    -    parameterizedTypeAlias: ParameterizedTypeAlias<String>,
    +    parameterizedTypeAlias: ParameterizedTypeAlias,
         nestedArray: Array<Map<String, Any>>?,
       ): Unit {
       }
	at com.squareup.kotlinpoet.ksp.test.processor.TestProcessorTest.smokeTest(TestProcessorTest.kt:146)
```

**Additional context**
Right now on a master branch `KSType.toTypeName` return non parameterized `ClassName` even for non aliased types and reproducer from above will break the with much more errors. This behaviour was changed in a #1321 and I don't know if this was intended.

ZacSweers:
This more or less undoes #1321 as it broke arguments handling in an unrelated way to annotations. I'm not able to actually repro the original issue it sought to solve in #1304, and added a regression test. This resolves #1513, though the examples there were not helpful either. All in all, I have a headache.

### DIFF
--- a/interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessor.kt
+++ b/interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessor.kt
@@ -22,8 +22,10 @@ import com.google.devtools.ksp.isConstructor
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.processing.SymbolProcessor
 import com.google.devtools.ksp.processing.SymbolProcessorEnvironment
+import com.google.devtools.ksp.symbol.ClassKind
 import com.google.devtools.ksp.symbol.KSAnnotated
 import com.google.devtools.ksp.symbol.KSClassDeclaration
+import com.squareup.kotlinpoet.ANY
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.ParameterSpec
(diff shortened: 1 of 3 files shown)


interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessor.kt
```diff
@@ -66,6 +68,32 @@ class TestProcessor(private val env: SymbolProcessorEnvironment) : SymbolProcess
...
+        val superclassReference = if (allSupertypes.isNotEmpty()) {
+          allSupertypes[0].takeIf {
+            val resolved = it.resolve()
+            resolved is KSClassDeclaration && resolved.classKind == ClassKind.CLASS
+          }
+        } else {
+          null
+        }
+        val superInterfaces = if (superclassReference != null) {
+          allSupertypes.drop(1)
```
Egorand, line 81: Are we guaranteed to have the superclass at a specific position in the `superTypes` list? Would it be safer to do `filterNot { it == superclassReference }`?

Egorand (approved):

interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessor.kt
ZacSweers: Good catch [bf481db](https://github.com/square/kotlinpoet/pull/1529/commits/bf481db9c82a4e2d273aa727fda8dac4a32eb7a4)

PR #1537, opened 2023-05-03
Split release jobs into a separate workflow

Egorand:
The existing workflow didn't get triggered by a new tag when I tried to release 1.13.1. Splitting release jobs into a separate file to have cleaner `on:` triggers, instead of writing more complex `if:` conditions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,54 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - "master"
+    tags: {}
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

Squiry:
Don't you need one more step for close+commit sonatype repository after uploading artefacts?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Release
+
+on:
+  push:
+    branches:
```
Squiry, line 5: I think `tags` part is enough for what you plan to do.

Egorand: The intent is to have this workflow cover both snapshots and releases: on a regular merge to master we'll fire this workflow and expect it to publish a snapshot (since the library version at that point is suffixed with `SNAPSHOT`), and on a new tag we'll fire it and expect it to publish a release.

Egorand:
> Don't you need one more step for close+commit sonatype repository after uploading artefacts?

Nope, it's already done by our publishing Gradle plugin. Here's the earlier PR that set it up: #1533.

PR #1558, opened 2023-05-22
Fix #1556 and #1557

### LINKED ISSUE #1525: Allow PropertySpec with context receivers and without getter or setter
Such code is valid in kotlin but fails to be generated by kotlinpoet with the error: `properties with context receivers require a getter`:

```kotlin
interface Bar {
   context(LoggingScope)
   val foo: Int
}
```

### LINKED ISSUE #1556: TypeSpec for Kotlin object can emit abstract functions
**Describe the bug**
The current validation logic in `TypeSpec.kt` does not prevent emitting the following code:

```kotlin
object Foo {
    abstract fun bar()
}
```

**To Reproduce**

Failing test:

```kotlin
  @Test fun abstractFunForbiddenInObject() {
    val type = TypeSpec.objectBuilder("Taco")

      type.addFunction(
        FunSpec.builder("eat")
          .addModifiers(ABSTRACT)
          .build(),
      )
        .build()
    }.hasMessageThat().isEqualTo("non-abstract type Taco cannot declare abstract function eat")
  }
```

**Expected behavior**
An exception similar to when we add an abstract function to a non-abstract class.

**Additional context**
I'm trying to fix #1525 which would involve moving validation logic from inside `PropertySpec.kt` into `TypeSpec.kt`. When I was doing this, I noticed some of the other logic in `TypeSpec.kt` has drifted.

### LINKED ISSUE #1557: TypeSpec for annotation can emit member functions and secondary constructors
**Describe the bug**
The current validation logic in `TypeSpec.kt` does not prevent emitting the following code:

```kotlin
public annotation class Taco() {
  public constructor(`value`: kotlin.String)
}
```

```kotlin
public annotation class Taco() {
  public fun eat(`value`: kotlin.String): kotlin.Unit
}
```

**To Reproduce**

Generates secondary constructor:
```kotlin
    TypeSpec.annotationBuilder("Taco")
      .primaryConstructor(
        FunSpec.constructorBuilder()
          .build(),
      )
      .addFunction(
        FunSpec.constructorBuilder()
          .addParameter("value", String::class)
          .addModifiers(PUBLIC, ABSTRACT)
          .build(),
        )
      .build()
```

Generates member function:
```kotlin
    TypeSpec.annotationBuilder("Taco")
      .primaryConstructor(
        FunSpec.constructorBuilder()
          .build(),
      )
      .addFunction(
        FunSpec.builder("eat")
          .addParameter("value", String::class)
          .addModifiers(PUBLIC, ABSTRACT)
          .build(),
        )
      .build()
```   

**Expected behavior**
Exceptions with a message like "annotation class Taco cannot declare member function eat"

**Additional context**
I'm trying to fix #1525 which would involve moving validation logic from inside `PropertySpec.kt` into `TypeSpec.kt`. When I was doing this, I noticed some of the other logic in `TypeSpec.kt` has drifted.

drawers:
Got bored at work so decided to look at `TypeSpec.kt` for #1556 and #1557 with the aim to eventually solve #1525

See also this comment:
https://github.com/square/kotlinpoet/issues/1556#issuecomment-1557620954

Closes #1556. Closes #1557.

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -441,7 +441,6 @@ public class TypeSpec private constructor(
 
     internal fun implicitFunctionModifiers(modifiers: Set<KModifier> = setOf()): Set<KModifier> {
       return defaultImplicitFunctionModifiers + when {
-        ANNOTATION in modifiers -> setOf(ABSTRACT)
         EXPECT in modifiers -> setOf(EXPECT)
         EXTERNAL in modifiers -> setOf(EXTERNAL)
         else -> emptySet()
@@ -771,8 +770,7 @@ public class TypeSpec private constructor(
         "typevariables are forbidden on anonymous types"
       }
 
-      val isAbstract = ABSTRACT in modifiers || SEALED in modifiers || kind != Kind.CLASS ||
-        !isSimpleClass
+      val isAbstract = ABSTRACT in modifiers || SEALED in modifiers || kind == Kind.INTERFACE || isEnum
       for (funSpec in funSpecs) {
         require(isAbstract || ABSTRACT !in funSpec.modifiers) {
           "non-abstract type $name cannot declare abstract function ${funSpec.name}"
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -771,8 +770,7 @@ public class TypeSpec private constructor(
         "typevariables are forbidden on anonymous types"
       }
 
-      val isAbstract = ABSTRACT in modifiers || SEALED in modifiers || kind != Kind.CLASS ||
```
drawers, line 774: `kind != Kind.CLASS` is the condition that is allowing objects to have abstract functions

JakeWharton (approved):

PR #1559, opened 2023-05-23
Do not emit superfluous `Unit` in return type position

### LINKED ISSUE #1540: `Unit`-returning function with block body can omit return type
Even in explicit API mode this is marked as a redundant declaration. Not true for expression bodies, however!

JakeWharton:
The only time it needs produced is when the function body is an expression in a non-ctor/getter/setter.

This contains a behavior change which explicitly makes the default return type `Unit`. This was somewhat the case before, except when you emitted a function body which could be turned into an expression. In this case you were able to get an implicit (i.e., unspecified) return type as none was declared. This would violate usages in explicit API contexts. A function with an omitted return type and a body which can be turned into an expression will now see an explicit `Unit` return type emitted. The fix is to declare the proper return type or remove the use of `return`.

Closes #1540

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
@@ -48,7 +48,7 @@ public class FunSpec private constructor(
   public val typeVariables: List<TypeVariableName> = builder.typeVariables.toImmutableList()
   public val receiverType: TypeName? = builder.receiverType
 
-  public val returnType: TypeName? = builder.returnType
+  public val returnType: TypeName = builder.returnType
   public val parameters: List<ParameterSpec> = builder.parameters.toImmutableList()
   public val delegateConstructor: String? = builder.delegateConstructor
   public val delegateConstructorArguments: List<CodeBlock> =
@@ -165,10 +165,8 @@ public class FunSpec private constructor(
       }
     }
 
-    if (returnType != null) {
+    if (returnType != UNIT || emitUnitReturnType()) {
       codeWriter.emitCode(": %T", returnType)
-    } else if (emitUnitReturnType()) {
-      codeWriter.emitCode(": %T", UNIT)
     }
 
     if (delegateConstructor != null) {
@@ -217,10 +215,7 @@ public class FunSpec private constructor(
   /**
    * Returns whether [Unit] should be emitted as the return type.
    *
-   * [Unit] is emitted as return type on a function unless:
-   *   - It's a constructor
-   *   - It's a getter/setter on a property
-   *   - It's an expression body
+   * [Unit] is emitted as return type on a function only if it is an expression body.
    */
   private fun emitUnitReturnType(): Boolean {
     if (isConstructor) {
@@ -231,7 +226,7 @@ public class FunSpec private constructor(
       return false
     }
 
-    return body.asExpressionBody() == null
+    return body.asExpressionBody() != null
   }
 
   private fun CodeBlock.asExpressionBody(): CodeBlock? {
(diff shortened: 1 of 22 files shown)


kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -388,12 +388,13 @@ class FunSpecTest {
...
   @Test fun expressionBodyIsDetectedReturnWithNonBreakingSpace() {
     val funSpec = FunSpec.builder("foo")
+      .returns(INT)
       .addStatement("return·1")
       .build()
 
     assertThat(funSpec.toString()).isEqualTo(
       """
-      |public fun foo() = 1
+      |public fun foo(): kotlin.Int = 1
```
JakeWharton, line 397: Here is an example of the behavior change.

This function previously had no return type and was relying on the fact that KotlinPoet would turn it into an expression function. If we turned that behavior off (which should be something in our power, as it shouldn't change the compiled result) the code would fail to compile.

With this behavior change, the code will now be emitted like this:
```kotlin
public fun foo(): kotlin.Unit = 1
```
which will also fail to compile. We are basically giving the expression form the same semantics as the non-expression form.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -48,7 +48,7 @@ public class FunSpec private constructor(
   public val typeVariables: List<TypeVariableName> = builder.typeVariables.toImmutableList()
   public val receiverType: TypeName? = builder.receiverType
 
-  public val returnType: TypeName? = builder.returnType
+  public val returnType: TypeName = builder.returnType
```
JakeWharton, line 51: This is binary compatible. It is also source compatible, although may produce some new warnings if you're performing a null check on it.

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
Egorand: I think this makes sense. Having to set the return type explicitly also protects from cases where you add an extra statement to the function body, which turns it from expression to block and breaks compilation.

The behaviour change is definitely something we'll want to call out in the changelog. Is it a good time to add the "unreleased" section?

Egorand (approved):

kotlinpoet/src/test/java/com/squareup/kotlinpoet/FunSpecTest.kt
JakeWharton: Yep, added!

JakeWharton: It is not supported, and any historical support was accidental and considered a bug.

Egorand: KotlinPoet doesn't support automatic type inference. There are several closed feature requests you can look up where we provide detailed justification for this design decision.

JakeWharton: We do not guarantee the use of an expression body as the output and so we require that you write a function body with an explicit return type and `return` statement. Whether or not the emitted code uses an expression body when producing the output is an implementation detail of the library.

m0ne: Alright, good to know. I personally would have enjoyed an option for this ‘functionality’, but I respect your decision. Thanks for the fast replies.

Issue #1554, opened 2023-05-19
Better generic support?


Omico:
Previously, I had a hard time finding how to use `parameterizedBy` to generate a generic parameter with `*` because the docs are missing the part. Luckily, I found it in the test code. Then, I have more questions. Why should we do that?https://github.com/square/kotlinpoet/blob/17c20110014b570a1abea56077e86070983a13f2/kotlinpoet/src/test/java/com/squareup/kotlinpoet/ParameterizedTypeNameTest.kt#L119-L126

Luckily I found another example.
https://github.com/square/kotlinpoet/blob/17c20110014b570a1abea56077e86070983a13f2/kotlinpoet/src/test/java/com/squareup/kotlinpoet/TypeSpecTest.kt#L95

Seems we are missing functions like the ones below. Usage `KClass::class.parameterizedBy(STAR)`

```kotlin
    /** Returns a parameterized type, applying `typeArguments` to `this`. */
    @JvmStatic
    @JvmName("get")
    public fun KClass<*>.parameterizedBy(
      typeArguments: Iterable<TypeName>,
    ): ParameterizedTypeName = ParameterizedTypeName(null, asClassName(), typeArguments.toList())

    /** Returns a parameterized type, applying `typeArguments` to `this`. */
    @JvmStatic
    @JvmName("get")
    public fun KClass<*>.parameterizedBy(
      vararg typeArguments: TypeName,
    ): ParameterizedTypeName = ParameterizedTypeName(null, asClassName(), typeArguments.toList())

```
Do I need to send a PR for this? Or did I miss anything?

JakeWharton:
Generally we don't mix and match representations. So KClass is Kotlin reflect and TypeName is KotlinPoet which is why you can't mix the two. Instead, first you would convert the KClass into a ClassName and then you can manipulate it with things like parameterizedBy.

Omico:
Ok, thank you for your answer. I really hope we can add the usage of generic with `*` into docs. It cost me at least 20 mins to dig through the code and find out how to use it. 😂

Egorand:
Wanna send a PR that adds those docs?

Omico:
> Wanna send a PR that adds those docs?

Sorry for the late reply, I have submitted a patch, please review it.

PR #1572, opened 2023-05-25
Add usage of STAR

Omico:
Improvement of #1554.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -315,6 +315,7 @@ KotlinPoet has classes for building each of these:
 
 ```kotlin
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.STAR
 
 val hoverboard = ClassName("com.mattel", "Hoverboard")
 val list = ClassName("kotlin.collections", "List")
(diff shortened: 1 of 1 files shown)


JakeWharton:
I think if we're going to start expanding the docs I would prefer a dedicate page on modeling types. This would cover creating classes, parameterized types, wildcard types, parsing types from reflection/metadata/strings, referencing other generated types, using types to create specs, etc.

The giant index page is a holdover from when all this content was in the README and it doesn't quite translate to the web.

docs/index.md
```diff
@@ -339,8 +340,15 @@ val printThings = FunSpec.builder("printThings")
...
   .addStatement("println(things)")
   .build()
+
+val printKClass = FunSpec.builder("printKClass")
+  .addParameter("kClass", KClass::class.asClassName().parameterizedBy(STAR))
+  .addStatement("println(kClass)")
+  .build()
 ```
 
+The `STAR` is represented as `*` in KotlinPoet. You can find more in [`com.squareup.kotlinpoet.TypeName`](https://github.com/square/kotlinpoet/blob/master/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeName.kt)
```
Egorand, line 350: I think it's better to link to KDoc rather than sources. Plus, linking to an entire file might not be that helpful.

Omico:
> I think if we're going to start expanding the docs I would prefer a dedicate page on modeling types. This would cover creating classes, parameterized types, wildcard types, parsing types from reflection/metadata/strings, referencing other generated types, using types to create specs, etc.
> 
> The giant index page is a holdover from when all this content was in the README and it doesn't quite translate to the web.

I completely agree with you. We can make further improvements in other PRs.

ZacSweers (approved):

PR #1575, opened 2023-05-25
Allow properties with context receivers in interfaces to omit declaring accessors

### LINKED ISSUE #1525: Allow PropertySpec with context receivers and without getter or setter

drawers:
Closes #1525 

* Move check for getters and setters on properties with context receivers into `TypeSpec.kt`
* Disallow abstract properties in non-abstract types

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt
@@ -56,12 +56,6 @@ public class PropertySpec private constructor(
     require(mutable || setter == null) {
       "only a mutable property can have a setter"
     }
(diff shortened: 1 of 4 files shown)


Egorand (approved):
Thanks!

Too bad we can't perform all of these checks when creating the `PropertySpec`, but it seems that we don't have enough context until we add that property to a `TypeSpec`. We can perform part of the checks inside `PropertySpec`'s constructor (i.e. has an `abstract` modifier, context receivers and no getters/setters - fail), but not sure whether there's value in splitting these checks between multiple types and thus complicating logic.

PR #1580, opened 2023-05-29
Restore ABI stability for annotatable and documentable builders

### LINKED ISSUE #1579: 1.14.0 has breaking ABI changes to documentable and annotatable builders
**Describe the bug**
It seems that the traits (or how they mix with the new JVM default behavior) broke binary compatibility

```
> Task :libraries:summarizer:kspReleaseKotlin FAILED
e: java.lang.NoSuchMethodError: 'com.squareup.kotlinpoet.TypeSpec$Builder com.squareup.kotlinpoet.TypeSpec$Builder.addAnnotation(com.squareup.kotlinpoet.AnnotationSpec)'
        at com.slack.circuit.codegen.CircuitSymbolProcessor.generateFactory(CircuitSymbolProcessorProvider.kt:154)
        at com.slack.circuit.codegen.CircuitSymbolProcessor.process(CircuitSymbolProcessorProvider.kt:111)
        at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$6$1.invoke(KotlinSymbolProcessingExtension.kt:291)
```

ZacSweers:
Resolves #1579

Originally tried to be clever by using value classes to get the impl without the overhead, and it _almost_ worked except that the resulting bytecode still tried to cast the underlying `apply` to the value class instead of the enclosing builder it was inlined into. Fell back to a more manual approach of just adding explicit overrides that just called super.

### DIFF
--- a/kotlinpoet/api/kotlinpoet.api
+++ b/kotlinpoet/api/kotlinpoet.api
@@ -332,6 +332,10 @@ public final class com/squareup/kotlinpoet/FunSpec$Builder : com/squareup/kotlin
 	public final fun addCode (Lcom/squareup/kotlinpoet/CodeBlock;)Lcom/squareup/kotlinpoet/FunSpec$Builder;
 	public final fun addCode (Ljava/lang/String;[Ljava/lang/Object;)Lcom/squareup/kotlinpoet/FunSpec$Builder;
 	public final fun addComment (Ljava/lang/String;[Ljava/lang/Object;)Lcom/squareup/kotlinpoet/FunSpec$Builder;
+	public synthetic fun addKdoc (Lcom/squareup/kotlinpoet/CodeBlock;)Lcom/squareup/kotlinpoet/Documentable$Builder;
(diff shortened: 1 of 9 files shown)


Egorand (approved):
Hmm, really confused why removing these redundant overrides breaks binary compatibility..

JakeWharton:
Because the generic return type T winds up in bytecode as its lower bound which is the interface Builder rather than the concrete spec Builder.

Also the binary compatibility validator seems to be a complete garbage tool and doesn't fail the build when you do this. Do we need to switch to metalava or japicmp, or does it just have terrible default behavior?

ZacSweers:
I think it focuses only on being a change detector and relies on code review to decide if the change is ok. It doesn't do what japicmp does as far as "is this compatible with a previous release" kind of checks, but it's also more accurate in some ways. Maybe there's a mode like that that I'm missing though

ZacSweers:
Filed a FR here: https://github.com/Kotlin/binary-compatibility-validator/issues/140

Egorand:
Should we set up japicmp in the meantime?

ZacSweers:
Seems like a good idea

PR #1589, opened 2023-06-04
Deprecate TypeSpec.expectClassBuilder() and TypeSpec.valueClassBuilder()

### LINKED ISSUE #1385: No TypeSpec.expectObjectBuilder
We've got `TypeSpec.objectBuilder` and `TypeSpec.expectClassBuilder` but there is no companion method that returns `Builder(Kind.OBJECT, name, EXPECT)`.

You can work around it with `TypeSpec.objectBuilder(className).addModifiers(KModifier.EXPECT)` but the API does feel inconsistent because of it.

drawers:
Closes #1385

### DIFF
--- a/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
@@ -886,14 +886,14 @@ public class TypeSpec private constructor(
     @JvmStatic public fun classBuilder(className: ClassName): Builder = classBuilder(className.simpleName)
 
     @Deprecated(
-      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0",
+      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0.",
       ReplaceWith("TypeSpec.classBuilder(name).addModifiers(KModifier.EXPECT)"),
     )
     @JvmStatic
     public fun expectClassBuilder(name: String): Builder = Builder(Kind.CLASS, name, EXPECT)
 
     @Deprecated(
-      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0",
+      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0.",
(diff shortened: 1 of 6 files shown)


kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -883,22 +883,32 @@ public class TypeSpec private constructor(
...
+    public fun expectClassBuilder(className: ClassName): Builder = classBuilder(className.simpleName).addModifiers(EXPECT)
 
-    @JvmStatic public fun valueClassBuilder(name: String): Builder =
-      Builder(Kind.CLASS, name, VALUE)
+    @Deprecated(
+      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0.",
+      ReplaceWith("TypeSpec.classBuilder(name).addModifiers(KModifier.VALUE)"),
+    )
+    @JvmStatic
+    public fun valueClassBuilder(name: String): Builder = Builder(Kind.CLASS, name, VALUE)
```
drawers, line 907: This function wasn't mentioned explicitly in the issue comments for #1385 but I assume the intention is to reduce the API surface so I deprecated it too.

kotlinpoet/src/main/java/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -883,22 +883,27 @@ public class TypeSpec private constructor(
...
 
-    @JvmStatic public fun classBuilder(className: ClassName): Builder =
-      classBuilder(className.simpleName)
+    @JvmStatic public fun classBuilder(className: ClassName): Builder = classBuilder(className.simpleName)
 
-    @JvmStatic public fun expectClassBuilder(name: String): Builder =
-      Builder(Kind.CLASS, name, EXPECT)
+    @Deprecated(
+      "Use classBuilder() instead. This function will be removed in KotlinPoet 2.0",
+      ReplaceWith("TypeSpec.classBuilder(name).addModifiers(KModifier.EXPECT)"),
```
Egorand, line 890: Nice!

Egorand (approved):
Thanks!

PR #1597, opened 2023-06-12
Revert soft keyword escaping

Egorand:
A manual revert of #994.


adrw (approved):

swankjesse (commented):
Instead of this, could we do something different in NameAllocator ? Like detect soft keywords and backtick them?

JakeWharton:
The two behaviors are intrinsically linked. The list of keywords determines both which are backticked when emitted and which get underscore prefixes from `NameAllocator`. It's actually somewhat of a surprising behavior.

Egorand:
We won't do this. Instead, we exposed API to control `NameAllocator`'s behaviour w.r.t. keywords: #1803.

PR #1625, opened 2023-07-12
Use backticks for keywords in NameAllocator

Egorand:


JakeWharton:
One problem with this is that if you use the name in an output somewhere it will become backticked. You can see this in your Wire PR where the `toString()` becomes `` `when`=blah `` instead of `when=blah`.

JakeWharton:
Can we return "when" and require the use of `%N` which does(?) the backtick escaping?

Egorand:
Yep, `%N` does backticks. Do you mean require its use in Wire? I think it's worth trying!

JakeWharton:
> Do you mean require its use in Wire?

I more meant in general. If we start returning "when" directly, then people using string templating will start to get code that doesn't compile. Are we comfortable saying "You should have been using/must use %N"?

Egorand:
I think eventually (in 2.0 probably?) we should do that. To avoid breaking current behaviour though, @swankjesse & I decided to introduce API to control how keywords are treated: #1803.

Egorand:
Closed in favour of #1803.

PR #1627, opened 2023-07-13
Update kotlinx-metadata to 0.7.0

ZacSweers:
Following the guidance here: https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/Migration.md#migrating-from-06x-to-070

This removes a lot of our preview helper APIs as most of them are part of the first party APIs now


interop/kotlinx-metadata/api/kotlinx-metadata.api
```diff
@@ -1,97 +1,3 @@
-public final class com/squareup/kotlinpoet/metadata/FlagsKt {
```
Egorand, line 1: Is it worth it being super diligent and deprecating all of these with `@ReplaceWith`?

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ElementsClassInspector.kt
```diff
@@ -333,8 +332,7 @@ public class ElementsClassInspector private constructor(
 
     val propertyData = declarationContainer.properties
       .asSequence()
-      .filter { it.isDeclaration }
-      .filterNot { it.isSynthesized }
```
Egorand, line 337: No need for this `filterNot` anymore?

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ReflectiveClassInspector.kt
```diff
@@ -319,8 +318,7 @@ public class ReflectiveClassInspector private constructor(
 
     val propertyData = declarationContainer.properties
       .asSequence()
-      .filter { it.isDeclaration }
-      .filterNot { it.isSynthesized }
```
Egorand, line 323: Same here.

Egorand (approved):

interop/kotlinx-metadata/api/kotlinx-metadata.api
ZacSweers: I don't think so, especially as many of them are just import changes

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/classinspectors/ElementsClassInspector.kt
ZacSweers: Nope, one thing learned in this is that this is mutually exclusive with declaration/delegation/etc

Egorand: Nice

interop/kotlinx-metadata/api/kotlinx-metadata.api
Egorand: Might make sense to call out the deletions in the release notes.

PR #1631, opened 2023-07-14
Replace occurrences of 'until' with '..<'

Egorand:


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
```diff
@@ -64,7 +64,7 @@ private fun isDefaultValue(value: Any?, defaultValue: Any?): Boolean {
     }
   }
   if (value is List<*> && defaultValue is List<*>) {
-    return value.size == defaultValue.size && (0 until defaultValue.size).all { index ->
+    return value.size == defaultValue.size && (0..<defaultValue.size).all { index ->
```
JakeWharton, line 67:
```suggestion
    return value.size == defaultValue.size && defaultValue.indices.all { index ->
```

JakeWharton (approved):

PR #1654, opened 2023-08-04
Switching to Kotlin Multiplatform for JVM use

takahirom:
#304

This Pull Request aims to transition our codebase from org.jetbrains.kotlin.jvm to org.jetbrains.kotlin.multiplatform with JVM configuration. The primary intention behind this shift is to establish a foundation for adaptation to other platforms in the future.

Additionally, I encountered an issue when trying to run Dokka with a single-target multiplatform project using Kotlin 1.9.0. As a workaround, I've had to comment out the Dokka-related processes for now. You can find more information about this issue here: https://github.com/Kotlin/dokka/issues/3038

Please review these changes, and let me know if there are any concerns or suggestions. Thanks!


build.gradle.kts
```diff
@@ -42,6 +43,11 @@ allprojects {
 }
 
 subprojects {
+  apply(plugin = "org.jetbrains.kotlin.multiplatform")
+  configure<KotlinMultiplatformExtension> {
+    jvm {}
+  }
```
JakeWharton, line 49: Please apply and configure each project individually. I believe only the core artifact should be multiplatform. The others should all remain on `.jvm` plugin as they are fundamentally tied to the JVM in some way (KSP, JavaPoet, etc.).

JakeWharton (commented):
I'm fine with this as a starting point, but I would prefer it only be done for the main artifact.

JakeWharton:
I'll also note that the biggest inhibitor to doing this work already is needing something like https://youtrack.jetbrains.com/issue/KT-34668 or https://youtrack.jetbrains.com/issue/KT-20427. I tried migrating the project many years ago and found the duplication/indirection needed to be too much to bear.

build.gradle.kts
takahirom: I've made the changes, but it seems like we're encountering issues with sections like `configure<KotlinProjectExtension>` and `tasks.withType<KotlinCompile>`, which fail due to the absence of the plugin. We might need to resort to code duplication, or we could establish a common configuration using `build-logic`. Do you have a preference for how we should proceed?

takahirom: Sorry, I might be able to use `plugins.withId`.

takahirom:
@JakeWharton 
Just to confirm, when you say 'the main artifact', are you referring to the ':kotlinpoet' module? So, should I apply the Kotlin Multiplatform plugin to the ':kotlinpoet' module and leave the other modules as they are?

takahirom:
I have applied the KMP plugin exclusively to the kotlinpoet module. Please review the changes and let me know if there are any concerns or suggestions for improvement.

takahirom:
I don't know why, but it seems that the format has not been applied to the main branch.
I'm not sure if I should fix this or use excludeTarget("kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/ClassName.kt") as a spotless setting. 👀

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kotlinpoet:spotlessKotlinCheck'.
> The following files had format violations:
      src/commonMain/kotlin/com/squareup/kotlinpoet/ClassName.kt
          @@ -1,5 +1,5 @@
           /*
8 actionable tasks: 8 executed
          -·*·Copyright·(C)·2014·Google,·Inc.
          +·*·Copyright·(C)·2014·Square,·Inc.
```

JakeWharton:
Add the exclude, please. Basically assume whatever is in the file currently is correct.

takahirom:
Oh, sorry, I overlooked the 'targetExclude' in the kotlinpoet module. I've fixed it since the source set name changed.
https://github.com/square/kotlinpoet/pull/1654/commits/665b933e90d22cd5d2e8eebcff0923254823e78a

takahirom:
JavaClassWithArrayValueAnnotation is a Java file, and the test can't find it since it hasn't been compiled. Integrating this test into the Kotlin Multiplatform module is a bit challenging, so I propose we delete it.

takahirom:
TypeNameTest and JavaClassWithArrayValueAnnotation are Java classes in the test. We can separate the Java test module. But it may take some maintenance costs. What do you think?

takahirom:
I was able to run the Java-related tests.
I didn't know that we could compile Java code in jvmTest/java.
image

JakeWharton (approved):
So this all looks great, but I'm not sure we can land it without figuring out the Dokka situation because that's a bit of a blocker.

takahirom:
I've applied the workaround for Dokka from this link: https://github.com/Kotlin/dokka/issues/3122. Now, we can view the Dokka report.
image

Ayfri:
Any update ?

JakeWharton:
Note that merging this does not immediately change the supported targets. It's still JVM only.

JakeWharton (approved):

kotlinpoet/build.gradle.kts
```diff
@@ -13,40 +13,66 @@
...
+    withJava()
+  }
+
+  // Check if the current running task is dokkaHtml
+  val isDokkaHtml = project.gradle.startParameter.taskNames.any { it.contains("dokkaHtml") }
+  if (isDokkaHtml) {
+    // TODO: Remove the phantom target after https://github.com/Kotlin/dokka/issues/3122
+    js()
+  }
+
```
JakeWharton, line 44:
```suggestion
```
Seems fixed in Dokka 1.9.10

gradle/libs.versions.toml
```diff
@@ -19,6 +19,7 @@ ksp = "1.9.10-1.0.13"
 ktlint = "0.48.2"
 
 [plugins]
+kotlin-multiplatform = { id = "org.jetbrains.kotlin.multiplatform", version.ref = "kotlin" }
 kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
 dokka = { id = "org.jetbrains.dokka", version = "1.8.20" }
```
JakeWharton, line 24:
```suggestion
dokka = { id = "org.jetbrains.dokka", version = "1.9.10" }
```

JakeWharton (approved):

kotlinpoet/build.gradle.kts
```diff
@@ -13,40 +13,59 @@
...
+  manifest {
+    attributes("Automatic-Module-Name" to "com.squareup.kotlinpoet")
+  }
 }
+
+tasks.named<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>("compileTestKotlinJvm") {
+  compilerOptions {
+    freeCompilerArgs.add("-opt-in=com.squareup.kotlinpoet.DelicateKotlinPoetApi")
+  }
+}
```
JakeWharton, line 70: There's a DSL for this we can use in a follow-up

https://github.com/cashapp/molecule/blob/782985fcf92d6385d164bbed91184f99b1f9141e/molecule-runtime/build.gradle#L42-L46

Issue #279, opened 2017-11-10
Using kotlin mutable collections for typenames


AlecStrong:
If I do

```
MutableList::class.asTypeName()
```

the generated code uses `kotlin.collections.List`

I have worked around with `ClassName("kotlin.collections", "MutableList")`

swankjesse:
This might be a problem with Kotlin itself.

```
    assertThat(MutableList::class.qualifiedName).isEqualTo("kotlin.collections.MutableList")
```

```
org.junit.ComparisonFailure: 
Expected :kotlin.collections.MutableList
Actual   :kotlin.collections.List
```

swankjesse:
... looks like `MutableList` is completely erased at compile-time.

```
assertThat(MutableList::class).isEqualTo(List::class)
```

AlecStrong:
should this be something kotlinpoet addresses then? `List::class.asMutableTypeName()` or better to just leave it as the manual `ClassName` usage

given your findings its probably impossible to have something thats safe at compile time

swankjesse:
`asMutableTypeName()` is an interesting idea!

afollestad:
Yeah something like...
```kotlin
internal fun KClass<*>.asMutableTypeName(): TypeName {
  return when (this) {
    List::class -> ClassName("kotlin.collections", "MutableList")
    Map::class -> ClassName("kotlin.collections", "MutableMap")
    else -> error("Unknown type: $this")
  }
}
```
...is handy

Egorand:
@afollestad btw we've introduced a large number of constants for frequently used `ClassName`s:
```kotlin
fun KClass<*>.asMutableTypeName(): ClassName {
  return when (this) {
    List::class -> MUTABLE_LIST
    Map::class -> MUTABLE_MAP
    Set::class -> MUTABLE_SET
    else -> asTypeName()
  }
}
```

ZacSweers:
This, annoyingly, isn't covered by the new `typeOf()` API either :/

```kotlin
@Test
fun typeNameOf_mutable() {
  val type = typeNameOf<MutableList<String>>()
  assertThat(type.toString()).isEqualTo("kotlin.collections.MutableList<kotlin.String>")
}
```

yields
```
expected: kotlin.collections.MutableList<kotlin.String>
but was : kotlin.collections.List<kotlin.String>
```

ZacSweers:
This is filed now https://youtrack.jetbrains.com/issue/KT-35877

[...]

PR #1661, opened 2023-08-15
Test expected to fail

oldergod:
I was expecting the test to fail but no.


oldergod:
on the JVM both List and MutableList are typealiases to `java.util.List`

JakeWharton:
#279

PR #1688, opened 2023-09-05
Un-pin Markdown version

Egorand:
It was previously pinned by mkdocs (https://github.com/mkdocs/mkdocs/issues/2892), but was unpinned in https://github.com/mkdocs/mkdocs/commit/562d5e14c1d2c914d42942ab2385822c04d5bc7b. Don't think we're using it directly, so deleting it altogether.


oldergod (approved):
🚀

Issue #1553, opened 2023-05-19
Common traits on builders for polymorphism


JakeWharton:
It's really hard to extract code which can operate on two different spec builders.

Off the top of my head, I strongly need:
- [x] The ability to add annotations (types, properties, functions)
  - #1564  
- [x] The ability to add kdoc (types, properties, functions)
  - #1571 

Other probable inclusions:
- [x] The ability to add types (files, type)
  - #1723
- [x] _Maybe_ the ability to add functions and properties (file, type)
  - https://github.com/square/kotlinpoet/pull/1867 
- [ ] The ability to add code (CodeBlock, fun spec)

I don't think we need a full class hierarchy similar to PSI/FIR, but trait-like interfaces for these fairly primitive operations would be welcome. Maybe named like `AnnotatedSpecBuilder`, `DocumentedSpecBuilder`, etc.

ZacSweers:
We've done a bit of stuff around Taggable and Context Receivers. Could that be a good starting point?

JakeWharton:
I've never used either and don't recall their implementations. So, maybe?

Egorand:
Last bit implemented in #2331, closing.

PR #1723, opened 2023-11-16
Extract interface for constructs that can hold a TypeSpec and their builders

drawers:
Part of #1553

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpecHolder.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpecHolder.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpecHolder.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+/** A spec which can contain other [TypeSpec] */
+public interface TypeSpecHolder {
```
drawers, line 19: Name follows convention in the project `OriginatingElementsHolder`

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpecHolder.kt
```diff
@@ -0,0 +1,31 @@
...
+  public val typeSpecs: List<TypeSpec>
+
+  public interface Builder<out T : Builder<T>> {
+
+    public fun addType(typeSpec: TypeSpec): T
```
drawers, line 24: Because we haven't provided a default implementation for `addType`, this interface less like a trait than `Documentable` and `Annotatable`. 

It also doesn't achieve a mutable property `val typeSpecs: MutableList<TypeSpec>` like `Documentable` and `Annotatable`

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpecHolder.kt
JakeWharton: But differs from `Annotatable`. Honestly I've been mulling this change because I couldn't think of an `-able` name. Probably should have gone with `AnnotationHolder`...

JakeWharton (approved):
Very nice!

PR #1727, opened 2023-11-18
Remove superfluous newline

### LINKED ISSUE #1725: Regression in KDoc generation whitespace
I think as a result of #1637, we unconditionally add a newline after the docs even if no tags are going to be emitted.

This is causing a failure in SQL Delight here https://github.com/cashapp/sqldelight/pull/4823

JakeWharton:
Closes #1725 

It's pretty amazing this wasn't tested anywhere (even implicitly).

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
@@ -380,25 +380,24 @@ public class TypeSpec private constructor(
    * property/parameter, and the parameter's KDoc will be printed in the type header.
    */
   private fun kdocWithConstructorDocs(): CodeBlock {
-    if (primaryConstructor == null) {
-      return kdoc.ensureEndsWithNewLine()
-    }
-    val constructorProperties = constructorProperties()
-    val parametersWithKdoc = primaryConstructor.parameters.filter { parameter ->
-      val propertyKdoc = constructorProperties[parameter.name]?.kdoc ?: CodeBlock.EMPTY
-      return@filter parameter.kdoc.isNotEmpty() && propertyKdoc.isNotEmpty() &&
-        parameter.kdoc != propertyKdoc
-    }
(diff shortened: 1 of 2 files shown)


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -390,12 +390,16 @@ public class TypeSpec private constructor(
...
+        }
+        if (primaryConstructor.kdoc.isNotEmpty()) {
+          add("@constructor %L", primaryConstructor.kdoc.ensureEndsWithNewLine())
+        }
+        parametersWithKdoc.forEach { parameter ->
+          add("@param %L %L", parameter.name, parameter.kdoc.ensureEndsWithNewLine())
+        }
       }
       build()
     }
```
Egorand, line 405: I wanted to avoid a thousand `if`s so did this, see if you like it better (sadly can't add as a suggestion):

```kotlin
val constructorKdoc = buildCodeBlock {
  if (primaryConstructor.kdoc.isNotEmpty()) {
    add("@constructor %L", primaryConstructor.kdoc.ensureEndsWithNewLine())
  }
  parametersWithKdoc.forEach { parameter ->
    add("@param %L %L", parameter.name, parameter.kdoc.ensureEndsWithNewLine())
  }
}
return listOf(kdoc.ensureEndsWithNewLine(), constructorKdoc)
  .filter(CodeBlock::isNotEmpty)
  .joinToCode(separator = "\n")
```

Egorand (approved):

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: Ship it

JakeWharton: Ended up pulling even more into this. Because the code above also had conditionals and functional list operations. Combined everything as best I could with minimal conditional logic.

PR #1742, opened 2023-11-27
ksp: more accurately represent function types

evant:
When converting a KSTypeReference, represent a function type with LambdaTypeName. The better matches the source code and allows for receiver types (ex: String.() -> Unit) to be represented correctly.

### DIFF
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/KsTypes.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/KsTypes.kt
@@ -15,6 +15,7 @@
  */
 package com.squareup.kotlinpoet.ksp
 
+import com.google.devtools.ksp.symbol.KSCallableReference
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSType
 import com.google.devtools.ksp.symbol.KSTypeAlias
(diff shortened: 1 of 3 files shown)


ZacSweers (approved):
Nice improvement 👍

One open question (that may be annoying to test) – do you know what happens when you have a function with more than 22 parameters? See https://github.com/Kotlin/KEEP/blob/master/proposals/functional-types-with-big-arity-on-jvm.md

evant:
> Or are we just going to safely assume that since they're otherwise plain 'ol types.

Yep they are treated just like any other type (just verified locally). No strong opinions here so I wouldn't mind adding to the test if we feel it's useful.

evant:
> One open question (that may be annoying to test) – do you know what happens when you have a function with more than 22 parameters?

Seems to generate fine if that's what you are asking, I could add something to the test if you want?

ZacSweers:
Yeah let's add a test just for coverage if you don't mind, otherwise lgtm 👍

evant:
Sorry, I'm a bit confused on what we'd actually be testing with that. From a codegen perspective nothing changes at 23+ params. According to the proposal you linked the generated bytecode is different but I don't think that smoke test is testing that anyway?

ZacSweers:
I updated it with what I meant. You're right, I just wanted to be sure :). I was curious if KSP would render that as a KSCallableReference still or if it would represent it as a `FunctionN` with arity

evant:
heh ok

JakeWharton:
Thanks!

PR #1756, opened 2023-12-02
fix nullability of lambdas in KSTypeReference.toTypeName

gabrielittner:
With the change of how lambdas are converted to `TypeName` in 1.15.2 the nullability of the converted reference is lost. This fixes it and makes the behavior match the `else` case (`KSType.toTypeName`) which also preserves nullability.


ZacSweers (approved):
Good catch, thanks!

PR #1763, opened 2023-12-04
Add Maven Central repo URL to RELEASING.md

Egorand:
I always check this URL to see whether the release has been published, and always forget what the URL is, so putting it here.

Is this the best URL by the way?


JakeWharton (approved):
I use repo1.maven.org, but I don't think it matters. Not sure which is canonical.

You should get yourself a https://github.com/JakeWharton/dependency-watch and never need to manually check again!

Egorand:
Oh nice!

PR #1790, opened 2023-12-29
Only run file system-related tests on Mac and Windows

### LINKED ISSUE #1721: Only run `:kotlinpoet:tests` on mac and windows
There's no need to do a full build on these machines (and tests from other modules). We only want to test filesystem-related interactions.

drawers:
Fixes #1721

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,9 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+        include:
+          - os: ubuntu-latest
+            full-build: true
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+        include:
+          - os: ubuntu-latest
+          - full-build: true
```
drawers, line 24: Followed the example of expanding configurations here:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations

drawers:
The `!` in GitHub Action expressions is trickier I thought - will spend some more time on this later.

PR #1791, opened 2023-12-30
Only run file system-related tests on Mac and Windows 

### LINKED ISSUE #1721: Only run `:kotlinpoet:tests` on mac and windows

drawers:
Fixes #1721

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,8 @@ jobs:
         run: ./gradlew build
 
       - name: File system check
-        if: matrix.os != 'ubuntu-latest
-        run: ./gradlew :kotlinpoet:allTests
+        if: "matrix.os != 'ubuntu-latest'"
+        run: ./gradlew :kotlinpoet:check
 
   build-docs:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,6 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        include:
-          - os: ubuntu-latest
-            full-build: true
 
     runs-on: ${{ matrix.os }}
 
@@ -36,11 +33,11 @@ jobs:
           java-version: 19
 
       - name: Full build
-        if: matrix.full-build
+        if: matrix.os == 'ubuntu-latest'
         run: ./gradlew build
 
       - name: File system check
-        if: "!matrix.full-build"
+        if: matrix.os != 'ubuntu-latest
         run: ./gradlew :kotlinpoet:allTests
 
   build-docs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: ./gradlew build
 
-      - name: File system check
+      - name: KotlinPoet check
         if: "matrix.os != 'ubuntu-latest'"
         run: ./gradlew :kotlinpoet:check
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,9 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+        include:
+          - os: ubuntu-latest
+            full-build: true
 
     runs-on: ${{ matrix.os }}
 
@@ -32,8 +35,13 @@ jobs:
           distribution: 'zulu'
           java-version: 19
 
-      - name: Test
+      - name: Full build
+        if: matrix.full-build
         run: ./gradlew build
+
+      - name: File system check
+        if: "!matrix.full-build"
+        run: ./gradlew :kotlinpoet:allTests
 
   build-docs:
     runs-on: ubuntu-latest


.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+        include:
+          - os: ubuntu-latest
+            full-build: true
```
drawers, line 24: Similar to the example here:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations

.github/workflows/build.yml
```diff
@@ -32,9 +35,14 @@ jobs:
           distribution: 'zulu'
           java-version: 19
 
-      - name: Test
+      - name: Full build
+        if: matrix.full-build
         run: ./gradlew build
 
+      - name: File system check
+        if: "!matrix.full-build"
```
drawers, line 43: I tried a few different things here including:

```yaml
        if: $${{ !matrix.full-build }}
        if: $${{ matrix.full-build != true }}
```

Escaping with quotation marks works but I am not sure why:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif
https://docs.github.com/en/actions/learn-github-actions/expressions

.github/workflows/build.yml
```diff
@@ -32,9 +35,14 @@ jobs:
...
+        run: ./gradlew :kotlinpoet:allTests
```
drawers, line 44: Wasn't quite sure which task to run in the case we just need to check file system interactions.


```shell
./gradlew :kotlinpoet:check --dry-run
```

gives me:

```
:kotlinpoet:checkKotlinGradlePluginConfigurationErrors SKIPPED
:kotlinpoet:compileKotlinJvm SKIPPED
:kotlinpoet:compileJava SKIPPED
:kotlinpoet:jvmProcessResources SKIPPED
:kotlinpoet:jvmMainClasses SKIPPED
:kotlinpoet:compileTestKotlinJvm SKIPPED
:kotlinpoet:compileTestJava SKIPPED
:kotlinpoet:jvmTestProcessResources SKIPPED
:kotlinpoet:jvmTestClasses SKIPPED
:kotlinpoet:jvmTest SKIPPED
:kotlinpoet:allTests SKIPPED
:kotlinpoet:jvmApiBuild SKIPPED
:kotlinpoet:jvmApiCheck SKIPPED
:kotlinpoet:apiCheck SKIPPED
:spotlessInternalRegisterDependencies SKIPPED
:kotlinpoet:spotlessKotlin SKIPPED
:kotlinpoet:spotlessKotlinCheck SKIPPED
:kotlinpoet:spotlessCheck SKIPPED
:kotlinpoet:testJdk11 SKIPPED
:kotlinpoet:testJdk17 SKIPPED
:kotlinpoet:testJdk19 SKIPPED
:kotlinpoet:testJdk8 SKIPPED
:kotlinpoet:check SKIPPED
BUILD SUCCESSFUL in 980ms
```

Whereas:

```shell
./gradlew :kotlinpoet:allTests
```

gives me:

```
:kotlinpoet:checkKotlinGradlePluginConfigurationErrors SKIPPED
:kotlinpoet:compileKotlinJvm SKIPPED
:kotlinpoet:compileJava SKIPPED
:kotlinpoet:jvmProcessResources SKIPPED
:kotlinpoet:jvmMainClasses SKIPPED
:kotlinpoet:compileTestKotlinJvm SKIPPED
:kotlinpoet:compileTestJava SKIPPED
:kotlinpoet:jvmTestProcessResources SKIPPED
:kotlinpoet:jvmTestClasses SKIPPED
:kotlinpoet:jvmTest SKIPPED
:kotlinpoet:allTests SKIPPED
```

leaving out testJdk8, testJdk11 etc.

I assumed the intention is to reduce GitHub Action minutes so `kotlinpoet:allTests` is okay?

.github/workflows/build.yml
Egorand: Would it be easier to just make checks against `os` instead to avoid introducing a new environment variable? E.g.:

```suggestion
        if: matrix.os != 'ubuntu-latest'
```

And similarly for the `Full build` job.

.github/workflows/build.yml
Egorand: My vote is to run `check`, so that we don't inadvertently exclude any tasks hooked into it, that are important to execute. Curious to hear from Jake though, as it's not completely clear to me which file system interactions were meant.

Egorand (approved):

.github/workflows/build.yml
```diff
@@ -32,9 +32,14 @@ jobs:
           distribution: 'zulu'
           java-version: 19
 
-      - name: Test
+      - name: Full build
+        if: matrix.os == 'ubuntu-latest'
         run: ./gradlew build
 
+      - name: File system check
```
JakeWharton, line 39:
```suggestion
      - name: KotlinPoet check
```

JakeWharton (approved):
This seems fine. I was going to suggest putting the Gradle task into the matrix but this is good enough (and won't need to change ever).

Egorand:
Merged, thank you!

PR #1793, opened 2024-01-02
Fix issue with missing suspend modifier in KSTypeReference.toTypeName

rickclephas:
Since #1742 suspend lambdas are missing their suspend modifier:
```kotlin
param8: suspend () -> String,
```

[1.15.1](https://github.com/square/kotlinpoet/releases/tag/1.15.1):
```kotlin
param8: SuspendFunction0<String>,
```

[1.15.2](https://github.com/square/kotlinpoet/releases/tag/1.15.2):
```kotlin
param8: () -> String,
```


rickclephas:
P.S. as far as I can tell annotations weren't supported in 1.15.1 either, so that doesn't seem like a bug (maybe a feature?).

PR #1803, opened 2024-01-12
Add NameAllocator API to control keyword pre-allocation

Egorand:


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
```diff
@@ -77,7 +77,39 @@ public class NameAllocator private constructor(
...
+   *
+   * ```kotlin
+   * val nameAllocator = NameAllocator(preAllocateKeywords = false)
+   * println(CodeBlock.of("%N", nameAllocator.newName("when"))) // prints "`when`"
+   * ```
+   *
+   * The default behaviour of [NameAllocator] is to pre-allocate keywords - this is the behaviour you'll
+   * get when using the no-arg constructor.
+   */
+  public constructor(preAllocateKeywords: Boolean) : this(
```
Egorand, line 109: @swankjesse I decided to go with this naming after re-reading the docs for NameAllocator, I think it's more consistent with the terminology we use there. Here's our original diff, let me know if you prefer the name we originally came up with:

```diff
diff --git a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
index cda10042..514fb8c4 100644
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
@@ -77,7 +77,20 @@ public class NameAllocator private constructor(
   private val allocatedNames: MutableSet<String>,
   private val tagToName: MutableMap<Any, String>,
 ) {
-  public constructor() : this(mutableSetOf(), mutableMapOf())
+  public constructor() : this(
+    allocatedNames = KEYWORDS.toMutableSet(),
+    tagToName = mutableMapOf(),
+  )
+
+  /**
+   * @param allowKeywords true to return keywords like "data" and "value" from [newName], where they'll potentially
+   *     be emitted with backticks when used to generate source code. The default is false, which will cause programs
+   *     to generate APIs like "data_" and "value_" which do not need to be escaped with backticks.
+   */
+  public constructor(allowKeywords: Boolean) : this(
+    allocatedNames = if (allowKeywords) mutableSetOf<String>() else KEYWORDS.toMutableSet(),
+    tagToName = mutableMapOf(),
+  )
 
   /**
    * Return a new name using `suggestion` that will not be a Java identifier or clash with other
@@ -89,7 +102,7 @@ public class NameAllocator private constructor(
     tag: Any = UUID.randomUUID().toString(),
   ): String {
     var result = toJavaIdentifier(suggestion)
-    while (result.isKeyword || !allocatedNames.add(result)) {
+    while (!allocatedNames.add(result)) {
       result += "_"
     }
```

JakeWharton, line 109: I think "pre-allocate" means "before allocate" and "preallocate" means "allocate in advance" so we want the casing here and elsewhere in the change to be `preallocateKeywords`.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
```diff
@@ -77,7 +77,39 @@ public class NameAllocator private constructor(
   private val allocatedNames: MutableSet<String>,
   private val tagToName: MutableMap<Any, String>,
 ) {
-  public constructor() : this(mutableSetOf(), mutableMapOf())
+  public constructor() : this(preAllocateKeywords = true)
+
+  /**
+   * @param preAllocateKeywords If true, all Kotlin keywords will be pre-allocated and suffixed with
+   * underscores to avoid using them as identifiers:
```
JakeWharton, line 84: Split this into two sentences. Currently it sounds like the preallocated names receive the underscore, but I'm fact it's the requested names which collide with keywords that receive it.

JakeWharton (approved):

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
Egorand: Ah, didn't realize "preallocate" can be a single word, thanks! [2d690ae](https://github.com/square/kotlinpoet/pull/1803/commits/2d690ae28bd662948487c69d54fa1959f5aa9a01)

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
Egorand: Done. [b9d801e](https://github.com/square/kotlinpoet/pull/1803/commits/b9d801e32fffdcd8483310091e2def49d9edb6b1)

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/NameAllocator.kt
JakeWharton: English is severely underspecified! I'm still not 100% sure of this, but most signs point toward "preallocate" being the more correct form. It was a weird rabbit hole to fall into this morning while walking the dog...

Egorand: Lol. English needs official documentation.

swankjesse (approved):

PR #1813, opened 2024-01-24
Omit implicit modifiers on `FileSpec.scriptBuilder`

### LINKED ISSUE #1620: Broken functions in scripts
**Describe the bug**
Functions in generated script (via `FileSpec.scriptBuilder`) get `public` modifier which is prohibited for local functions in Kotlin scripts.

**To Reproduce**
```kotlin
fun main() {
        .apply { returns(UNIT) }
        .build()
    val fileSpec = FileSpec.scriptBuilder("dummy")
        .apply { addFunction(funSpec) }
        .build()

    println(fileSpec.toString())
    
    // public fun foo() {
    //}

    //ERROR: Modifier 'public' is not applicable to 'local function'
}
```

**Expected behavior**
- don't add implicit `public` modifier to functions in script body
OR
- allow to switch off `explicit API` compatibility mode

**Additional context**
KotlinPoet 1.14.2

psymoon:
Closes #1620

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
@@ -238,13 +238,14 @@ internal class CodeWriter constructor(
     codeBlock: CodeBlock,
     isConstantContext: Boolean = false,
     ensureTrailingNewline: Boolean = false,
+    omitImplicitModifiers: Boolean = false,
   ) = apply {
     var a = 0
     var deferredTypeName: ClassName? = null // used by "import static" logic
     val partIterator = codeBlock.formatParts.listIterator()
     while (partIterator.hasNext()) {
       when (val part = partIterator.next()) {
-        "%L" -> emitLiteral(codeBlock.args[a++], isConstantContext)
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -393,15 +394,15 @@ internal class CodeWriter constructor(
...
+  private fun emitLiteral(o: Any?, isConstantContext: Boolean, omitImplicitModifiers: Boolean) {
     when (o) {
       is TypeSpec -> o.emit(this, null)
       is AnnotationSpec -> o.emit(this, inline = true, asParameter = isConstantContext)
       is PropertySpec -> o.emit(this, emptySet())
       is FunSpec -> o.emit(
         codeWriter = this,
         enclosingName = null,
-        implicitModifiers = setOf(KModifier.PUBLIC),
+        implicitModifiers = setOfNotNull(KModifier.PUBLIC.takeUnless { omitImplicitModifiers }),
```
JakeWharton, line 405: Would prefer a simple if/else here
```suggestion
        implicitModifiers = if (omitImplicitModifiers) emptySet() else setOf(KModifier.PUBLIC),
```

JakeWharton (approved):

Egorand (approved):
Dope, thank you!

PR #1827, opened 2024-02-08
Fix trailing newline in PropertySpec

### LINKED ISSUE #941: Consecutive blank lines with top level property

psymoon:
This happens because things like `.endControlFlow` adds `\n`. This removes any trailing newline before emitting the `initializer`.

Closes #941

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
@@ -138,7 +138,9 @@ internal fun stringLiteralWithQuotes(
   }
 }
 
-internal fun CodeBlock.ensureEndsWithNewLine() = if (isEmpty()) {
+internal fun CodeBlock.ensureEndsWithNewLine() = trimTrailingNewLine('\n')
+
+internal fun CodeBlock.trimTrailingNewLine(replaceWith: Char? = null) = if (isEmpty()) {
   this
 } else {
   with(toBuilder()) {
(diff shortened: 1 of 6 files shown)


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
```diff
@@ -138,19 +138,28 @@ internal fun stringLiteralWithQuotes(
   }
 }
 
-internal fun CodeBlock.ensureEndsWithNewLine() = if (isEmpty()) {
+internal fun CodeBlock.ensureEndsWithNewLine() = trimTrailingNewLine('\n')
+
+internal fun CodeBlock.trimTrailingNewLine(replaceWith: Char? = null) = if (isEmpty()) {
```
Egorand, line 143: I really like this!

Egorand (approved):
Very nice!

PR #1829, opened 2024-02-13
fix: make specs, builders and spec args serializable

monosoul:
This change makes spec classes (and their builders, and typical arg classes) serializable.

Use case:
I use `kotlinpoet` a lot with Gradle, have some code generation tasks there. Sometimes I want to use things like `ClassName` or `TypeSpec` as inputs to a Gradle task, but the problem is that Gradle task inputs have to be serializable. So every time I need that, I have to come up with workarounds: either using my own data classes leading to code duplication, or using serializable functions. While both options kinda work, I'd really love to ditch them and simply use `kotlinpoet` classes.

### DIFF
--- a/kotlinpoet/api/kotlinpoet.api
+++ b/kotlinpoet/api/kotlinpoet.api
@@ -11,7 +11,7 @@ public abstract interface class com/squareup/kotlinpoet/Annotatable$Builder {
 	public abstract fun getAnnotations ()Ljava/util/List;
 }
 
-public final class com/squareup/kotlinpoet/AnnotationSpec : com/squareup/kotlinpoet/Taggable {
+public final class com/squareup/kotlinpoet/AnnotationSpec : com/squareup/kotlinpoet/Taggable, java/io/Serializable {
 	public static final field Companion Lcom/squareup/kotlinpoet/AnnotationSpec$Companion;
 	public static final fun builder (Lcom/squareup/kotlinpoet/ClassName;)Lcom/squareup/kotlinpoet/AnnotationSpec$Builder;
 	public static final fun builder (Lcom/squareup/kotlinpoet/ParameterizedTypeName;)Lcom/squareup/kotlinpoet/AnnotationSpec$Builder;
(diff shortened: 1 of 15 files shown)


Egorand:
The use case makes complete sense, that said, there are a couple of downsides I see:

- We'll need to introduce and manage `serialVersionUID`s for the `Serializable` types, and we'll likely need to write tests that would guard from changes that would break serialization. So there's a non-negligible amount of maintenance overhead.
- This change will make it harder for us to make KotlinPoet platform-agnostic if we decide to do so, as it introduces yet another dependency on the JDK.

monosoul:
@Egorand thanks for the prompt reply here! Those are understandable concerns.

> This change will make it harder for us to make KotlinPoet platform-agnostic if we decide to do so, as it introduces yet another dependency on the JDK.

I can add a custom interface here (say `KotlinPoetSerializable`) and for the time being just make it extend `Serializable`, and when you decide to make `KotlinPoet` platform agnostic, you can just add platform-specific variants of that interface (i.e. use `expected`/`actual` declaration of Kotlin multiplatform), where the JVM-one would extend `Serializable`, and others might do nothing.


> We'll need to introduce and manage serialVersionUIDs for the Serializable types, and we'll likely need to write tests that would guard from changes that would break serialization. So there's a non-negligible amount of maintenance overhead.

Not sure if I can do much about that :thinking:

Egorand:
I'm hesitant on this, wanna see if @JakeWharton @psymoon @swankjesse and @ZacSweers have thoughts!

monosoul:
My solution so far was to use something like this:
```kotlin
fun interface CodegenPartProvider<T> : () -> T, Serializable {
    override fun invoke(): T
}
```

And then in the code I can do stuff like:
```kotlin
CodegenPartProvider {
   ClassName("", "")
}
```
That only works if the arguments captured by the lambda are serializable, of course, but at least I don't have to write duplicate classes for everything. Just makes the code a bit less readable.

Okay, thanks for having a look into that! Just wanted to check if that's something doable, since I'm sure I'm not the only one facing this issue :slightly_smiling_face:

PR #1830, opened 2024-02-15
Update Kotlin Metadata to version 0.9.0

sgjesse:
Version 0.9.0 deprecates several APIs including `read` where the replacement `readStrict` was not introduced until version 0.9.0 [https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/Migration.md](https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2FJetBrains%2Fkotlin%2Fblob%2Fmaster%2Flibraries%2Fkotlinx-metadata%2Fjvm%2FMigration.md)

- [ x] `docs/changelog.md`

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,9 @@ Change Log
 * Fix: Fix extension function imports (#1814).
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
+Change: kotlinx-metadata 0.9.0. This is a breaking change for users of the
+:kotlinpoet-metadata module, as `read` is deprecated in 0.9.0 and the
+replacement `readStrict` was introduced in 0.9.0 (#1830).
 
 ## Version 1.16.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,9 +6,7 @@ Change Log
 * Fix: Fix extension function imports (#1814).
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
-Change: kotlinx-metadata 0.9.0. This is a breaking change for users of the
-:kotlinpoet-metadata module, as `read` is deprecated in 0.9.0 and the
-replacement `readStrict` was introduced in 0.9.0 (#1830).
+Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is deprecated in 0.9.0 and replaced with `readStrict` (#1830).
 
 ## Version 1.16.0
 
--- a/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
+++ b/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
@@ -648,34 +648,32 @@ private fun KmProperty.toPropertySpec(
   val returnTypeName = returnType.toTypeName(typeParamResolver)
   val mutableAnnotations = mutableListOf<AnnotationSpec>()
   if (containerData != null && propertyData != null) {
-    if (hasGetter) {
-      getterSignature?.let { getterSignature ->
-        if (!containerData.isInterface &&
-          modality != Modality.OPEN && modality != Modality.ABSTRACT
+    getterSignature?.let { getterSignature ->
+      if (!containerData.isInterface &&
+        modality != Modality.OPEN && modality != Modality.ABSTRACT
+      ) {
+        // Infer if JvmName was used
+        // We skip interface types or open/abstract properties because they can't have @JvmName.
+        // For annotation properties, kotlinc puts JvmName annotations by default in
+        // bytecode but they're implicit in source, so we expect the simple name for
+        // annotation types.
+        val expectedMetadataName = if (containerData is ClassData &&
+          containerData.declarationContainer.isAnnotation
         ) {
-          // Infer if JvmName was used
-          // We skip interface types or open/abstract properties because they can't have @JvmName.
-          // For annotation properties, kotlinc puts JvmName annotations by default in
-          // bytecode but they're implicit in source, so we expect the simple name for
-          // annotation types.
-          val expectedMetadataName = if (containerData is ClassData &&
-            containerData.declarationContainer.isAnnotation
-          ) {
-            name
-          } else {
-            "get${name.safeCapitalize(Locale.US)}"
-          }
-          getterSignature.jvmNameAnnotation(
-            metadataName = expectedMetadataName,
-            useSiteTarget = UseSiteTarget.GET,
-          )?.let { jvmNameAnnotation ->
-            mutableAnnotations += jvmNameAnnotation
-          }
+          name
+        } else {
+          "get${name.safeCapitalize(Locale.US)}"
+        }
+        getterSignature.jvmNameAnnotation(
+          metadataName = expectedMetadataName,
+          useSiteTarget = UseSiteTarget.GET,
+        )?.let { jvmNameAnnotation ->
+          mutableAnnotations += jvmNameAnnotation
         }
       }
     }
-    if (hasSetter) {
-      setterSignature?.let { setterSignature ->
+    if (setter != null) {
+        setterSignature?.let { setterSignature ->
         if (containerData is ClassData &&
           !containerData.declarationContainer.isAnnotation &&
           !containerData.declarationContainer.isInterface &&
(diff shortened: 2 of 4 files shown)


interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -648,34 +648,32 @@ private fun KmProperty.toPropertySpec(
   val returnTypeName = returnType.toTypeName(typeParamResolver)
   val mutableAnnotations = mutableListOf<AnnotationSpec>()
   if (containerData != null && propertyData != null) {
-    if (hasGetter) {
```
Egorand, line 651: Don't we need this check anymore?

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -648,34 +648,32 @@ private fun KmProperty.toPropertySpec(
...
+          useSiteTarget = UseSiteTarget.GET,
+        )?.let { jvmNameAnnotation ->
+          mutableAnnotations += jvmNameAnnotation
         }
       }
     }
-    if (hasSetter) {
-      setterSignature?.let { setterSignature ->
+    if (setter != null) {
+        setterSignature?.let { setterSignature ->
```
Egorand, line 676: nit (should be caught by spotless):

```suggestion
      setterSignature?.let { setterSignature ->
```

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -777,16 +775,16 @@ private fun KmProperty.toPropertySpec(
       // since the delegate handles it
       // vals with initialized constants have a getter in bytecode but not a body in kotlin source
       val modifierSet = modifiers.toSet()
-      if (hasGetter && !isDelegated && modality != Modality.ABSTRACT) {
```
Egorand, line 780: ...and neither here?

docs/changelog.md
```diff
@@ -6,6 +6,9 @@ Change Log
 * Fix: Fix extension function imports (#1814).
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
+Change: kotlinx-metadata 0.9.0. This is a breaking change for users of the
```
Egorand, line 9: This will only be breaking if the users of `:kotlinpoet-metadata` somehow forced the use of `kotlinx-metadata` 0.8.0 though, right? Otherwise, I don't think it's a breaking change, we're simply using a new API, but that's an implementation detail.

Egorand (approved):

docs/changelog.md
sgjesse: Yes, it is only breaking if users of `:kotlinpoet-metadata` has other use of `kotlinx-metadata` where the `read` method is used (there might be other methods with this problem as well).

I put it in as `changelog.md` has: _Change: kotlinx-metadata 0.7.0. This is a breaking change for users of the `:kotlinpoet-metadata` module, as most `Flags`-API extensions have been removed in favor of the now-available first-party versions._ from when `kotlinx-metadata` was updated to 0.7.0.

Egorand: If I understand it correctly, `read` is deprecated, not removed entirely, right? If this is the case, it's not a breaking change, so I wouldn't mention it in the changelog.

With 0.7.0, it seems that we actually removed some `:kotlinpoet-metadata` API since they were replaced with similar first-party API, which is definitely a breaking change.

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
JakeWharton: Seems like it was already likely redundant before. The nullable signature safe call does the same thing.

sgjesse: Yes, it was redundant. This is the errors from building with 0.9.0 before any changes:
```
> Task :interop:kotlinx-metadata:compileKotlin
e: file:///usr/local/google/home/sgjesse/prj/kotlinpoet/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/KotlinPoetMetadata.kt:88:38 Using 'read(Metadata): KotlinClassMetadata' is an error. read() throws an error if metadata version is too high. Use either readStrict() if you want to retain this behavior, or readLenient() if you want to try to read newer metadata.
e: file:///usr/local/google/home/sgjesse/prj/kotlinpoet/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt:651:9 Using 'hasGetter: Boolean' is an error. Kotlin properties always have getters
e: file:///usr/local/google/home/sgjesse/prj/kotlinpoet/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt:677:9 Using 'hasSetter: Boolean' is an error. Check .setter for nullability instead
e: file:///usr/local/google/home/sgjesse/prj/kotlinpoet/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt:780:11 Using 'hasGetter: Boolean' is an error. Kotlin properties always have getters
e: file:///usr/local/google/home/sgjesse/prj/kotlinpoet/interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt:788:11 Using 'hasSetter: Boolean' is an error. Check .setter for nullability instead
```

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
JakeWharton: It's always true

interop/kotlinx-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
```diff
@@ -648,33 +646,31 @@ private fun KmProperty.toPropertySpec(
...
+          metadataName = expectedMetadataName,
+          useSiteTarget = UseSiteTarget.GET,
+        )?.let { jvmNameAnnotation ->
+          mutableAnnotations += jvmNameAnnotation
         }
       }
     }
-    if (hasSetter) {
+    if (setter != null) {
       setterSignature?.let { setterSignature ->
```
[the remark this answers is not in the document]
sgjesse: I think you are right, at least `kotlinp` seems to do just that: https://github.com/JetBrains/kotlin/blob/9fb920874c2b4d1c07d95d921894933670d82069/libraries/tools/kotlinp/jvm/src/org/jetbrains/kotlin/kotlinp/jvm/JvmKotlinp.kt#L116.

Implementation of `setterSignature`
https://github.com/JetBrains/kotlin/blob/9fb920874c2b4d1c07d95d921894933670d82069/libraries/kotlinx-metadata/jvm/src/kotlinx/metadata/jvm/JvmExtensions.kt#L140 and https://github.com/JetBrains/kotlin/blob/9fb920874c2b4d1c07d95d921894933670d82069/libraries/kotlinx-metadata/jvm/src/kotlinx/metadata/jvm/internal/JvmMetadataExtensions.kt#L67.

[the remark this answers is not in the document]
sgjesse: Removed the redundant `null` check noticed by @JakeWharton.

docs/changelog.md
```diff
@@ -6,6 +6,9 @@ Change Log
...
+:kotlinpoet-metadata module, as `read` is deprecated in 0.9.0 and the
+replacement `readStrict` was introduced in 0.9.0 (#1830).
```
Egorand, line 11:
```suggestion
Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is deprecated in 0.9.0 and replaced with `readStrict` (#1830).
```

Egorand (approved):

Egorand:
Should be ready to merge, but please sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) first. Thank you!

JakeWharton:
@Egorand Søren works at Google and is covered under their CCLA with us.

Egorand:
Awesome, merging then!

PR #1833, opened 2024-02-16
Ksp opt in workaround

### LINKED ISSUE #1831: Code generated with ksp interop's toAnnotationSpec() generates code that fails to compile when passed an OptIn annotation
**Describe the bug**
This might be a kotlin compiler bug, but I still think it should be worked around because otherwise you'd have to create your own version of `toAnnotationSpec()` just for `OptIn` annotations.

**To Reproduce**
Say you want to preserve `OptIn` annotations so that generated code copying source types still compiles. You have the source:

```kotlin
@OptIn(MyOptIn::class) @MyProcessor
class MyClass
```
and you run it through your processor, calling `annotation.toAnnotationSpec()` to preserve the `OptIn` annotation. You get

```kotlin
@OptIn(markerClass = arrayOf(MyOptIn::class))
class MyClass
```
which then fails to compile with

```
e: This class can only be used as an annotation or as an argument to @OptIn
```

**Expected behavior**
The generated code compiles.

**Additional context**
It is possible to detect an arg as vararg so it could be passed like you normally would instead of using the explicit array notation. That way you don't have any ugly checks looking specifically for `OptIn`

note: I've only tested on kotlin 1.9.21 and 1.9.22, I don't know if this issue is present on older versions but I would assume so

evant:

~Stacked on #1832.~ Closes #1831.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,7 +7,7 @@ Change Log
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
 Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is deprecated in 0.9.0 and replaced with `readStrict` (#1830).
-* Fix: `KSAnnotation.toAnnotationSpec` writes varargs in place instead of making them an array to work around a kotlin
+* Fix: `KSAnnotation.toAnnotationSpec` writes varargs in place instead of making them an array to work around a Kotlin
   issue with `OptIn` annotations (#1831).
 
 ## Version 1.16.0
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,8 @@ Change Log
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
 Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is deprecated in 0.9.0 and replaced with `readStrict` (#1830).
+* Fix: `KSAnnotation.toAnnotationSpec` writes varargs in place instead of making them an array to work around a kotlin
+  issue with `OptIn` annotations (#1831).
 
 ## Version 1.16.0
 
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
@@ -39,19 +39,35 @@ public fun KSAnnotation.toAnnotationSpec(omitDefaultValues: Boolean = false): An
     is ParameterizedTypeName -> AnnotationSpec.builder(type)
     else -> error("This is never possible.")
   }
+  val params = (annotationType.resolve().declaration as KSClassDeclaration).primaryConstructor?.parameters.orEmpty()
+    .associateBy { it.name }
   useSiteTarget?.let { builder.useSiteTarget(it.kpAnalog) }
   // TODO support type params once they're exposed https://github.com/google/ksp/issues/753
+  var varargValues: List<*>? = null
   for (argument in arguments) {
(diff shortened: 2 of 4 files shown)


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
```diff
@@ -39,19 +39,35 @@ public fun KSAnnotation.toAnnotationSpec(omitDefaultValues: Boolean = false): An
...
+    if (type?.isVararg == true) {
+      // wait to add varargs to end
+      varargValues = value as List<*>
+    } else {
+      val member = CodeBlock.builder()
+      member.add("%N = ", name)
+      addValueToBlock(value, member, omitDefaultValues)
+      builder.addMember(member.build())
+    }
+    if (varargValues != null) {
```
JakeWharton, line 64: Shouldn't this be outside the `for`?

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
```diff
@@ -39,19 +39,35 @@ public fun KSAnnotation.toAnnotationSpec(omitDefaultValues: Boolean = false): An
...
     if (omitDefaultValues) {
       val defaultValue = this.defaultArguments.firstOrNull { it.name?.asString() == name }?.value
       if (isDefaultValue(value, defaultValue)) { continue }
     }
-    val member = CodeBlock.builder()
-    member.add("%N = ", name)
-    addValueToBlock(value, member, omitDefaultValues)
-    builder.addMember(member.build())
+    if (type?.isVararg == true) {
+      // wait to add varargs to end
```
JakeWharton, line 56: super nit: [comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html)
```suggestion
      // Wait to add varargs to end.
```

interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/exampleAnnotations.kt
```diff
@@ -79,3 +79,5 @@ annotation class AnotherAnnotation(val input: String)
 enum class AnnotationEnumValue {
   ONE, TWO, THREE
 }
+
+annotation class AnnotationWithVararg(vararg val args: String)
```
JakeWharton, line 83: Can we add one with mixed vararg/regular arg?

JakeWharton (approved):

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
evant: yep!

interop/ksp/test-processor/src/main/kotlin/com/squareup/kotlinpoet/ksp/test/processor/exampleAnnotations.kt
evant: Added to the existing annotation since, though I could split it out into 2 tests if you feel it's important

JakeWharton:
Can you add a change log entry for the fix? My 2024 resolution was to get better at adding change log entries along with PRs and I've been doing terribly so far.

docs/changelog.md
```diff
@@ -7,6 +7,8 @@ Change Log
 * Fix: Omit implicit modifiers on FileSpec.scriptBuilder (#1813).
 * Fix: Fix trailing newline in PropertySpec (#1827).
 Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is deprecated in 0.9.0 and replaced with `readStrict` (#1830).
+* Fix: `KSAnnotation.toAnnotationSpec` writes varargs in place instead of making them an array to work around a kotlin
```
JakeWharton, line 10:
```suggestion
* Fix: `KSAnnotation.toAnnotationSpec` writes varargs in place instead of making them an array to work around a Kotlin
```

Issue #1753, opened 2023-12-02
CodeBlock composition: statement within statement is not supported


fejesjoco:
**Describe the bug**
I frequently have to deal with generating big function bodies, where I resort to composing multiple smaller CodeBlock's together. In these cases it can happen that a statement is nested in another one, e.g. via lambdas, anonymous classes, etc. By default this throws an error but it's easy to workaround and the workaround could be done in KotlinPoet itself.

**To Reproduce**
```
  val cb0 = buildCodeBlock {
    beginControlFlow("{ foobar ->")
    addStatement("println(foobar)")
    endControlFlow()
  }
  val cb1 = buildCodeBlock {
    addStatement("callWithLambda %L", cb0)
  }
  FileSpec.builder("com.test", "HelloWorld")
    .addType(
      TypeSpec.classBuilder("HelloWorld")
        .addFunction(FunSpec.builder("hello1").addCode(cb1).build())
        .build()
    )
    .build()
    .writeTo(System.out)
```

**Expected behavior**
If done correctly, this should print something like:

```
package com.test

public class HelloWorld {
  public fun hello1() {
    callWithLambda { foobar ->
      println(foobar)
    }
  }
}
```

**Additional context**
This code currently throws `Can't open a new statement until the current statement is closed`. If I use `cb0.toString()` instead of `cb0`, it works reasonably well and even follows the current indentation level. Supporting nesting with stack data structures could work too but I think that would be overkill compared to the simplicity of the string solution.

Egorand:
It would make a lot of sense to squash:

```kotlin
val cb0 = buildCodeBlock {
  beginControlFlow("{ foobar ->")
  addStatement("println(foobar)")
  endControlFlow()
}
val cb1 = buildCodeBlock {
  addStatement("callWithLambda %L", cb0)
}
```

into:

```kotlin
val cb1 = buildCodeBlock {
  beginControlFlow("callWithLambda { foobar ->")
  addStatement("println(foobar)")
  endControlFlow()
}
```

Is it not done in your real project because of reusability? 

The lack of support for nested statements is not really a bug, it's a limitation of the library.

JakeWharton:
And it's been around since its JavaPoet roots.

Egorand:
Don't think we'll consider adding support for nested statements, as it'll likely be complex to implement without any clear benefits. I'm gonna close this for now.

Egorand:
[...]

Issue #1346, opened 2022-08-10
Long lists line wrapping makes code not compile


pablisco:
If we have a list of strings that is too long the way Kotlin Poet wraps the line. However, it looks for spaces without taking into account if the space is in the code or inside a literal String. 

This means we can end with code that does not compile.

This is an example producer code:

```kotlin
        FileSpec.builder("example", "Example")
            .addProperty(PropertySpec.builder("test", List::class.parameterizedBy(String::class))
                .initializer((1..10).joinToString(", ", "listOf(", ")") { "\"$it $it $it\"" })
                .build())
            .build()
```

This generates the following code:

```kotlin
public val test: List<String> = listOf("1 1 1", "2 2 2", "3 3 3", "4 4 4", "5 5 5", "6 6 6", "7 7
    7", "8 8 8", "9 9 9", "10 10 10")
```

Which doesn't compile:

![image](https://user-images.githubusercontent.com/260697/183885877-de6235b1-dd28-40e2-a971-c562bc4b00d3.png)

There are two possible solutions I can think of (there may be more and both could be valid)

1. Allow defining the max line length (and potentially other format options) when creating a `FileSpec`. This could potentially scale a bit better as would help to fix this for other unforeseeable cases. At least until a fix is made. Also, there are some cases where it makes sense to have long lines.
2. Take into account Strings when breaking the line.

For now, the workaround I had to use is to add new line breaks to each entry. However, this is not ideal as some of the collections I'm working with can have hundreds of lists with hundreds of items each and want to avoid wrapping 🤞 

 Thank you for the help and thank you for this amazing tool 🙂

Egorand:
Use `%S` ([docs](https://square.github.io/kotlinpoet/#s-for-strings)) to tell KotlinPoet that those strings are in fact string literals and should be handled accordingly:

```kotlin
FileSpec.builder("example", "Example")
  .addProperty(PropertySpec.builder("test", List::class.parameterizedBy(String::class))
    .initializer((1..10).joinToString(", ", "listOf(", ")") { CodeBlock.of("%S %S %S", it, it, it) })
    .build())
  .build()
```

JakeWharton:
And consider using `map` and then `joinToCode` to keep a `CodeBlock` all the way through. This will allow using things such as member references for functions (like `listOf`) if you need them.

pablisco:
Oh, amazing!!! Thanks for the tip @JakeWharton. For reference, this is what I ended with:

```kotlin
[...]

Issue #1314, opened 2022-07-06
Consider reversing the "wrap by default" behaviour


Egorand:
Treating all spaces as wrapping by default leads to shorter lines and better formatting overall, but in certain cases can break compilation (there's a large number of issues mentioning this problem). Using `·` addresses the problem, but it's not ergonomic:

- `·` can't be typed on the keyboard, so it has to be copy-pasted from the docs.
- The discoverability of this feature is low, people often learn about it only after they've generated uncompilable code. It's especially frustrating for developers who build libraries on top of KotlinPoet.
- Having to use `·` feels like extra effort just to get the library to function properly, which breaks the rule of applying the least amount of effort to get correct results.

We used to have `%W` as a "wrapping space" format modifier prior to 1.0, which could be used optionally to improve the formatting of generated code. I suggest that we revert to something like this in 2.0.

slavonnet:
i si,ple replace all spaces with  `· ` for generated classes.  More elegant resolve is wil,l do Code Formating BEFORE write all class to file in Idea format way.  In kotlin poer soft wrapper is stupid and all time brake compilation in long generated code

pandaninjas:
One way to accomplish the same result could be to add an option to KotlinPoet that would enable or disable this behaviour.

JakeWharton:
I'll copy my comment from another issue I just closed as a duplicate of this one:

> I'm not a huge fan of adding arbitrary options which create a ton of knobs, but in this case it would be an easy way to start getting this functionality in the hands of users (myself included).

felixwiemuth:
I agree that wrapping by default should either be completely disabled, or this should be an option for the user (e.g. as an argument to `FileSpec.builder(...)`). My arguments are similar to the above, see details here: #1887. Mainly that it is easy to make mistakes and that therefore no wrapping should be the default, having to explicitly opt-in for the risky way.

mattshoe:
I would like to second the comment from @felixwiemuth 
Having potentially dangerous options as the default for no other reason than because it looks nice seems counterintuitive to me.

I would very much like to at least have the option to disable wrapping.

PR #1860, opened 2024-03-07
added a new format specifier (%U) for literals that should never wrap

seanreid-toast:
Using KotlinPoet, we found that the existing format specifiers weren't _quite_ appropriate for our use case.

In some cases, we have to generate classes that have large byte arrays. Because of the way those are compiled later, they can't be simple bytes arrays (they get encoded as instructions rather than end up in the constant pool), and we often hit the large method limit. Instead, we encode them as strings. We take the responsibility of escaping characters, and we don't want any line wrapping to make decoding easier.

We tried to use `%S`, but the automatic escaping and using multiline strings made guaranteeing the correctness of later decoding difficult.

We tried to use `%L`, but the automatic line wrapping caused compilation failures and also introduces decoding risk for us.

We didn't want to change behavior of existing specifiers, so adding a new one for a literal without wrapping seemed like the right course. Locally testing, it appears to do exactly what we need and cause no regressions in other functionality. The new specifier is `%U` for "unbroken" literals.

- [~] `docs/changelog.md` has been updated if applicable.
- [X] [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) signed.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,7 @@ Change: kotlinx-metadata 0.9.0. Note that the `KotlinClassMetadata .read` is dep
   issue with `OptIn` annotations (#1831).
 * Fix: `MemberName`s without a package are now correctly imported (#1841)
 * Fix: Throw if primary constructor delegates to other constructors (#1859).
+* New: `%U` format specifier for literals that should never have line breaks added (#????)
 
 ## Version 1.16.0
 
--- a/docs/l-for-literals.md
+++ b/docs/l-for-literals.md
@@ -20,6 +20,7 @@ private fun computeRange(name: String, from: Int, to: Int, op: String): FunSpec 
 ```
 
 Literals are emitted directly to the output code with no escaping. Arguments for literals may be
-strings, primitives, and a few KotlinPoet types described below.
+strings, primitives, and a few KotlinPoet types described below. Literals will have line breaks
+added if the literal is over 100 characters and has whitespace.
 
  [formatter]: https://developer.android.com/reference/java/util/Formatter.html
--- a/docs/u-for-unbroken-literals.md
+++ b/docs/u-for-unbroken-literals.md
@@ -1,0 +1,46 @@
+%U for Unbroken Literals
+===============
+
+Although the string specifier (`%S`) and literal specifier (`%L`) work in the vast majority of
+cases, sometimes escaping and line breaks are unacceptable in the output. In these cases, `%U` is
+the right specifier. `%U` has identical behavior to `%L` with the exception that it will never
+insert line breaks.
+
+This format specifier is useful where unexpected line breaks may break the correctness of the
+output. A specific example is the case where a byte array serialized as a pre-escaped String. Using
+any of the other specifiers in cases like this may unintentionally modify the content and make the
+emitted code incorrect.
+
+With `%U`:
+
+```kotlin
+private fun findLongInString(): FunSpec {
+  val longPreQuotedString = "\"very long string very long string very long string very long string very long string very long string very long string\""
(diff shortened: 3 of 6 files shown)


Egorand:
You can use the non-wrapping space character, `·`, if you need a literal that doesn't wrap:

```kotlin
@Test fun nonWrappingLiterals() {
  val wrapMe = FunSpec.builder("wrapMe")
    .returns(STRING)
    .addStatement(
      "%L",
      "very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string"
        .replace(' ', '·'),
    )
    .build()
  assertThat(toString(wrapMe)).isEqualTo(
    """
      |package com.squareup.tacos
      |
      |import kotlin.String
      |
      |public fun wrapMe(): String {
      |  very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string very long string
      |}
      |
    """.trimMargin(),
  )
}
```

Introducing a specialized format specifier feels redundant to me.

seanreid-toast:
> Introducing a specialized format specifier feels redundant to me.

Thanks for the feedback @Egorand, I definitely understand the point about this change being potentially redundant.

If you don't want to merge this PR, that's okay with me. However, I do think there are a few reasons why something like this, despite its redundancy, might be valuable to the project and should be considered.

1. Discovery of the line wrapping middle-dot concept is difficult. It's documented in `docs/functions.md` and not mentioned anywhere in `docs/l-for-literals.md`, where it might be more discoverable. I don't think I'm alone in missing it given the number of issues [reported](https://github.com/square/kotlinpoet/issues/1314) [about](https://github.com/square/kotlinpoet/issues/1753) [line](https://github.com/square/kotlinpoet/issues/1020) [wrapping](https://github.com/square/kotlinpoet/issues/1346) [behavior](https://github.com/square/kotlinpoet/issues/1252) (non-exhaustive links).

2. The concept of `·` does generally seem backwards to me in the case of `%L`. `%L` is for literals, but to prevent kotlinpoet from modifying the string that supposed to be literal, a developer must themselves modify the string so that kotlinpoet can reverse that modification it to emit the string literally. I think either reversing the behavior in a future major version (as proposed in [#1314](https://github.com/square/kotlinpoet/issues/1314)) or using a solution similar to one in this PR would be appropriate to better as it would eliminate the transient modify-to-unmodify workflow.

All that being said, if this PR isn't a direction you'd like to go, I'd understand. If you're open to me taking a pass at the docs to try to make discovery of `·` simpler instead, I'd be willing to propose something.

Egorand:
I think your use case is rather unusual, where you essentially have a string, but you can't use it with `%S`, so you're resorting to `%L`, and need to perform extra processing to tell the library that what you're passing in is actually a string - so I definitely understand that this feels backwards. From the 5 issues you linked, only #1753 is somewhat similar, but the use-case there is different; other issues are either bugs that were resolved, or failure to use `%S` for string literals. I do think that #1314 is the long term solution, though it likely requires a major version bump due to its impact.

In the general spirit of making common things easy and making less common things possible, the library does provide you with tools to work around the limitation you have, but I don't think it warrants dedicated API to support an uncommon use case. Regarding extra documentation, we would surely welcome PRs. I must mention though that the main use case for `·` is to control wrapping in constructs that KotlinPoet does not model as strictly, such as function bodies; in literals, `%S` prevents auto-wrapping, and `%L` expects input that either can't be wrapped or is safe to wrap.

If you'd like an extra set of eyes on the issues you're having with `%S`, I'm more than happy to take a look - please feel free to open a discussion and share relevant code snippets.

PR #1866, opened 2024-03-15
Wrap supertypes if primary constructor wraps

### LINKED ISSUE #1865: Wrap supertype/superinterfaces like parameters
Once we reach 3, wrap to individual lines

Today:
```kotlin
public class TestSchemaWidgetSystem<W : Any>(
  override val TestSchema: TestSchemaWidgetFactory<W>,
  override val RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
  override val RedwoodLazyLayout: RedwoodLazyLayoutWidgetFactory<W>,
) : WidgetSystem<W>, TestSchemaWidgetFactoryOwner<W>, RedwoodLayoutWidgetFactoryOwner<W>,
    RedwoodLazyLayoutWidgetFactoryOwner<W> {
```

What I want:
```kotlin
public class TestSchemaWidgetSystem<W : Any>(
  override val TestSchema: TestSchemaWidgetFactory<W>,
  override val RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
  override val RedwoodLazyLayout: RedwoodLazyLayoutWidgetFactory<W>,
) : WidgetSystem<W>,
    TestSchemaWidgetFactoryOwner<W>,
    RedwoodLayoutWidgetFactoryOwner<W>,
    RedwoodLazyLayoutWidgetFactoryOwner<W> {
```

JakeWharton:
This means there's always a ") : " on a line which prefixes the supertypes, so force them to also wrap to multiple lines.

I decided to take a more opinionated approach that wasn't based on count, but instead is based on whether the primary constructor wraps to multiple lines. Closes #1865.

### DIFF
--- a/kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
+++ b/kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
@@ -596,6 +596,47 @@ class TypeSpecTest {
     )
   }
 
+  @Test fun enumWithPrimaryConstructorAndMultipleInterfaces() {
+    val roshambo = TypeSpec.enumBuilder("Roshambo")
+      .addSuperinterface(Runnable::class)
+      .addSuperinterface(Cloneable::class)
+      .addEnumConstant(
+        "SCISSORS",
+        TypeSpec.anonymousClassBuilder()
+          .addSuperclassConstructorParameter("%S", "peace sign")
+          .build(),
+      )
+      .addProperty(
+        PropertySpec.builder("handPosition", String::class, KModifier.PRIVATE)
+          .initializer("handPosition")
+          .build(),
+      )
+      .primaryConstructor(
+        FunSpec.constructorBuilder()
+          .addParameter("handPosition", String::class)
+          .build(),
+      )
+      .build()
+    assertThat(toString(roshambo)).isEqualTo(
+      """
+        |package com.squareup.tacos
+        |
+        |import java.lang.Runnable
+        |import kotlin.Cloneable
+        |import kotlin.String
+        |
+        |public enum class Roshambo(
+        |  private val handPosition: String,
+        |) : Runnable,
+        |    Cloneable {
+        |  SCISSORS("peace sign"),
+        |  ;
+        |}
+        |
+      """.trimMargin(),
+    )
+  }
+
   /** https://github.com/square/javapoet/issues/193  */
   @Test fun enumsMayDefineAbstractFunctions() {
     val roshambo = TypeSpec.enumBuilder("Tortilla")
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -596,6 +596,47 @@ class TypeSpecTest {
...
+        |
+        |import java.lang.Runnable
+        |import kotlin.Cloneable
+        |import kotlin.String
+        |
+        |public enum class Roshambo(
+        |  private val handPosition: String,
+        |) : Runnable,
+        |    Cloneable {
+        |  SCISSORS("peace sign"),
```
JakeWharton, line 632: ktlint wants this to be
```suggestion
        |public enum class Roshambo(
        |  private val handPosition: String,
        |) : Runnable,
        |  Cloneable {
        |  SCISSORS("peace sign"),
```
which I absolutely hate. Visually there's no delineation between the supertypes and the body.

We could force a leading newline and do
```suggestion
        |public enum class Roshambo(
        |  private val handPosition: String,
        |) : Runnable,
        |  Cloneable {
        |
        |  SCISSORS("peace sign"),
```
which ktlint allows. It's not great, but I'd be okay with this.

Thoughts? We could also just keep what I've already done. Most people aren't formatting generated code and I prefer mine best.

Egorand: The [code style](https://kotlinlang.org/docs/coding-conventions.html#class-headers) suggests that all supertypes should have the same indentation, so I think it's a bug in ktlint? I also prefer it this way, so let's keep it as is.

kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -1061,6 +1102,63 @@ class TypeSpecTest {
...
+      .build()
+    assertThat(toString(typeSpec)).isEqualTo(
+      """
+        |package com.squareup.tacos
+        |
+        |import java.io.Serializable
+        |import java.util.AbstractSet
+        |import kotlin.Comparable
+        |
+        |public abstract class Taco() : AbstractSet<Food>(), Serializable, Comparable<Taco>
```
Egorand, line 1123: Will this currently simply overflow for longer lists? The suggestion in the style guide is to wrap after the `:`, but I believe it'll be quite hard to recognize this case in the code.

image

Egorand (approved):

kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
JakeWharton: They only mention horizontal alignment when you make a line break after the colon for non-primary constrictor cases.

The example with a wrapping primary constructor only aligns because they are doing a 4-space indent. So unfortunately I think ktlint is correct in what it does.

That being said, we can still leave it.

Egorand: > The example with a wrapping primary constructor only aligns because they are doing a 4-space indent.

Good point.

kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
JakeWharton: I want an API like 
```kotlin
val mark = codeWriter.mark()
val originalLine = codeWriter.lineNumber

// Do normal write 
if (codeWriter.lineNumber > originalLine) {
  codeWriter.reset(mark)

  // Do wrapped write
}
```
Would be useful for so many constricts. Probably even deserves a helper function that just takes two lambdas.

PR #1867, opened 2024-03-15
Add `MemberSpecHolder`

JakeWharton:
Holds properties and functions.

Part of #1553 

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
@@ -48,7 +48,8 @@ public class TypeSpec private constructor(
   ContextReceivable by contextReceivers,
   Annotatable,
   Documentable,
-  TypeSpecHolder {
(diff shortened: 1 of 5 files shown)


kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -766,6 +734,43 @@ public class TypeSpec private constructor(
 
     @Suppress("RedundantOverride")
     override fun addTypes(typeSpecs: Iterable<TypeSpec>): Builder = super.addTypes(typeSpecs)
+
+    @Suppress("RedundantOverride")
```
Egorand, line 738: Do we need to keep these overrides for backwards compatibility? Should we mark them as "delete in 2.0"?

JakeWharton: That's right. We have so many that you can't even see they're already in a comment region which does indicate they're only for compatibility so we should know they can be deleted for 2.0.

PR #1881, opened 2024-04-12
omit more permissive modifiers on restricted visibility types

j-roskopf:
closes https://github.com/square/kotlinpoet/issues/1301


JakeWharton:
I would like to see the tests cover more cases. Basically four classes with each of the four visibilities, each having four functions or properties with the four visibilities. I'm sure many of the cases are covered implicitly in other tests, but having one that actually covers all the cases specific to this functionality would be nice.

j-roskopf:
@JakeWharton Updated tests to better add better coverage for the matrix you mentioned. I added tests to cover 5 classes (public, private, protected, internal, unspecified) with properties + functions over the same 5

kotlinpoet/src/commonTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -5778,6 +5779,385 @@ class TypeSpecTest {
...
+      private class Taco {
+        fun unspecifiedStringMethod(): String = "taco"
+
+        private fun privateStringMethod(): String = "taco"
+
+        internal fun internalStringMethod(): String = "taco"
+
+        protected fun protectedStringMethod(): String = "taco"
+
+        public fun publicStringMethod(): String = "taco"
```
JakeWharton, line 6035: A bunch of these tests seem incorrect with the goal. In my original bug there's a private class showing that the public modifier isn't needed.

j-roskopf: I wanted to verify that if any access modifier was specified, it was respected and added to the property / function even if the contain class was private / internal / protected / unspecified / public. So I added a matrix of those 5 each containing a function across the same 5, with the main change in behavior being that where the class is private / internal / protected, unspecifiedStringMethod has no access modifier specified in the source code because no modifier was specified through kotlin poet. 

please let me know if i'm way off base!

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -161,11 +161,19 @@ internal class CodeWriter constructor(
       emit(KModifier.PUBLIC.keyword)
       emit(" ")
     }
+
+    // public always being an implicit modifier in addition to whatever we inherited.
+    // we don't want to throw away the implicit modifier we inherited
+    val implicitModifierContainsAtLeastTwoAccessMonitors = implicitModifiers.count {
```
Egorand, line 167: What are "access monitors"?

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -652,6 +660,11 @@ internal class CodeWriter constructor(
       return false
     }
 
+    if (implicitModifiers.contains(KModifier.INTERNAL) || implicitModifiers.contains(KModifier.PRIVATE) || implicitModifiers.contains(KModifier.PROTECTED)) {
```
Egorand, line 663:
```suggestion
    if (implicitModifiers.containsAnyOf(KModifier.INTERNAL, KModifier.PRIVATE, KModifier.PROTECTED)) {
```

(it's a handy helper defined in `Util.kt`)

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
j-roskopf: hey! access modifiers are the set of internal, private, public, protected. based that nomenclature off this comment https://github.com/square/kotlinpoet/blob/1242b415d1fb54aa78e318e3c0694b0637aaee69/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/KModifier.kt#L31

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
j-roskopf: makes sense! made the change

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
Egorand: Ah, I see! The code says "access monitors", which looks like a typo!

j-roskopf: even if your comment asking what are 'access monitors', my brain still read that as access modifiers 🙃 whoops! fixed the typo

Egorand: Happens to me all the time! No worries & thanks for fixing.

JordanLongstaff:
Bump. I'd like to see this merged ASAP, please. And there are merge conflicts to resolve first.

j-roskopf:
@JakeWharton please let me know if you thought my original implementation was off base from what you were thinking. happy to address the conflicts and bring the PR to a mergable state

PR #1956, opened 2024-08-04
KSP typealias fixes

### LINKED ISSUE #1945: KSAnnotation.toAnnotationSpec() fails if the KSAnnotation is a typealias to another Annotation
**Describe the bug**

Consider following scenario:
```kotlin
annotation class SomeAnnotation
typealias SomeOtherAnnotation = SomeAnnotation

@SomeOtherAnnotation
fun doSomething() {
}
```
Calling `KSAnnotation.toAnnotationSpec()` on the `doSomething` function annotations, KSP will throw:

```
e: java.lang.ClassCastException: class com.google.devtools.ksp.impl.symbol.kotlin.KSTypeAliasImpl cannot be cast to class com.google.devtools.ksp.symbol.KSClassDeclaration (com.google.devtools.ksp.impl.symbol.kotlin.KSTypeAliasImpl and com.google.devtools.ksp.symbol.KSClassDeclaration are in unnamed module of loader 'app')
	at com.squareup.kotlinpoet.ksp.AnnotationsKt.toAnnotationSpec(Annotations.kt:48)
	at com.squareup.kotlinpoet.ksp.AnnotationsKt.toAnnotationSpec$default(Annotations.kt:35)...
```

**Expected behavior**
- I would expect this to work, without errors.
- `KSAnnotation.toAnnotationSpec()` should return resolved annotation

**Additional context**
https://github.com/square/kotlinpoet/blob/79ec8a42447a94db5b738fa303c8d433ad459bcb/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt#L48

This seems like it should be a safe cast to KSClassDeclaration

ZacSweers:
See changelog!

Resolves #1945

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,6 +4,8 @@ Change Log
 ## Unreleased
 
 * Fix: Enum classes that only have an init block now also generate the required semicolon (#1952)
+* Fix: Preserve typealiases in `KSAnnotation.toAnnotationSpec()`. (#1945)
+* Fix: Preserve nullability in `KSType.toClassName()`. (#1956)
 
 ## Version 1.18.1
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,8 @@ Change Log
 * Fix: Enum classes that only have an init block now also generate the required semicolon (#1952)
 * Fix: Preserve typealiases in `KSAnnotation.toAnnotationSpec()`. (#1945)
 * Fix: Preserve nullability in `KSType.toClassName()`. (#1956)
+* New: Add `KSTypeAlias.toClassName()`. (#1956)
+* New: Add `KSType.toClassNameOrNull()`. (#1956)
 
 ## Version 1.18.1
 
--- a/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
+++ b/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
@@ -108,14 +108,6 @@ private val AnnotationUseSiteTarget.kpAnalog: UseSiteTarget
     AnnotationUseSiteTarget.DELEGATE -> UseSiteTarget.DELEGATE
   }
 
-internal fun KSType.unwrapTypeAlias(): KSType {
-  return if (this.declaration is KSTypeAlias) {
-    (this.declaration as KSTypeAlias).type.resolve()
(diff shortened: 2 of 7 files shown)


interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
```diff
@@ -21,31 +21,28 @@ import com.google.devtools.ksp.symbol.KSAnnotation
...
-        AnnotationSpec.builder(className.parameterizedBy(typeArgs))
-      }
-    }
+  val typeName = annotationType.resolve().toTypeName()
+
+  val builder = if (typeName is ClassName) {
+    AnnotationSpec.builder(typeName)
+  } else {
+    AnnotationSpec.builder(typeName as ParameterizedTypeName)
+  }
```
Egorand, line 42: If `typeName` is neither `ClassName` nor `ParameterizedTypeName` - we crash?

docs/changelog.md
```diff
@@ -4,6 +4,10 @@ Change Log
 ## Unreleased
 
 * Fix: Enum classes that only have an init block now also generate the required semicolon (#1952)
+* Fix: Preserve typealiases in `KSAnnotation.toAnnotationSpec()`. (#1945)
```
Egorand, line 7: nit:

```suggestion
* Fix: Preserve typealiases in `KSAnnotation.toAnnotationSpec()`. (#1956)
```

Egorand (approved):

interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/Annotations.kt
ZacSweers: Yes but it should be impossible for that to happen here

docs/changelog.md
ZacSweers: This one actually linked the issue, so we link issues or fixes?

Egorand: I link fixes when compiling the changelog for a release, since fixes contain actually diffs. Does it make more sense to link issues?

ZacSweers: Up to you! I always just pick whichever one was contributed externally for their visibility as a default, but no strong feelings

ZacSweers: Gonna roll with this

PR #1959, opened 2024-08-08
Make the kotlinpoet module multi-platform and add source set configuration for the JS and wasmJs platforms

ForteScarlet:
#304 

Hi! This PR attempts to change the `kotlinpoet` module into a true multi-platform module: adding source code sets for platforms other than the JVM platform: JS and WasmJs. No native platforms have been added yet.

There should be no change in behaviour on the JVM platform after the change, and the unit tests all pass locally.

There are some limitations and issues with JS and wasmJs.

- The `kotlin-reflect` in JS and WasmJs is very limited, so it's almost impossible to get valid results with `KClass.asXxx`.
- The `IDENTIFIER_REGEX` in the `Util.kt` cannot be initialised, and you get the error: `PatternSyntaxException: No such character class`.  (I'm not very good at regular expressions, I don't understand the problem with them.)
- `CodePoint` related implementations are temporary on JS and WasmJS, not sure if that's correct. 
- `isJavaIdentifierStart` and `isJavaIdentifierPart` are not implemented on JS and WasmJs at the moment and only return default values temporarily.
- `formatNumericValue` is not implemented on JS and WasmJs at the moment and currently returns `toString` itself directly, unformatted.

I have tried not to change the logic and ensure API compatibility, but I have to say it's still a lot of changes. I wonder if such a change would be allowed or accepted? I think this might simply advance #304 ?

 Please review these changes and let me know whether you will be accepted or not, thanks!


Egorand:
This PR is gigantic and due to that it's virtually unreviewable. Please split it up into multiple incremental commits, each of which has a clear commit message that describes what it's doing. Also, please try to avoid noisy changes such as moving function or property annotations from the same line to the previous line.

To be honest I'm also not completely sure there's value in getting the library to work in a non-Java environment, given the limitations. I'll let other contributors voice their opinions, probably worth waiting until then before you start cleaning it up and splitting into commits.

ForteScarlet:
Thank you for your reply! I think running in a non-Java environment can make browser-based code generation or mobile code generation in non-Java environments easier. However, let's wait for the results of your discussion and look forward to the follow-up~

JakeWharton:
I agree that JS is valuable.

I also agree that we need to do this piece by piece.

The TypeName hierarchy seems like a good candidate.

ForteScarlet:
Hey! A while ago, I split and re-pushed the commit message as much as possible (I think I forgot to mention it here, sorry). The current changes should have the vast majority implemented via KMP, and the behavior on the JVM platform should be no different than before.

Some of the content that may be problematic in the platform implementation, ambiguous, or I'm not quite sure how to implement it in the platform, I've marked with a TODO.

Feel free to review them and let me know if there's anything I can do, thanks!

Egorand:
Thanks! Will start the review today.

kotlinpoet/build.gradle.kts
```diff
@@ -23,11 +24,12 @@ spotless {
   kotlin {
     targetExclude(
       // Non-Square licensed files
-      "src/commonMain/kotlin/com/squareup/kotlinpoet/ClassName.kt",
-      "src/commonTest/kotlin/com/squareup/kotlinpoet/AbstractTypesTest.kt",
-      "src/commonTest/kotlin/com/squareup/kotlinpoet/ClassNameTest.kt",
-      "src/commonTest/kotlin/com/squareup/kotlinpoet/TypesEclipseTest.kt",
-      "src/commonTest/kotlin/com/squareup/kotlinpoet/TypesTest.kt",
+      "src/*Main/kotlin/com/squareup/kotlinpoet/ClassName.kt",
+      "src/*Main/kotlin/com/squareup/kotlinpoet/ClassName.*.kt",
```
Egorand, line 28: What extensions does this pattern target?

kotlinpoet/build.gradle.kts
```diff
@@ -37,19 +39,61 @@ kotlin {
...
+        }
+      }
+    }
+    nodejs {
+      testTask {
+        useMocha()
+      }
+    }
+    binaries.library()
+
```
Egorand, line 56: nit: delete empty line.

kotlinpoet/build.gradle.kts
```diff
@@ -37,19 +39,61 @@ kotlin {
...
+      }
+    }
+    binaries.library()
+  }
+
   @OptIn(ExperimentalKotlinGradlePluginApi::class)
   compilerOptions {
     allWarningsAsErrors.set(true)
     optIn.add("com.squareup.kotlinpoet.DelicateKotlinPoetApi")
+    freeCompilerArgs.add("-Xexpect-actual-classes")
```
Egorand, line 80: According to https://youtrack.jetbrains.com/issue/KT-61573, this feature is stable since 1.9.20, so this shouldn't be needed?

kotlinpoet/build.gradle.kts
```diff
@@ -62,6 +106,18 @@ kotlin {
...
+      dependsOn(commonMain.get())
+    }
+
+    jsMain {
+      dependsOn(nonJvmMain)
+    }
+    wasmJsMain {
+      dependsOn(nonJvmMain)
+    }
+
```
Egorand, line 120: nit: remove empty line.

kotlinpoet/gradle.properties
```diff
@@ -2,3 +2,4 @@ POM_ARTIFACT_ID=kotlinpoet
 POM_NAME=KotlinPoet
 POM_DESCRIPTION=Use beautiful Kotlin code to generate beautiful Kotlin code.
 POM_PACKAGING=jar
+kotlin.mpp.applyDefaultHierarchyTemplate=false
```
Egorand, line 5: Any issues with adopting the default hierarchy template vs opting out?

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
Egorand, line 1: Do we need `nonJvm` in the file extension, and if yes, why?

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
```diff
@@ -15,39 +15,36 @@
...
 package com.squareup.kotlinpoet
 
-import com.squareup.kotlinpoet.CodeBlock.Companion.isPlaceholder
-import java.util.Collections
+import kotlin.reflect.KClass
 
 internal object NullAppendable : Appendable {
-  override fun append(charSequence: CharSequence) = this
-  override fun append(charSequence: CharSequence, start: Int, end: Int) = this
+  override fun append(charSequence: CharSequence?) = this
```
Egorand, line 22: Curious about the reason for changing the parameter type here.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
```diff
@@ -0,0 +1,111 @@
...
+internal actual fun <T : Comparable<T>> Sequence<T>.toSortedSet(): Set<T> =
+  toSortedSetKt()
+
+internal actual fun <T : Comparable<T>> List<T>.toSortedSet(): Set<T> =
+  toSortedSetKt()
+
+internal actual fun <T : Comparable<T>> sortedSetOf(): MutableSet<T> =
+  TreeSet()
+
+internal actual inline fun <reified E : Enum<E>> enumSetOf(vararg values: E): MutableSet<E> {
```
Egorand, line 53: Where is this function used? Why does the implementation need a `when`, can we implement it in the same way `toEnumSet` was implemented?

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,113 @@
...
+
+internal actual fun Char.isJavaIdentifierStart(): Boolean {
+  return isLetter() ||
+    this in CharCategory.LETTER_NUMBER ||
+    this == '$' ||
+    this == '_'
+}
+
+/*
+* A character may be part of a Java identifier if any of the following conditions are true:
```
Egorand, line 73: nit: fix indentation.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,113 @@
...
+* - it is a letter
+* - it is a currency symbol (such as '$')
+* - it is a connecting punctuation character (such as '_')
+* - it is a digit
+* - it is a numeric letter (such as a Roman numeral character)
+* - it is a combining mark
+* - it is a non-spacing mark
+* isIdentifierIgnorable returns true for the character
+ */
+
```
Egorand, line 83: nit: remove empty line.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,113 @@
...
+    isDigit() ||
+    this in CharCategory.LETTER_NUMBER ||
+    this in CharCategory.NON_SPACING_MARK ||
+    this == '_' ||
+    this == '$' ||
+    isIdentifierIgnorable()
+}
+
+internal fun Char.isIdentifierIgnorable(): Boolean {
+  /*
```
Egorand, line 97: nit: non-documentation comments don't need the `/** */` syntax, can just use `//`.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmAnnotationMirror.jvm.kt
Egorand, line 1: Do we need `jvm` in file extensions? What does it serve for?

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmModifier.nonJvm.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+public actual enum class JvmModifier {
+  PUBLIC,
```
Egorand, line 19: Do we actually need to re-declare the values in an actual enum class?

Egorand (commented):
(reviewed first 5 commits so far)

kotlinpoet/build.gradle.kts
ForteScarlet: Since the original `ClassName.kt` was migrated to commonMain, jvmMain, jsMain, wasmJsMain, etc., this pattern corresponds to the `ClassName` file under each platform module.
The other `*Test` actually correspond to files that currently exist only in `jvmTest`.

kotlinpoet/build.gradle.kts
ForteScarlet: https://youtrack.jetbrains.com/issue/KT-61573 seems to mean that `expect/actual fun` has stabilized, but `expect/actual classes` is still in Beta.

You can also see in https://kotlinlang.org/docs/components-stability.html#current-stability-of-kotlin-components that the Status of `expected and actual classes` is still Beta.

The `expect/actual` classes will also show warnings if you try to remove this option, both in the IDE and when executing the gradle command for compilation.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
ForteScarlet: `.nonJvm` is not required, like `.jvm` and the like on other platforms.

kotlinpoet/gradle.properties
ForteScarlet: Since nonJvm is not the default hierarchy, we don't use 'default hierarchy'. 
If we don't add this property, a warning will appear at `build`:

```
The Default Kotlin Hierarchy Template was not applied to 'project ':kotlinpoet'':
Explicit .dependsOn() edges were configured for the following source sets:
[jsMain, nonJvmMain, wasmJsMain]

Consider removing dependsOn-calls or disabling the default template by adding
    'kotlin.mpp.applyDefaultHierarchyTemplate=false'
to your gradle.properties

Learn more about hierarchy templates: https://kotl.in/hierarchy-template
```

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
ForteScarlet: Previously, `Appendable` was actually `java.lang.Appendable `in Java, and the type of the parameter in its `append(charSequence: CharSequence)` was the platform type, which could actually be `null`.

In the expected definition of `kotlin.text.Appendable`, these functions are actually `append(value: CharSequence?)`, and when detached from the JVM, `?` becomes mandatory.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
ForteScarlet: `enumSetOf` is mainly used to replace the previous direct use of `EnumSet.of(...)`, compared to `toEnumSet`, in the JVM may be a little more efficient? But it's not very efficient, so `toEnumSet` is fine.
The use of `when` is also only because it avoids a `copyOf` copy, so there may be a small efficiency gain, but it's not much.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmAnnotationMirror.jvm.kt
ForteScarlet: The `.jvm` extension serves no real purpose, and is probably more useful for distinguishing between files of the same name on different platforms.
I've seen this kind of file naming in some multiplatform libraries before (e.g. [Ktor](https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-core/jvm/src/io/ktor/client/engine/HttpClientEngineBase.jvm.kt)), and I've kept it because I think it's good.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmModifier.nonJvm.kt
ForteScarlet: It seems that `expect enum class` needs to be re-declare in `actual` if it's not a `typealias`, otherwise you'll get an error:
```
The following declaration is incompatible because some entries from expected enum are missing in the actual enum:
    expect enum class JvmModifier : Enum<JvmModifier>
```

kotlinpoet/build.gradle.kts
Egorand: Ah, you're right, thanks for the clarification!

kotlinpoet/gradle.properties
Egorand: Makes sense!

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Util.kt
Egorand: Gotcha, thanks!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmAnnotationMirror.jvm.kt
Egorand: Got it, I'm okay with using it. Good to know other libraries use this pattern.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
Egorand: Got it, fine with leaving as is.

kotlinpoet/build.gradle.kts
```diff
@@ -37,19 +39,61 @@ kotlin {
...
+        useKarma {
+          useChromeHeadless()
+        }
+      }
+    }
+    nodejs {
+      testTask {
+        useMocha()
+      }
+    }
```
JakeWharton, line 54: There's no need to run the tests in both the browser and on Node. Just Node is fine.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/ClassName.kt
```diff
@@ -249,28 +246,20 @@ public class ClassName internal constructor(
...
-  return ClassName(names)
-}
+public expect fun JvmClass<*>.asClassName(): ClassName
+
+/** Returns the class name for `element`. */
+@DelicateKotlinPoetApi(
+  message = "Element APIs don't give complete information on Kotlin types. Consider using" +
+    " the kotlinpoet-metadata APIs instead."
+)
+public expect fun JvmTypeElement.asClassName(): ClassName
```
JakeWharton, line 256: Things that are JVM-specific should not be in common. Moreover, kotlinpoet-metadata is JVM-only due to kotlin-metadata being JVM-only.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/Closeable.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+public expect interface Closeable : AutoCloseable
```
JakeWharton, line 18: We should use `AutoCloseable` on all platforms and only add the JVM's `Closeable` where needed. We don't want to put a type in our public API for just doing this, though.

ForteScarlet: I didn't realize that the types that implement `Closeable` are internal types, you're right, it doesn't seem necessary to define an additional `Closeable`

ForteScarlet:
Changes have been made to the needed adjustments mentioned above. Feel free to let me know of any other reviews after that~

egorikftp:
@JakeWharton @Egorand Could you please take a look the PR?
I'm developing plugin for converting xml/svg into Compose ImageVector and I want to make a WASM version, but kotlinpoet not works on wasm target yet 🙂

https://github.com/ComposeGears/Valkyrie/issues/180

JakeWharton:
I don't think this PR is reviewable. We need to start with something smaller in scope like the TypeName hierarchy by itself. There's way too much API surface here to do at once.

ForteScarlet:
@JakeWharton I think I may have misunderstood a bit before. I thought I needed to split the content of this PR into more defined commits based on the changes, but now it seems I need to submit the PRs for each "small part" of the changes step-by-step, right? 😂

Egorand:
Yeah, that was the original suggestion, but it seems it'll be really hard to review it all as a single PR even broken down into individual commits. Agree we should try to merge as many bits as possible independently.

JakeWharton:
I definitely mentioned only doing something small like TypeName first _somewhere_, but I can't find where. It's too hard to judge whether this is correct or not since it's so massive, and it'll take forever for us to review and then iterate on the feedback and such.

JakeWharton:
Since we have no users on other targets, we can expect/actual odd dependencies and use `TODO()` implementations for now.

Egorand:
> I submitted a more ‘smaller’ change #1992 a while ago following the commit steps in this pr. If you have any suggestions or need me to do, please feel free to let me know, thanks! 😊

Yep yep, thanks for that, planning to review this week - apologies for the delay!

PR #1976, opened 2024-09-15
Add FileSpec.Builder.relativePath()

TBSten:
    be added to the changelog.


JakeWharton:
Hmm I'm not sure how much this should be our responsibility since you could always write to a non-standard path yourself using the `toString()` output. Can you talk more about why you need this and why the default structure is not suitable?

TBSten:
> Hmm I'm not sure how much this should be our responsibility since you could always write to a non-standard path yourself using the toString() output

Oh, I didn't know `toString()` existed 🙏.
That looks good enough to me 😅.

Thank you for comment !

PR #1999, opened 2024-10-10
Bump Java versions for build and tests

Egorand:


build.gradle.kts
```diff
@@ -116,7 +116,7 @@ subprojects {
     fun Project.setupCheckTask(testTaskName: String) {
       // Copied from https://github.com/square/retrofit/blob/master/retrofit/build.gradle#L28.
       // Create a test task for each supported JDK. We check every "LTS" + current version.
-      val versionsToTest = listOf(8, 11, 17, 19)
+      val versionsToTest = listOf(8, 11, 17, 21, 22)
```
JakeWharton, line 119:
```suggestion
      val versionsToTest = listOf(8, 11, 17, 21, 23)
```
Latest is 23, or does Gradle not support that yet?

JakeWharton (approved):

build.gradle.kts
Egorand: 8.10 does!

PR #2001, opened 2024-10-11
Introduce the new ♢ modifier and don't wrap spaces by default

### LINKED ISSUE #1314: Consider reversing the "wrap by default" behaviour

### LINKED ISSUE #2000: Introduce WHITE DIAMOND SUIT as the new breaking space character

Egorand:
Resolves #2000. Closes #1314.

### DIFF
--- a/interop/kotlin-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
+++ b/interop/kotlin-metadata/src/main/kotlin/com/squareup/kotlinpoet/metadata/specs/KotlinPoetMetadataSpecs.kt
@@ -251,7 +251,7 @@ public fun KmPackage.toFileSpec(
     .build()
 }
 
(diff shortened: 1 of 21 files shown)


JakeWharton (approved):
Very nice

PR #2003, opened 2024-10-14
Update line wrapping-related documentation

Egorand:


docs/functions.md
```diff
@@ -94,67 +94,48 @@ FunSpec.builder("add")
   .build()
 ```
 
-## Spaces wrap by default!
+## Spaces don't wrap by default
```
JakeWharton, line 97: Might want to consider being a bit more active in describing the current state of the world rather than having it be somewhat in context of the previous behavior.
```suggestion
## Wrapping is explicit
```

docs/functions.md
```diff
@@ -94,67 +94,48 @@ FunSpec.builder("add")
...
-{ string -> println(string) }
+public fun foo() = (100..10000).map { number -> number * number }.map { number -> number.toString() }.also { string -> println(string) }
 ```
 
-Unfortunately this code is broken: the compiler expects `also` and `{` to be on the same line.
-KotlinPoet is unable to understand the context of the expression and fix the formatting for you, but
-there's a trick you can use to declare a non-breaking space - use the `·` symbol where you would
-otherwise use a space. Let's apply this to our example:
+While the output is correct, the resulting line of code is quite long and hard to read. KotlinPoet
+nis unable to understand the context of the expression and adjust the formatting for you, but
```
JakeWharton, line 116:
```suggestion
is unable to understand the context of the expression and adjust the formatting for you, but
```

docs/functions.md
```diff
@@ -94,67 +94,48 @@ FunSpec.builder("add")
...
 ```
 
-## Spaces wrap by default!
+## Spaces don't wrap by default
 
-In order to provide meaningful formatting, KotlinPoet would replace spaces, found in blocks of code,
-with new line symbols, in cases when the line of code exceeds the length limit. Let's take this
-function for example:
+In order to guarantee code correctness, starting in version 2.0, KotlinPoet will never replace
+spaces, found in blocks of code, with new line symbols, even in cases when the line of code exceeds
```
JakeWharton, line 100: Think these aren't needed
```suggestion
spaces found in blocks of code with new line symbols, even in cases when the line of code exceeds
```

JakeWharton (approved):
We might want to come up with a better example, but this is good enough for now just to get it out the door.

PR #2004, opened 2024-10-15
Return FunSpec.Builder from jvmModifiers to not break chaining

### LINKED ISSUE #638: `jvmModifiers` doesn't return the Builder

Egorand:
Fixes #638.

- [ ] `docs/changelog.md` has been updated if applicable.
    be added to the changelog.


JakeWharton (commented):
This breaks binary compatibility. We need to retain the old signature and use `@JvmName` on the new one.

JakeWharton:
I definitely would not break anything for 2.0. The sole behavior change is the wrapping change, though that doesn't break anything. In 3.0 we could move deprecations to be hidden. I don't want someone to have a reason not to upgrade due to compatibility.

Issue #973, opened 2020-08-28
Use consistent collection type in builders for modifiers and type variables.


GlassBricks:
Right now, modifiers are represented using a `MutableSet` in `TypeAliasSpec.Builder`, `TypeSpec.Builder`, but using a`MutableList` in `FunSpec.Builder`, `ParameterSpec.Builder`, `PropertySpec.Builder`.
It would make always use a Set to match the Specs.

Type Variables are represented using a `MutableList` in `FunSpec.Builder`, `PropertySpec.Builder`, and `TypeSpec.Builder`, but a `MutableSet` in `TypeAlias.Builder`.
It would make always use a List (to match the Specs).

This would be backwards incompatible so it will have to wait until 2.0

PR #2005, opened 2024-10-15
Introduce Modifiable interface to standardized modifiers API

Egorand:
Part of #973.

- [ ] `docs/changelog.md` has been updated if applicable.
    be added to the changelog.


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -326,14 +328,6 @@ public class FunSpec private constructor(
     @ExperimentalKotlinPoetApi
     override val contextReceiverTypes: MutableList<TypeName> = mutableListOf()
 
-    public fun addModifiers(vararg modifiers: KModifier): Builder = apply {
-      this.modifiers += modifiers
-    }
-
-    public fun addModifiers(modifiers: Iterable<KModifier>): Builder = apply {
-      this.modifiers += modifiers
-    }
```
JakeWharton, line 335: These need to be retained for binary compatibility, such as with `addAnnotation` below. And in other specs.

PR #2006, opened 2024-10-15
Migrating the `Util.kt` and `KModifier` from JVM to common

ForteScarlet:
Part of #304, #1959

Migrating `Util.kt` and `KModifier` from JVM to common.

- Non-JVM platform implementations have some questionable implementations (They mark the TODO).
- There is a known issue ([KT-71003](https://youtrack.jetbrains.com/issue/KT-71003)) regarding the use of Kotlin `Regex` in WasmJS.
- there are some functions that need to wait for other types of migration.

On the JVM, however, the effect is the same before and after migration.

Let me know if there's anything I need to do, thanks.
If these changes are not reviewed until after 2.0, you can always tell me, convert it to draft, or close it~

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet
+
+import com.squareup.kotlinpoet.CodeBlock.Companion.isPlaceholder
+import java.util.Collections
+import java.util.EnumSet
+
+internal actual fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> =
+  Collections.unmodifiableMap(LinkedHashMap(this))
+
+internal actual fun <T> Collection<T>.toImmutableList(): List<T> =
+  Collections.unmodifiableList(ArrayList(this))
+
+internal actual fun <T> Collection<T>.toImmutableSet(): Set<T> =
+  Collections.unmodifiableSet(LinkedHashSet(this))
+
+internal actual fun formatIsoControlCode(code: Int): String =
+  String.format("\\u%04x", code)
+
+internal actual fun Int.toHexStr(): String =
+  Integer.toHexString(this)
+
+// TODO Waiting for `CodeBlock` migration.
+internal fun CodeBlock.ensureEndsWithNewLine() = trimTrailingNewLine('\n')
+
+// TODO Waiting for `CodeBlock` migration.
+internal fun CodeBlock.trimTrailingNewLine(replaceWith: Char? = null) = if (isEmpty()) {
+  this
+} else {
+  with(toBuilder()) {
+    val lastFormatPart = trim().formatParts.last()
+    if (lastFormatPart.isPlaceholder && args.isNotEmpty()) {
+      val lastArg = args.last()
+      if (lastArg is String) {
+        val trimmedArg = lastArg.trimEnd('\n')
+        args[args.size - 1] = if (replaceWith != null) {
+          trimmedArg + replaceWith
+        } else {
+          trimmedArg
+        }
+      }
+    } else {
+      formatParts[formatParts.lastIndexOf(lastFormatPart)] = lastFormatPart.trimEnd('\n')
+      if (replaceWith != null) {
+        formatParts += "$replaceWith"
+      }
+    }
+    return@with build()
+  }
+}
+
+private val IDENTIFIER_REGEX = IDENTIFIER_REGEX_VALUE.toRegex()
+
+internal actual val String.isIdentifier: Boolean
+  get() = IDENTIFIER_REGEX.matches(this)
+
+internal actual inline fun <reified E : Enum<E>> enumSetOf(vararg values: E): MutableSet<E> {
+  return when(values.size) {
+    0 -> EnumSet.noneOf(E::class.java)
+    1 -> EnumSet.of(values[0])
+    2 -> EnumSet.of(values[0], values[1])
+    3 -> EnumSet.of(values[0], values[1], values[2])
+    4 -> EnumSet.of(values[0], values[1], values[2], values[3])
+    5 -> EnumSet.of(values[0], values[1], values[2], values[3], values[4])
+    else -> EnumSet.copyOf(values.toSet())
+  }
+}
+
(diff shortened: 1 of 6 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
```diff
@@ -0,0 +1,87 @@
...
+  return when (values.size) {
+    0 -> EnumSet.noneOf(E::class.java)
+    1 -> EnumSet.of(values[0])
+    2 -> EnumSet.of(values[0], values[1])
+    3 -> EnumSet.of(values[0], values[1], values[2])
+    4 -> EnumSet.of(values[0], values[1], values[2], values[3])
+    5 -> EnumSet.of(values[0], values[1], values[2], values[3], values[4])
+    else -> EnumSet.copyOf(values.toSet())
+  }
+}
```
JakeWharton, line 81: There is no need to do this. Just use a regular set and mutable set in common code. We're no longer gaining any advantage of `EnumSet` having to go through this allocation-heavy path.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@OptIn(ExperimentalStdlibApi::class)
+internal fun Appendable.appendFormat04x(code: Int) {
+  val hex = code.toHexString(HexFormatWithoutLeadingZeros)
+  if (hex.length < 4) {
+    repeat(4 - hex.length) { append('0') }
+  }
+  append(hex)
+}
```
JakeWharton, line 48: This seems like it could be simplified to `toString(16).padStart(4, '0')` and placed directly in common code.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+
+@OptIn(ExperimentalStdlibApi::class)
+internal actual fun Int.toHexStr(): String =
+  toHexString(HexFormatWithoutLeadingZeros)
```
JakeWharton, line 52: Is this just `toString(16)`? If so, we can use that directly in common code (without relying on unstable APIs, as well).

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+package com.squareup.kotlinpoet
+
+internal actual fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> =
+  toMap()
+
+internal actual fun <T> Collection<T>.toImmutableList(): List<T> =
+  toList()
+
+internal actual fun <T> Collection<T>.toImmutableSet(): Set<T> =
+  toSet()
```
JakeWharton, line 25: If we're not going to put these in an unmodifiable wrapper on other targets, do we still need to do it on JVM? Are we really worried about Java callers mutating things? I don't think I am.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
ForteScarlet: Yes, it seems so. I'll simplify it

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
JakeWharton: The original implementation was copied when we forked from JavaPoet, for what it's worth.

ForteScarlet: Well, you have a point. If we no longer care about calls that mutate things on the JVM, do we keep `toImmutable*` in common and have its result just return a `to*` (like `toSet`), or do we simply remove them and use `to*` instead where appropriate

JakeWharton: I think it's defending against something that is not a real problem in practice, and I would eliminate the concept entirely and use Kotlin's collection APIs directly everywhere. This should probably be done in its own PR rather than cluttering this one.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/Util.jvm.kt
Egorand: [d7c57e4](https://github.com/square/kotlinpoet/pull/2006/commits/d7c57e4058efa53c4194db2384402f2653d0cb5f)

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
Egorand: [0920d3e](https://github.com/square/kotlinpoet/pull/2006/commits/0920d3e8a10b16f7b71484c7c642a48019ce2a7d)

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
Egorand: [0920d3e](https://github.com/square/kotlinpoet/pull/2006/commits/0920d3e8a10b16f7b71484c7c642a48019ce2a7d)

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
```diff
@@ -0,0 +1,71 @@
...
+internal actual fun Char.isJavaIdentifierStart(): Boolean {
+  return isLetter() ||
+    this in CharCategory.LETTER_NUMBER ||
+    this == '$' ||
+    this == '_'
+}
+
+internal actual fun Char.isJavaIdentifierPart(): Boolean {
+  //  TODO
+  //   A character may be part of a Java identifier if any of the following conditions are true:
```
Egorand, line 36: Should we maybe extract these checks into common to have consistent implementations on all platforms?

Egorand (approved):

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
Egorand: Can be done in later PRs.

Egorand:
@JakeWharton please merge if everything looks good to you.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/Util.nonJvm.kt
ForteScarlet: Currently, I am dividing them because I am unsure if the current implementation of non-JVM is the same as the one in JVM.
It would be wonderful to merge them if we can confirm that they have the same effects!

Egorand:
Merged the PR, seems that there are no additional concerns. Thanks!

Going forward, please feel free to stack PRs on top of each other so you're not blocked by reviews.

PR #2007, opened 2024-10-15
Track new JavaPoet group and package name

squarejesse:
Rename the existing interop module to javapoet-deprecated and introduce a new module called javapoet that integrates with the Palantir JavaPoet fork.


JakeWharton:
The old implementation was never marked as stable, but at least the JavaPoet APIs behind it were stable. These new APIs are also not marked as stable, but the Palantir fork of JavaPoet is itself not stable either.

Do we even expect anyone to use this? The only project in the entire world using the old artifact is https://github.com/zlataovce/takenaka. I'm not sure I would bother until people start asking for this, which I'm not sure anyone ever would.

Issue #2033, opened 2024-12-03
Anonymous Class TypeSpec do not support delegates


TrevorSStone:
**Describe the bug**
When creating an anonymous class using kotlinpoet, delegates are silently ignored. Delegates work on anonymous classes and thus should be supported. 

**To Reproduce**
```kotlin
  @Test fun testDelegateOnAnonymousObject() {
    val type = TypeSpec.anonymousClassBuilder()
      .addSuperinterface(
        Consumer::class.parameterizedBy(String::class),
        CodeBlock.of("java.util.function.Consumer({ println(it) })"),
      )
      .build()

    val expect = """
        |object : java.util.function.Consumer<kotlin.String> by java.util.function.Consumer({ println(it) }) {
        |}
    """.trimMargin()

    assertThat(type.toString()).isEqualTo(expect)
  }

@Test fun testMultipleDelegatesOnAnonymousObject() {
    val type = TypeSpec.anonymousClassBuilder()
      .primaryConstructor(
        FunSpec.constructorBuilder()
          .build(),
      )
      .addSuperinterface(
        Function::class.parameterizedBy(String::class, Int::class),
        CodeBlock.of("kotlin.Function ({ text -> text.toIntOrNull() ?: 0 })"),
      )
      .addSuperinterface(
        Runnable::class,
        CodeBlock.of("java.lang.Runnable ({ %T.debug(\"Hello world\") })", Logger::class.asTypeName()),
      )
      .build()

    val expect = """
        |object : kotlin.Function<kotlin.String, kotlin.Int> by kotlin.Function ({ text -> text.toIntOrNull() ?: 0 }), java.lang.Runnable by java.lang.Runnable ({ java.util.logging.Logger.debug("Hello world") }) {
        |}
    """.trimMargin()

    assertThat(type.toString()).isEqualTo(expect)
  }
```

**Expected behavior**
```

expected:
    object : java.util.function.Consumer<kotlin.String> by java.util.function.Consumer({ println(it) }) {
    }
but was:
    object : java.util.function.Consumer<kotlin.String> {
    }
<Click to see difference>

expected:
    object : java.util.function.Consumer<kotlin.String> by java.util.function.Consumer({ println(it) }) {
    }
but was:
    object : java.util.function.Consumer<kotlin.String> {
    }
	at com.squareup.kotlinpoet.TypeSpecTest.testDelegateOnAnonymousObject(TypeSpecTest.kt:4433)

```

**Additional context**
I have a fix and will link the PR

PR #2034, opened 2024-12-03
Add support for delegates to Anonymous classes

TrevorSStone:
    be added to the changelog.

Fixes https://github.com/square/kotlinpoet/issues/2033

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
@@ -159,7 +159,7 @@ public class TypeSpec private constructor(
         }
 
         if (allSuperTypes.isNotEmpty()) {
-          codeWriter.emitCode(allSuperTypes.joinToCode(separator = ", ", prefix = " : "))
+          codeWriter.emitCode(allSuperTypes.joinToCode(prefix = " : "))
         }
         if (hasNoBody) {
           codeWriter.emit(" {\n}")
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
@@ -142,23 +142,24 @@ public class TypeSpec private constructor(
         codeWriter.emitCode("object")
         val supertype = if (superclass != ANY) {
           if (!areNestedExternal && !modifiers.contains(EXPECT)) {
-            listOf(CodeBlock.of(" %T(%L)", superclass, superclassConstructorParametersBlock))
+            listOf(CodeBlock.of("%T(%L)", superclass, superclassConstructorParametersBlock))
           } else {
-            listOf(CodeBlock.of(" %T", superclass))
+            listOf(CodeBlock.of("%T", superclass))
           }
         } else {
           listOf()
         }
 
-        val allSuperTypes = supertype + if (superinterfaces.isNotEmpty()) {
-          superinterfaces.keys.map { CodeBlock.of(" %T", it) }
-        } else {
-          emptyList()
+        val allSuperTypes = supertype + superinterfaces.entries.map { (type, init) ->
+          if (init == null) {
+            CodeBlock.of("%T", type)
+          } else {
+            CodeBlock.of("%T by %L", type, init)
+          }
         }
 
         if (allSuperTypes.isNotEmpty()) {
-          codeWriter.emitCode(" :")
-          codeWriter.emitCode(allSuperTypes.joinToCode(","))
+          codeWriter.emitCode(allSuperTypes.joinToCode(separator = ", ", prefix = " : "))
         }
         if (hasNoBody) {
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -4416,6 +4416,47 @@ class TypeSpecTest {
...
+
+    assertThat(type.toString()).isEqualTo(expect)
+  }
+  // https://github.com/square/kotlinpoet/issues/2033
+  @Test fun testMultipleDelegatesOnAnonymousObject() {
+    val type = TypeSpec.anonymousClassBuilder()
+      .primaryConstructor(
+        FunSpec.constructorBuilder()
+          .build(),
+      )
```
Egorand, line 4441: nit: the constructor is ignored, right? Or does removing it actually modify the output?

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
```diff
@@ -4416,6 +4416,47 @@ class TypeSpecTest {
...
+      .addSuperinterface(
+        Function::class.parameterizedBy(String::class, Int::class),
+        CodeBlock.of("kotlin.Function ({ text -> text.toIntOrNull() ?: 0 })"),
+      )
+      .addSuperinterface(
+        Runnable::class,
+        CodeBlock.of("java.lang.Runnable ({ %T.debug(\"Hello world\") })", Logger::class.asTypeName()),
```
Egorand, line 4448: nit - this should not be necessary:

```suggestion
        CodeBlock.of("java.lang.Runnable ({ %T.debug(\"Hello world\") })", Logger::class),
```

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -142,23 +142,24 @@ public class TypeSpec private constructor(
...
+            CodeBlock.of("%T", type)
+          } else {
+            CodeBlock.of("%T by %L", type, init)
+          }
         }
 
         if (allSuperTypes.isNotEmpty()) {
-          codeWriter.emitCode(" :")
-          codeWriter.emitCode(allSuperTypes.joinToCode(","))
+          codeWriter.emitCode(allSuperTypes.joinToCode(separator = ", ", prefix = " : "))
```
Egorand, line 162: Nice use of `joinToCode()`!

nit - this is the default separator so isn't necessary:

```suggestion
          codeWriter.emitCode(allSuperTypes.joinToCode(prefix = " : "))
```

Egorand (approved):
Awesome, thanks for the fix!

Left a few suggestions, would be great to incorporate them but not critical.

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
TrevorSStone: You're correct. This was copy pasta from the `testMultipleDelegates` test I based mine off of

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/TypeSpecTest.kt
TrevorSStone: you're correct. I'll also remove it from the orig test

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
TrevorSStone: My personal philosophy would be to keep the separator as-is. We "want" it to be `", "`, the fact that it is also the default behavior is incidental.

TrevorSStone: Actually, I was thinking about this like using a library, but this is actually defined in this project, so I'll change it. Thanks for the review

Egorand: Agree, specifying the argument explicitly doesn't hurt either, but yeah, mostly from the perspective that we own that method, plus `", "` seems to be a pretty standard separator in "joinTo" Kotlin methods.

Egorand (approved):

Egorand:
@TrevorSStone can you please run `./gradlew spotlessApply` and push to fix code style issues?

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
JakeWharton: And like Kotlin's `joinToString`, we can never really change the default value or it would basically break all of its users.

PR #2040, opened 2024-12-16
Migrate OriginatingElementsHolder and Taggable from JVM to common

ForteScarlet:
Part of https://github.com/square/kotlinpoet/issues/304, https://github.com/square/kotlinpoet/pull/1959

Migrate `OriginatingElementsHolder` and `Taggable` from JVM to common

### DIFF
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/OriginatingElementsHolder.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/OriginatingElementsHolder.kt
@@ -15,24 +15,26 @@
  */
 package com.squareup.kotlinpoet
 
-import javax.lang.model.element.Element
+import com.squareup.kotlinpoet.jvm.alias.JvmDefaultWithCompatibility
+import com.squareup.kotlinpoet.jvm.alias.JvmElement
+import kotlin.jvm.JvmInline
 
 /** A type that can have originating [elements][Element]. */
 public interface OriginatingElementsHolder {
 
   /** The originating elements of this type. */
-  public val originatingElements: List<Element>
+  public val originatingElements: List<JvmElement>
 
   /** The builder analogue to [OriginatingElementsHolder] types. */
   @JvmDefaultWithCompatibility
   public interface Builder<out T : Builder<T>> {
 
     /** Mutable map of the current originating elements this builder contains. */
-    public val originatingElements: MutableList<Element>
+    public val originatingElements: MutableList<JvmElement>
 
     /** Adds an [originatingElement] to this type's list of originating elements. */
     @Suppress("UNCHECKED_CAST")
-    public fun addOriginatingElement(originatingElement: Element): T = apply {
+    public fun addOriginatingElement(originatingElement: JvmElement): T = apply {
       originatingElements += originatingElement
     } as T
   }
@@ -41,10 +43,10 @@ public interface OriginatingElementsHolder {
 internal fun OriginatingElementsHolder.Builder<*>.buildOriginatingElements() =
   OriginatingElements(originatingElements.toImmutableList())
 
-internal fun List<Element>.buildOriginatingElements() =
+internal fun List<JvmElement>.buildOriginatingElements() =
   OriginatingElements(toImmutableList())
 
 @JvmInline
 internal value class OriginatingElements(
-  override val originatingElements: List<Element>,
+  override val originatingElements: List<JvmElement>,
 ) : OriginatingElementsHolder
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+import kotlin.reflect.KClass
+
+/**
+ * An expected typealias for `java.lang.Class`.
+ */
+public expect class JvmClass<T : Any>
+
+/**
+ * Convert [JvmClass] to [KClass].
+ */
+public expect val <T : Any> JvmClass<T>.kotlin: KClass<T>
--- a/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmTypeAliasKotlinPoetApi.kt
+++ b/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmTypeAliasKotlinPoetApi.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+/**
+ * An expected type for JVM to `typealias` only.
+ */
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY, AnnotationTarget.TYPEALIAS)
+@RequiresOptIn(
+  level = RequiresOptIn.Level.ERROR,
+  message = "An expected type for JVM to `typealias` only." +
+    " Don't implement or use it in non-JVM platforms.",
+)
+public annotation class JvmTypeAliasKotlinPoetApi
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.jvm.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.jvm.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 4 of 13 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmElement.jvm.kt
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+import javax.lang.model.element.Element
+import javax.lang.model.element.ExecutableElement
+import javax.lang.model.element.TypeElement
+import javax.lang.model.element.VariableElement
+
+/**
```
Egorand, line 23: nit: docs in this file feel redundant, they more or less just restate what the code says.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmTypeAliasKotlinPoetApi.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY, AnnotationTarget.TYPEALIAS)
+@RequiresOptIn(
+  level = RequiresOptIn.Level.ERROR,
+  message = "An expected type for JVM to `typealias` only." +
+    " Don't implement or use it in non-JVM platforms.",
+)
+public annotation class JvmTypeAliasKotlinPoetApi
```
Egorand, line 29: Should we maybe call it `JvmOnlyKotlinPoetApi`? IMO this makes it a bit more clear that this type only exists in KotlinPoet for JVM.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import kotlin.reflect.KClass
+
+/**
+ * An expected typealias for `java.lang.Class`.
+ */
+public expect class JvmClass<T : Any>
+
+/**
+ * Convert [JvmClass] to [KClass].
```
Egorand, line 26: nit: redundant comment.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.jvm.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlin.jvm.kotlin as jvmKotlin
+import kotlin.reflect.KClass
+
+/**
+ * An expected typealias for [Class].
+ */
+public actual typealias JvmClass<T> = Class<T>
+
+/**
+ * Convert [JvmClass] to [KClass].
```
Egorand, line 27: nit: redundant comment.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.jvm.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+import kotlin.jvm.kotlin as jvmKotlin
+import kotlin.reflect.KClass
+
+/**
+ * An expected typealias for [Class].
```
Egorand, line 22: nit: redundant comment.

kotlinpoet/src/nonJvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.nonJvm.kt
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+import kotlin.reflect.KClass
+
+@JvmTypeAliasKotlinPoetApi
+public actual class JvmClass<@Suppress("unused")
+  T : Any,> private constructor()
```
Egorand, line 22: nit:

```suggestion
public actual class JvmClass<@Suppress("unused") T : Any> private constructor()
```

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmDefaultWithCompatibility.jvm.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.kotlinpoet.jvm.alias
+
+/**
+ * A typealias annotation for [kotlin.jvm.JvmDefaultWithCompatibility].
```
Egorand, line 19: nit: redundant comment.

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+public expect val <T : Any> JvmClass<T>.kotlin: KClass<T>
```
Egorand, line 28: Should we make this `val` internal? It shouldn't really be used outside of the library.

Egorand (approved):
Looks good, left a few nits! Thanks!

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmTypeAliasKotlinPoetApi.kt
ForteScarlet: Sounds good!

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/jvm/alias/JvmClass.kt
ForteScarlet: Indeed, it doesn't seem to be used outside. I would change it to `internal` .

kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/OriginatingElementsHolder.kt
```diff
@@ -15,24 +15,26 @@
  */
 package com.squareup.kotlinpoet
 
-import javax.lang.model.element.Element
+import com.squareup.kotlinpoet.jvm.alias.JvmDefaultWithCompatibility
+import com.squareup.kotlinpoet.jvm.alias.JvmElement
+import kotlin.jvm.JvmInline
 
 /** A type that can have originating [elements][Element]. */
 public interface OriginatingElementsHolder {
```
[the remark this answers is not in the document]
ForteScarlet: It is implemented by many other types, e.g. `FunSpec`, `TypeSpec`. I think even though it's JVM-only, it still needs to be in the common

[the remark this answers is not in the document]
JakeWharton: The expect/actual mechanism allows actuals on the JVM to implement interfaces not exposed by the expect.

Here's an example in Okio:
- Expect: https://github.com/square/okio/blob/b6b4d52e5f0099328a5bc288249058e573dc19ff/okio/src/commonMain/kotlin/okio/Sink.kt#L45
- Actual: https://github.com/square/okio/blob/b6b4d52e5f0099328a5bc288249058e573dc19ff/okio/src/jvmMain/kotlin/okio/Sink.kt#L22

We should not expose JVM-only APIs in common. The need to create a ton of dummy typealiases to JVM types is a good indicator for an API that should not be in common.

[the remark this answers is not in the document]
ForteScarlet: In the case of `FunSpec`, for example, if you want to use the expect/actual mechanism to make `FunSpec` in the JVM implement `OriginatingElementsHolder` in a way that other platforms don't, then `FunSpec` itself must be expect/actual (unless the extra in the interface is not abstract).

This can create a lot of duplicate logic or generate a lot of extra work. That's what I use typealias for: to reduce the amount of work due to compatibility.

If we don't accept any JVM-only typealias exposed in common, then almost all major types will become expect/actual, even though their logic doesn't really differ.

This is the same problem you mentioned in https://github.com/square/kotlinpoet/issues/304#issuecomment-1582560046, and I'm trying to get around it with typealias.

Exposing JVM-only types in common I think is a compromise for compatibility and maybe acceptable? 🤔 If it's unacceptable, that's fine too, except that later migrations to those major types will become more difficult.

> There doesn't seem to be any best of both worlds until [KT-20427](https://youtrack.jetbrains.com/issue/KT-20427/Allow-expect-declarations-with-a-default-implementation) is resolved?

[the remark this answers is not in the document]
Egorand: Would that require expect/actuals of `FunSpec` and co?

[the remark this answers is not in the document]
ForteScarlet: If an interface is expect/actual and there is an extra abstract method in the JVM, then `FunSpec` must also be expect/actual. Otherwise, in common it can't know about that extra part of the JVM and thus can't implement it.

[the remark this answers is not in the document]
JakeWharton: The primary reason this work hasn't been done is because of this problem. I don't love the idea of exposing a large, JVM-based API surface that is effectively useless to other platforms.

I would rather we either wait for the language to evolve, or explore alternative ways of simplifying the duplication which comes from doing it properly today. For example, we could pull out sealed abstract base classes which contain the common code of each type, and then `expect`/`actual` a mostly-empty subtype which then would allow retaining platform-specific API surfaces.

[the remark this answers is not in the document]
ForteScarlet: This is indeed tricky. Even more painful is the fact that [KT-20427](https://youtrack.jetbrains.com/issue/KT-20427/Allow-expect-declarations-with-a-default-implementation) has made little visible progress to date. 😭
<small><i>(Come on, Jetbrains! ✊)</i></small>

But it seems that most of the forced ‘compatibility’ is related to `DelicateKotlinPoetApi` or `javax.lang.model`. Do they have an obvious role in the Kotlin API? Or do they need to be kept? If they are not planned to be retained in the future (3.0, for example), then some of the current _interim measures_ might be acceptable? Personal opinion.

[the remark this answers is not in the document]
JakeWharton: K2 blocked all significant language features for nearly three years. They're only starting to catch up to addressing things, but I still wouldn't hold my breath for that feature for a while.

Delicate APIs are still stable APIs–we can't break them. The usage of Java mirrors and elements are tied to annotation processors which generate Kotlin. While kapt is set to be deprecated, we need to retain those APIs for existing tools.

We could attempt to pursue a 3.x by developing a multiplatform variant of KotlinPoet in common _from scratch_ in the `app.cash.kotlinpoet` package. We would develop this over time, allowing people to try it out in the 2.x releases since it would sit as a sibling. Then, when we're happy, we would delete the `com.squareup.kotlinpoet` API in its entirety, change the Maven groupId, and release 3.0. This would ensure that 2.x users never see a breakage, while allowing 3.x to be iterated on and tested.

It would allow fixing design mistakes both in regard to how the JVM-specific APIs are integrated, but also the pervasive use of builders instead of something more appropriate to Kotlin such as fully mutable types that get captured into immutable specs, with helpers for things like moving the mutable types into the receiver position of a function.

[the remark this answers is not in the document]
ForteScarlet: Sounds like a good programme!🤩 Please feel free to let me know if there is anything I can do to help!

PR #2041, opened 2024-12-16
Migrate LineWrapper from JVM to common

ForteScarlet:
Part of https://github.com/square/kotlinpoet/issues/304, https://github.com/square/kotlinpoet/pull/1959

Migrate `LineWrapper` from JVM to common


Egorand (approved):
That was easy!

PR #2052, opened 2024-12-30
Prevent aliased imports from colliding with existing imports

### LINKED ISSUE #2046: Import aliases may introduce new naming collisions
**Describe the bug**
Import aliases introduced to avoid name collisions may themselves introduce new name collisions.

**To Reproduce**
```kotlin
class ImportAliasTest {

    @Test
    fun `name collision`() {
        val dupName1 = ClassName("com.foo", "MyClass")
        val dupName2 = ClassName("com.bar", "FooMyClass")
        val dupName3 = ClassName("com.baz", "MyClass")
        val file = FileSpec
            .builder("com.sample", "Sample")
            .addFunction(
                FunSpec.builder("foo")
                    .addStatement("%T", dupName1)
                    .addStatement("%T", dupName2)
                    .addStatement("%T", dupName3)
                    .build()
            )
            .build()
        assertEquals("?", file.toString())
    }
}
```

It is not clear what KotlinPoet _should_ generate (hence `"?"`), but what it does generate currently is invalid:

```Kotlin
package com.sample

import com.bar.FooMyClass
import com.baz.MyClass as BazMyClass
import com.foo.MyClass as FooMyClass

public fun foo() {
  com.foo.MyClass
  FooMyClass
  BazMyClass
}
```

**Expected behavior**
The generated code should not have any name collisions. One option might be:
```kotlin
package com.sample

import com.bar.FooMyClass as BarFooMyClass
import com.baz.MyClass as BazMyClass
import com.foo.MyClass as FooMyClass

public fun foo() {
  FooMyClass
  BarFooMyClass
  BazMyClass
}
```

damianw:
Fixes #2046.

The simplest solution seemed to be to also check against currently-imported names when generating import aliases -- if a name is already imported, use more segments. This introduced the possibility that we might run out of segments though, which I handled by appending underscores if we reach that condition.

- ✅  `docs/changelog.md` has been updated if applicable
- ✅ [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) signed.


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -777,24 +778,33 @@ internal class CodeWriter(
...
-      while (aliasNames.size != canonicalNamesToQualifiedNames.size) {
+      outer@ while (aliasNames.size != canonicalNamesToQualifiedNames.size) {
         segmentsToUse += 1
         aliasNames.clear()
         for ((segments, qualifiedName) in canonicalNameSegmentsToQualifiedNames) {
           val aliasPrefix = segments.takeLast(min(segmentsToUse, segments.size))
             .joinToString(separator = "")
             .replaceFirstChar { if (!capitalizeAliases) it.lowercaseChar() else it }
-          val aliasName = aliasPrefix + simpleName.replaceFirstChar(Char::uppercaseChar)
+          val aliasSuffix = (0..<(segmentsToUse - segments.size)).joinToString("") { "_" }
```
JakeWharton, line 800: I think this is
```suggestion
          "_".repeat(segmentsToUse - segments.size)
```

JakeWharton (approved):

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
damianw: Ah perfect, thanks!

damianw: Because this value can (and usually is) negative, this needed to be `"_".repeat((segmentsToUse - segments.size).coerceAtLeast(0))`

JakeWharton:
Will wait for others to check it out later this week, or next. It's an off week for everyone (or it should be).

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -751,7 +752,7 @@ internal class CodeWriter(
       val imported = mutableMapOf<String, Set<T>>()
       forEach { (simpleName, qualifiedNames) ->
         val canonicalNamesToQualifiedNames = qualifiedNames.associateBy { it.computeCanonicalName() }
-        if (canonicalNamesToQualifiedNames.size == 1 && simpleName !in referencedNames) {
+        if (canonicalNamesToQualifiedNames.size == 1 && simpleName !in referencedNames && simpleName !in imported) {
```
Egorand, line 755: My understanding is that the `simpleName !in imported` is needed for the case demonstrated by the `importAliasCollisionWithRegularImportWhenAliasesComputedFirst` test: if there's a generated alias that conflicts with a simple name of a different class, we'll be forced to generate an alias for that simple name, so non-conflicting simple names don't take precedence. I think this is a bit unfortunate, and I wonder if pre-sorting `(simpleName, qualifiedNames)` pairs by the number of `qualifiedNames` would be an easy improvement here? That way we'll first process non-conflicting names and only then get to conflicts that require aliases.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -787,14 +789,21 @@ internal class CodeWriter(
...
         aliasNames.clear()
         for ((segments, qualifiedName) in canonicalNameSegmentsToQualifiedNames) {
           val aliasPrefix = segments.takeLast(min(segmentsToUse, segments.size))
             .joinToString(separator = "")
             .replaceFirstChar { if (!capitalizeAliases) it.lowercaseChar() else it }
-          val aliasName = aliasPrefix + simpleName.replaceFirstChar(Char::uppercaseChar)
+          val aliasSuffix = "_".repeat((segmentsToUse - segments.size).coerceAtLeast(0))
+          val aliasName = aliasPrefix + simpleName.replaceFirstChar(Char::uppercaseChar) + aliasSuffix
+          // If this name has already been imported (e.g. a regular import already exists with this name),
+          // continue trying with a greater number of segments.
```
Egorand, line 802: It seems that we do this for all names if we encounter a conflict for just one of them, as demonstrated by the `importAliasCollisionWithImportWhenRegularImportComputedFirst` test: if we linked `com.squareup.meal.TacoMeal` to `TacoMeal`, we'll generate `import com.squareup.burrito.Meal as SquareupBurritoMeal` and `import com.squareup.taco.Meal as SquareupTacoMeal` imports, even though `import com.squareup.burrito.Meal as BurritoMeal` would've been a shorter non-conflicting alias import. Can we maybe track `canonicalNamesToQualifiedNames` we've found non-conflicting imports for, and only try a larger number of segments for those that are still conflicting?

Egorand (approved):
Looks good! Left a couple improvement suggestions, they're largely cosmetic since the the top priority is the correctness of generated aliases, but if you have the appetite for it - would be great to land them before merging!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
damianw: Sure, that makes sense. I can make that change - this is also probably better because the imports end up being the same regardless of which appears in the file first.

Egorand: Yep!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
damianw: Yup that's true. I think for this one I'd like to leave it as-is if that's alright with you since optimizing this part is a little more involved than what you pointed out in the first comment. I think this scenario should be rare enough that it's probably not likely to make that much of a difference.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt
```diff
@@ -749,34 +750,37 @@ internal class CodeWriter(
...
-          generateImportAliases(simpleName, canonicalNamesToQualifiedNames, capitalizeAliases)
-            .onEach { (a, qualifiedName) ->
-              val alias = a.escapeAsAlias()
-              val canonicalName = qualifiedName.computeCanonicalName()
-              generatedImports[canonicalName] = Import(canonicalName, alias)
-
-              imported[alias] = setOf(qualifiedName)
-            }
+      entries
+        .sortedBy { (_, qualifiedNames) -> qualifiedNames.size }
```
Egorand, line 754:
```suggestion
      entries
        // Pre-sorting entries by the number of qualified names to ensure we first process simple names
        // that don't require aliases (qualifiedNames.size == 1), that way we don't run into conflicts when
        // a newly generated alias now clashes with a simple name that didn't originally need an alias.
        .sortedBy { (_, qualifiedNames) -> qualifiedNames.size }
```

Egorand:
@JakeWharton this should be ready for merging.

Issue #2114, opened 2025-04-16
Allow to use context parameters as Names


IRus:
**Is your feature request related to a problem? Please describe.**

As @Egorand commented on https://github.com/square/kotlinpoet/pull/2112#discussion_r2047066894 it would be nice to be able to use Context Parameters in positions where name `%N` is expected:

```
val logger = ContextParameter("logger", Logger::class)
val config = ContextParameter("config", Config::class)
val processData = FunSpec.builder("processData")
  .contextParameter(logger)
  .contextParameter(config)
  .addStatement("%N.info(\"Processing with config: ${'$'}%N\")", logger, config)
  .build()
```

**Describe the solution you'd like**

Allow to use ContextParameter as a name

**Describe alternatives you've considered**

**Additional context**

IRus:
Closed by https://github.com/square/kotlinpoet/pull/2117

PR #2117, opened 2025-04-17
Allow to use context parameters as Names (#2114)

IRus:
    be added to the changelog.

### DIFF
--- a/docs/context-parameters.md
+++ b/docs/context-parameters.md
@@ -22,10 +22,16 @@ public fun greet() {
 You can add multiple context parameters:
 
 ```kotlin
+val loggerType = ClassName("java.util.logging", "Logger")
+val configType = ClassName("com.example", "Config")
+
+val logger = ContextParameter("logger", loggerType)
+val config = ContextParameter("config", configType)
+
 val processData = FunSpec.builder("processData")
-  .contextParameter("logger", ClassName("com.example", "Logger"))
-  .contextParameter("config", ClassName("com.example", "Config"))
-  .addStatement("logger.info(\"Processing with config: ${'$'}config\")")
+  .contextParameter("logger", loggerType)
+  .contextParameter("config", configType)
+  .addStatement("%N.info(\"Processing with config: ${'$'}%N\")", logger, config)
   .build()
 ```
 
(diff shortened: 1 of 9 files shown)


docs/context-parameters.md
```diff
@@ -22,10 +22,16 @@ public fun greet() {
...
+
+val logger = ContextParameter("logger", loggerType)
+val config = ContextParameter("config", configType)
+
 val processData = FunSpec.builder("processData")
-  .contextParameter("logger", ClassName("com.example", "Logger"))
-  .contextParameter("config", ClassName("com.example", "Config"))
-  .addStatement("logger.info(\"Processing with config: ${'$'}config\")")
+  .contextParameter("logger", loggerType)
+  .contextParameter("config", configType)
```
Egorand, line 33: nit:

```suggestion
  .contextParameter(logger)
  .contextParameter(config)
```

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -702,6 +702,30 @@ class FunSpecTest {
...
+  @Test fun contextParameterInAddStatement() {
+    val loggerType = ClassName("java.util.logging", "Logger")
+    val configType = ClassName("com.example", "Config")
+
+    val logger = ContextParameter("logger", loggerType)
+    val config = ContextParameter("config", configType)
+
+    val processData = FunSpec.builder("processData")
+      .contextParameter("logger", loggerType)
+      .contextParameter("config", configType)
```
Egorand, line 714: same:

```suggestion
      .contextParameter(logger)
      .contextParameter(config)
```

Egorand (approved):
Very nice, thanks!

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/FunSpecTest.kt
IRus: Nice addition!

 I double check current implementation, removed overrides that I think actually not required (I was basically copy-pasting receivers implementation initially, and now done more assessment to what actually have to be done).
 And added context parameters into `toBuilder` functions.

Egorand: Looks great, thank you!

PR #2142, opened 2025-05-22
Fix annotation array args with annotation elements

haruue:
Previously, `AnnotationSpec.Builder.memberForValue()` did not correctly handle annotation instances inside annotation array parameters. For example, given:

```kotlin
annotation class AnnotationC(val value: String)
annotation class AnnotationD(
  val s: Array<AnnotationC> = [AnnotationC("foo")]
)
```

When generating code from existing annotation instances using reflection, KotlinPoet would produce the following invalid Kotlin code:

```kotlin
@AnnotationD(
  s = arrayOf(@com.example.AnnotationC("foo")),
)
public class Taco
```

With this change, annotation values in arrays are handled by delegating to `AnnotationSpec.get()`, producing valid Kotlin code. Related test functions have also been updated accordingly.

    be added to the changelog.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,8 @@ Change Log
 
 ## Unreleased
 
+ * Fix: Annotation array parameters with annotation elements now correctly handled. (#2142)
+
 ## Version 2.2.0
 
 Thanks to [@IRus][IRus] for contributing to this release.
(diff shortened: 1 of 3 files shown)


Egorand (approved):
Awesome, thanks!

Can you please update the changelog? Just add a line describing your change under the "Unreleased" section, similar to existing entries.

PR #2162, opened 2025-06-28
Handle recursively bound generics in KType.asTypeName

### LINKED ISSUE #1914: Stack overflow when generating names if types are "curiously recurring" (recursively bound generics)
**Describe the bug**
Building a type name for a type whose bound is *recursively generic* produces a stack overflow.
However, this pattern of use of generics is very common, and used also in the standard library (see example). 

**To Reproduce**

Consider the following `main` :

```
import com.squareup.kotlinpoet.asTypeName
import kotlin.reflect.jvm.kotlinFunction

fun main() {
    Class.forName("kotlin.collections.ArraysKt").methods.first {
        // Selects fun <T : Comparable<T>> Array<T>.max(): T
        it.name == "max" && it.parameters.first().parameterizedType.typeName.contains("Comparable")
    }.kotlinFunction?.parameters?.first()?.type
        ?.asTypeName() // Stack overflow due to the recurring type.
}
```

**Expected behavior**
A `TypeName` for `T : Comparable<T>` is produced.

**Additional context**
We are writing a code generator that runs through the standard library, and we noticed we cannot use it on comparators.

Egorand:
Fixes #1914.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ reports
 *.iml
 classes
 
+.vscode
+
 # Mkdocs files
 docs/1.x/*
 site
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,6 +4,7 @@ Change Log
 ## Unreleased
 
  * Fix: Annotation array parameters with annotation elements now correctly handled. (#2142)
+ * Fix: `KType.asTypeName` now correctly handles recursively bound generics (e.g. `T : Comparable<T>`). (#1914)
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -17,6 +17,8 @@ reports
 *.iml
 classes
 
+.vscode
```
ZacSweers, line 20: 👀

ZacSweers (approved):

.gitignore
Egorand: That's for vibe-coders!

PR #2234, opened 2025-10-23
Full stack traces always

JakeWharton:


JakeWharton:
I'll take bad defaults for 500, Alex.

PR #2266, opened 2025-12-17
Enhance import generation for extension members in KDocs

### LINKED ISSUE #2189: Imports are not resolved for extension properties used via %M in KDocs
**Describe the bug**
When defining a `MemberName` with `isExtension = true`, if the only usage is in a KDoc via `%M`, the FQ name will get embedded rather than the simple name. If another usage of the member exists in the file, the simple name will get embedded.

**To Reproduce**

This works as expected:

```kotlin
fun main() {
  val dpFactoryFunction =
    MemberName(packageName = "androidx.compose.ui.unit", simpleName = "dp", isExtension = true)

  val properties =
    listOf(
      PropertySpec.builder("bar", com.squareup.kotlinpoet.INT)
        .initializer("42")
        .addKdoc("`42.%M`", dpFactoryFunction)
        .build(),
      PropertySpec.builder("baz", com.squareup.kotlinpoet.FLOAT)
        .initializer("42.%M.value", dpFactoryFunction)
        .build(),
    )

  val file = FileSpec.builder("foo", "Foo").addProperties(properties).build()
  file.writeTo(System.out)
}
```

Generating the following (correct) code:

```kotlin
package foo

import androidx.compose.ui.unit.dp
import kotlin.Float
import kotlin.Int

/**
 * `42.dp`
 */
public val bar: Int = 42

public val baz: Float = 42.dp.value
```

Removing the `baz` property results in the FQ name being used (incorrectly) in the remaining `bar` property's KDoc:

```kotlin
package foo

import kotlin.Int

/**
 * `42.androidx.compose.ui.unit.dp`
 */
public val bar: Int = 42
```

I also tried manually calling `.addImport` with the `dpFactoryFunction` on the `FileSpec.Builder`, but the resolution behavior does not change.

**Expected behavior**
Unclear to me if the usage of an extension via `%M` _only_ in imports should trigger the generation of an import (I personally do prefer it that way), or if it should effectively do what `%N` does and just output `dp` in this case without being concerned about imports.

I think it could be argued that the right thing to use here is `%N` and not `%M`, but it seems unexpected to me that I get the intended behavior if another property exists in the file that triggers the import.

**Additional context**
Currently the workaround of just switching from `%M` to `%N` does work, but I had some code-paths that were being shared for KDoc value generation and actual code generation that worked nicely until I hit this edge-case.

RaoPrashanth:
# Fix: Generate imports for extension members used via %M in KDocs

Refactored the condition to prioritize extension members for import generation regardless of the KDoc context:
- Extension members are now always marked as importable, ensuring they can be referenced by simple name and have proper clickable links in generated documentation
- Non-extension members in KDocs continue to use the existing behavior (not marked as importable)

**Fixes #2189**

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/MemberNameTest.kt
+++ b/kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/MemberNameTest.kt
@@ -714,4 +714,176 @@ class MemberNameTest {
           .trimIndent()
       )
   }
+
+  // https://github.com/square/kotlinpoet/issues/2189
+  @Test
+  fun `extension member used only in kdoc generates import`() {
+    val dpFactoryFunction =
+      MemberName(packageName = "androidx.compose.ui.unit", simpleName = "dp", isExtension = true)
+
+    val property =
+      PropertySpec.builder("bar", INT)
+        .initializer("42")
+        .addKdoc("`42.%M`", dpFactoryFunction)
+        .build()
+
+    val file = FileSpec.builder("foo", "Foo").addProperty(property).build()
+    assertThat(file.toString())
+      .isEqualTo(
+        """
+        |package foo
+        |
+        |import androidx.compose.ui.unit.dp
+        |import kotlin.Int
+        |
+        |/**
+        | * `42.dp`
+        | */
+        |public val bar: Int = 42
+        |"""
+          .trimMargin()
+      )
+  }
+
+  // https://github.com/square/kotlinpoet/issues/2189
+  @Test
+  fun `extension member used in kdoc and code generates single import`() {
+    val dpFactoryFunction =
+      MemberName(packageName = "androidx.compose.ui.unit", simpleName = "dp", isExtension = true)
+
+    val properties =
+      listOf(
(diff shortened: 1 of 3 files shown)


JakeWharton (approved):
Tests seem a bit overkill/redundant but that's fine. Looks good otherwise.

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/MemberNameTest.kt
```diff
@@ -714,4 +714,176 @@ class MemberNameTest {
...
+  }
+
+  // https://github.com/square/kotlinpoet/issues/2189
+  @Test
+  fun `non-extension member used only in kdoc generates import and linkable`() {
+    val createTaco = MemberName("com.squareup.tacos", "createTaco")
+
+    val function =
+      FunSpec.builder("main")
+        .addKdoc("Creates a taco using [%M]\n", createTaco)
```
Egorand: I wonder if it'd be useful to always wrap %T or %M members appearing in KDoc with [] so they're automatically clickable? We can consider this as a follow-up.

PR #2315, opened 2026-05-16
Keep // prefix on wrapped file comment lines

### LINKED ISSUE #1922: Wrapped comments are missing the \\ prefix
**Describe the bug**
When adding a comment that has long text that is long enough to wrap, additional lines are indented and do not have the \\ at the beginning

**To Reproduce**
```
val reallyLongLine = "This is a really long line of text that will have to be wrapped when it is used in a comment because it is so long"

someFileSpec.addFileComment("%L", reallyLongLine)
```
which produces this file comment:
```
// This is a really long line of text that will have to be wrapped when it is used in a comment
        because it is so long
```
**Expected behavior**
It should produce this file comment:
```
// This is a really long line of text that will have to be wrapped when it is used in a comment
// because it is so long
```
**Additional context**
I believe the issue are [these 2 lines](https://github.com/square/kotlinpoet/blob/d2090c3bb2f9935cbbef3d2bcc1b092b15558bc1/kotlinpoet/src/commonMain/kotlin/com/squareup/kotlinpoet/CodeWriter.kt#L621), which should probably be something more like this:
```
          indentLevel = if (kdoc || comment) indentLevel else indentLevel + 2,
          linePrefix = if (kdoc) " * " else if (comment) "// " else "",
 ```

eyupcanakman:
When a file comment added with `addFileComment` is long enough to wrap, the wrapped continuation lines kept their indentation but dropped the `//` prefix, so the wrapped text stopped being a valid comment. The wrapping path in `CodeWriter` already keeps the KDoc star prefix but passed an empty prefix for comments. Comments now get the same handling.

Fixes #1922.

    be added to the changelog.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,8 @@ Change Log
 
 ## Unreleased
 
+ * Fix: Keep the `//` prefix on wrapped file comment lines. (#1922)
+
 ## Version 2.3.0
 
 Thanks to [@haruue][haruue], [@hfhbd][hfhbd], [@yz4230][yz4230], [@mina-jaff][mina-jaff], [@BoD][BoD],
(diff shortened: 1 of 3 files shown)


Egorand (approved):
Looks good, thank you!

PR #2320, opened 2026-05-23
Handle recursively bound generics in TypeVariableName equals/hashCode

### LINKED ISSUE #1737: TypeName subclasses can cause StackOverflowError in hashCode/equals in cyclic types
**Describe the bug**
When calling hashCode/equals on a TypeName with a cyclic definition, there can be a StackOverflowError

**To Reproduce**
I'm not 100% what type is causing the issue. However, reading the implementations of these functions, it's clear that they call into on another, with no guards for the cyclic case.

**Expected behavior**
These functions should not throw.

**Additional context**
hashCode is relatively easy to fix by ignoring potentially cyclic fields. However, equals is trickier.

Having solved a [very similar problem in the JSCompiler typesystem](https://github.com/google/closure-compiler/blob/1d74e90f6e9c5316099e56988cd4b1e803e96119/src/com/google/javascript/rhino/jstype/EqualityChecker.java), the solution was to implement equals as a visitor that traversed over the type-graph, accumulating state to detect cycles.

Here are the top few loops in the stacktrace for equals:
```
Exception in thread "main" java.lang.StackOverflowError
        at java.base/java.util.Collections$UnmodifiableCollection$1.<init>(Collections.java:1074)
        at java.base/java.util.Collections$UnmodifiableCollection.iterator(Collections.java:1074)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:623)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.TypeName.equals(TypeName.kt:103)
        at com.squareup.kotlinpoet.WildcardTypeName.equals(WildcardTypeName.kt:58)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.ParameterizedTypeName.equals(ParameterizedTypeName.kt:115)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.TypeVariableName.equals(TypeVariableName.kt:82)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.WildcardTypeName.equals(WildcardTypeName.kt:63)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.ParameterizedTypeName.equals(ParameterizedTypeName.kt:115)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.TypeVariableName.equals(TypeVariableName.kt:82)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.WildcardTypeName.equals(WildcardTypeName.kt:63)
        at java.base/java.util.Objects.equals(Objects.java:64)
        at java.base/java.util.ArrayList.equalsRange(ArrayList.java:625)
        at java.base/java.util.ArrayList.equals(ArrayList.java:612)
        at java.base/java.util.Collections$UnmodifiableList.equals(Collections.java:1498)
        at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
        at com.squareup.kotlinpoet.ParameterizedTypeName.equals(ParameterizedTypeName.kt:115)
```

eyupcanakman:
`TypeVariableName` builds cyclic graphs for recursively bound generics like `Enum<E : Enum<E>>`, where the bound points back at the variable. `equals` and `hashCode` followed those bounds with no guard and overflowed the stack. `equals` now walks them through a shared identity guard that stops the moment it revisits a pair. `hashCode` leaves bounds out, which JavaPoet does too.

Fixes #1737.

    be added to the changelog.

### DIFF
--- a/kotlinpoet/api/kotlinpoet.api
+++ b/kotlinpoet/api/kotlinpoet.api
@@ -85,7 +85,6 @@ public final class com/squareup/kotlinpoet/ClassName : com/squareup/kotlinpoet/T
 	public fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/ClassName;
 	public synthetic fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public final fun enclosingClassName ()Lcom/squareup/kotlinpoet/ClassName;
-	public fun equals (Ljava/lang/Object;)Z
 	public final fun getCanonicalName ()Ljava/lang/String;
 	public final fun getPackageName ()Ljava/lang/String;
 	public final fun getSimpleName ()Ljava/lang/String;
@@ -526,7 +525,6 @@ public final class com/squareup/kotlinpoet/LambdaTypeName : com/squareup/kotlinp
 	public synthetic fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public final fun copy (ZLjava/util/List;ZLjava/util/Map;)Lcom/squareup/kotlinpoet/LambdaTypeName;
 	public static synthetic fun copy$default (Lcom/squareup/kotlinpoet/LambdaTypeName;ZLjava/util/List;ZLjava/util/Map;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/LambdaTypeName;
-	public fun equals (Ljava/lang/Object;)Z
 	public static final fun get (Lcom/squareup/kotlinpoet/TypeName;Ljava/util/List;Lcom/squareup/kotlinpoet/TypeName;)Lcom/squareup/kotlinpoet/LambdaTypeName;
 	public static final fun get (Lcom/squareup/kotlinpoet/TypeName;[Lcom/squareup/kotlinpoet/ParameterSpec;Lcom/squareup/kotlinpoet/TypeName;)Lcom/squareup/kotlinpoet/LambdaTypeName;
 	public static final fun get (Lcom/squareup/kotlinpoet/TypeName;[Lcom/squareup/kotlinpoet/TypeName;Lcom/squareup/kotlinpoet/TypeName;)Lcom/squareup/kotlinpoet/LambdaTypeName;
@@ -702,7 +700,6 @@ public final class com/squareup/kotlinpoet/ParameterizedTypeName : com/squareup/
 	public synthetic fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public final fun copy (ZLjava/util/List;Ljava/util/Map;Ljava/util/List;)Lcom/squareup/kotlinpoet/ParameterizedTypeName;
 	public static synthetic fun copy$default (Lcom/squareup/kotlinpoet/ParameterizedTypeName;ZLjava/util/List;Ljava/util/Map;Ljava/util/List;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/ParameterizedTypeName;
-	public fun equals (Ljava/lang/Object;)Z
 	public static final fun get (Lcom/squareup/kotlinpoet/ClassName;Lcom/squareup/kotlinpoet/TypeName;)Lcom/squareup/kotlinpoet/ParameterizedTypeName;
 	public static final fun get (Lcom/squareup/kotlinpoet/ClassName;Ljava/util/List;)Lcom/squareup/kotlinpoet/ParameterizedTypeName;
 	public static final fun get (Lcom/squareup/kotlinpoet/ClassName;[Lcom/squareup/kotlinpoet/TypeName;)Lcom/squareup/kotlinpoet/ParameterizedTypeName;
@@ -913,7 +910,7 @@ public abstract class com/squareup/kotlinpoet/TypeName : com/squareup/kotlinpoet
 	public abstract fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public static synthetic fun copy$default (Lcom/squareup/kotlinpoet/TypeName;ZLjava/util/List;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/TypeName;
 	public static synthetic fun copy$default (Lcom/squareup/kotlinpoet/TypeName;ZLjava/util/List;Ljava/util/Map;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/TypeName;
-	public fun equals (Ljava/lang/Object;)Z
+	public final fun equals (Ljava/lang/Object;)Z
 	public fun getAnnotations ()Ljava/util/List;
 	public fun getTags ()Ljava/util/Map;
 	public fun hashCode ()I
@@ -1171,7 +1168,6 @@ public final class com/squareup/kotlinpoet/TypeVariableName : com/squareup/kotli
 	public synthetic fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeVariableName;
 	public static synthetic fun copy$default (Lcom/squareup/kotlinpoet/TypeVariableName;ZLjava/util/List;Ljava/util/List;ZLjava/util/Map;ILjava/lang/Object;)Lcom/squareup/kotlinpoet/TypeVariableName;
-	public fun equals (Ljava/lang/Object;)Z
 	public static final fun get (Ljava/lang/String;)Lcom/squareup/kotlinpoet/TypeVariableName;
 	public static final fun get (Ljava/lang/String;Lcom/squareup/kotlinpoet/KModifier;)Lcom/squareup/kotlinpoet/TypeVariableName;
 	public static final fun get (Ljava/lang/String;Ljava/util/List;)Lcom/squareup/kotlinpoet/TypeVariableName;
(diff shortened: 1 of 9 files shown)


JakeWharton:
Since we own this type and all subtypes I think I'd prefer to do a local queue-and-loop approach as opposed to something thread-local.

eyupcanakman:
Done, reworked it as a local queue-and-loop over the bound graph instead of the thread-local.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -79,17 +79,65 @@ private constructor(
...
+      val (left, right) = stack.removeLast()
+      if (left === right) continue
+      if (left.javaClass != right.javaClass) return false
+      if (seen.any { it.first === left && it.second === right }) continue
+      seen += left to right
+      if (left.isNullable != right.isNullable) return false
+      if (left.annotations != right.annotations) return false
+
+      when (left) {
+        is ParameterizedTypeName -> {
```
Egorand: I don't love it that we're duplicating `equals` implementations for these `TypeName` subclasses, it's a footgun. Maybe we can try the following approach?

```kotlin
class TypeName {
  ...
  // Mark as final so subtypes can't override it
  final override fun equals(other: Any?): Boolean {
    if (this === other) return true
    if (javaClass != other?.javaClass) return false
    return equals(other as TypeName, seen = mutableSetOf<>())
  }

  // Subtypes should override this method.
  protected fun equals(other: TypeName, seen: Set<Pair<TypeName, TypeName>>): Boolean {
    if (this to other in seen) return true
    return ...also { seen += this to other }
  } 
}
```

Similar approach could work for `hashCode`, that way we won't have diverging implementations for `equals` and `hashCode`.

eyupcanakman: The duplication bugs me too and I'd like to fold it back. The catch with that exact shape is the `seen` set has to match by reference identity rather than value. `this to other in seen` over a `Set<Pair<TypeName, TypeName>>` compares structurally, so the lookup calls `equals`/`hashCode` on the same cyclic `TypeName`s, and for `Enum<E : Enum<E>>` it drops straight back into the overflow this PR fixes. That's why the walk keeps an identity (`===`) seen-list, and why `hashCode` leaves the self-referential bound out of the hash (matching JavaPoet) rather than threading `seen` through it.

I can still pull the duplication out though. A base `equals` can set up the one identity guard and hand off to a `protected equals(other, seen)` that each subtype overrides for its own fields, with a small helper for comparing child `TypeName`s through the guard. Want me to rework it that way?

Egorand: > I can still pull the duplication out though. A base `equals` can set up the one identity guard and hand off to a `protected equals(other, seen)` that each subtype overrides for its own fields, with a small helper for comparing child `TypeName`s through the guard. Want me to rework it that way?

Yep, it'd be great to try and see what it looks like, thanks!

eyupcanakman: Reworked it the way we discussed. `TypeName.equals` is now final and hands off to an `equalsWithGuard(other, seen)` that each subtype overrides for its own fields, so the duplication is gone.

Two notes. The `seen` set keys on reference identity (an `IdentityHashMap`) rather than `Pair` equality, since a structural `this to other in seen` would re-enter `equals`/`hashCode` on the cyclic nodes and overflow again. And `hashCode` keeps dropping the self-referential bound instead of threading `seen` through it, which alone breaks every cycle. JavaPoet leaves type-variable bounds out of its hash too.

kotlinpoet/api/kotlinpoet.api
```diff
@@ -85,7 +85,6 @@ public final class com/squareup/kotlinpoet/ClassName : com/squareup/kotlinpoet/T
 	public fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/ClassName;
 	public synthetic fun copy (ZLjava/util/List;Ljava/util/Map;)Lcom/squareup/kotlinpoet/TypeName;
 	public final fun enclosingClassName ()Lcom/squareup/kotlinpoet/ClassName;
-	public fun equals (Ljava/lang/Object;)Z
```
Egorand, line 88: Technically, IIUC, we'll be breaking any potential Java subclasses of `TypeName` in library user code, there should be no Kotlin subclasses since `TypeName` hierarchy is sealed. I think it's fine, though we can omit `final` to prevent potential API breakage.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeVariableName.kt
```diff
@@ -71,15 +71,17 @@ private constructor(
...
 
-  override fun equals(other: Any?): Boolean {
-    if (this === other) return true
-    if (javaClass != other?.javaClass) return false
-    if (!super.equals(other)) return false
+  override fun equalsWithGuard(other: TypeName, seen: RecursiveComparison): Boolean {
+    // A recursively bounded variable like `Enum<E : Enum<E>>` reaches itself through its bounds,
+    // the only place a type-name cycle can form. Recording the pair here lets the repeat visit
+    // cut the cycle instead of overflowing the stack. (#1737)
+    if (!seen.enter(this, other)) return true
```
Egorand, line 78: Can this check be done inside both versions of `deepEquals()` instead, so that all types are protected by this defense mechanism for free? Currently it's only needed here in `TypeVariableName`, but future use case might require a similar check in other `TypeName` subtypes.

Egorand (approved):
I think it turned out nicely, left a couple nits.

kotlinpoet/api/kotlinpoet.api
eyupcanakman: Since `TypeName` is sealed there's no external subclass that `final` could break, and keeping it stops a subtype from overriding `equals` directly and skipping the guard. I'm fine dropping it if you'd rather though, it's a one-liner.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeVariableName.kt
eyupcanakman: I looked at moving it into `deepEquals`, but a cycle can only start at a `TypeVariableName` since its bounds are the one `TypeName` field that can point back at itself, and the others are built bottom-up. Running `enter()` in `deepEquals` would fire on every nested comparison including `ClassName`, which puts the `IdentityHashMap` allocation back on the hot path, so I left the guard at the cycle origin. If another subtype ever gets a field that can point back the same way, we can add one there too.

kotlinpoet/api/kotlinpoet.api
Egorand: Yeah, I was worried if a _Java_ class could accidentally subclass `TypeName`, but I couldn't find a way to do it in a test project. So let's keep the `final`!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeVariableName.kt
Egorand: Agreed!

Egorand:
Merged, thank you!

PR #2325, opened 2026-06-01
Add support for explicit backing fields

### LINKED ISSUE #2317: Support explicit backing fields
Explicit backing fields ([KEEP-0430](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0430-explicit-backing-fields.md)) became stable in Kotlin 2.4. Here's an example of the syntax:

```kotlin
class SomeViewModel : ViewModel() {
    val city: LiveData<String>
        field = MutableLiveData<String>()
  
    fun updateCity(newValue: String) {
        city.value = newValue
    }
}
```

This can be implemented by adding a `backingField: CodeBlock` property to `PropertySpec`, which can be populated via `PropertySpec.Builder`.

Egorand:
Closes #2317 

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/PropertySpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/PropertySpec.kt
@@ -52,6 +52,8 @@ private constructor(
   public val getter: FunSpec? = builder.getter
   public val setter: FunSpec? = builder.setter
   public val receiverType: TypeName? = builder.receiverType
+  public val backingFieldType: TypeName = builder.backingFieldType
+  public val backingFieldInitializer: CodeBlock? = builder.backingFieldInitializer
 
   init {
     require(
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/PropertySpec.kt
```diff
@@ -134,6 +136,17 @@ private constructor(
...
+      codeWriter.emitCode("⇥")
+      codeWriter.emit("field")
+      if (backingFieldType != type) {
+        codeWriter.emitCode(": %T", backingFieldType)
+      }
+      if (backingFieldInitializer != null) {
+        codeWriter.emitCode(" = %L", backingFieldInitializer)
+      }
+      codeWriter.emitCode("⇤")
+    }
```
Egorand, line 149: There's a bunch of [restrictions](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0430-explicit-backing-fields.md#restrictions-on-a-property-with-an-explicit-backing-field) on which other property features can and cannot be used in combination with explicit backing fields, I decided to not encode any of them into KotlinPoet and to let the compiler enforce them.

JakeWharton (approved):

PR #2331, opened 2026-06-06
Add `CodeBlockHolder`

eyupcanakman:
Holds a `CodeBlock` body. On `FileSpec` the code methods stay script-only, as before.

Part of #1553

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
@@ -281,7 +281,7 @@ private constructor(builder: Builder, private val tagMap: TagMap = builder.build
       get() = memberImports.toList()
 
     public val members: MutableList<Any> = mutableListOf()
-    internal val body = CodeBlock.builder()
+    override val body: CodeBlock.Builder = CodeBlock.builder()
 
     /**
      * Add an annotation to the file.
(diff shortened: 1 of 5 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -463,17 +464,17 @@ private constructor(builder: Builder, private val tagMap: TagMap = builder.build
 
     public fun indent(indent: String): Builder = apply { this.indent = indent }
 
-    public fun addCode(format: String, vararg args: Any?): Builder = apply {
+    override fun addCode(format: String, vararg args: Any?): Builder = apply {
       check(isScript) { "addCode() is only allowed in script files" }
       body.add(format, *args)
     }
```
Egorand: Should `CodeBlockHolder.Builder` methods have default implementations, so that methods here can be refactored like this?


```suggestion
    override fun addCode(format: String, vararg args: Any?): Builder {
check(isScript) { "addCode() is only allowed in script files" }
return super.addCode(format, *args)
}
```

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -507,15 +509,15 @@ private constructor(
       modifiers: Iterable<KModifier>,
     ): Builder = addParameter(name, type.asTypeName(), modifiers)
 
-    public fun addCode(format: String, vararg args: Any?): Builder = apply {
+    override fun addCode(format: String, vararg args: Any?): Builder = apply {
```
Egorand: ...and then the implementations of builder methods here can be deleted?

Egorand (approved):
Looks good!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
JakeWharton: Unfortunately you can't entirely delete these as we need the more-specific return type to maintain binary compatibility.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
eyupcanakman: Switched to default impls on the interface, matching `Documentable`. One thing though, that makes `body` public, so a non-script `FileSpec` can now skip its `isScript` guard by touching `body` directly. Want me to keep these abstract instead?

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
Egorand: Ah, true!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
Egorand: Hmm, yeah, and since we have to provide synthetic overrides for binary compat anyway, reverting to abstract is probably a better move. Something to improve in 3.0 (whenever that happens). Wanna revert the latest commit?

eyupcanakman: Done.

Egorand: Sweet, merging, thank you!

PR #2332, opened 2026-06-08
Fix delegated constructor multiline formatting

### LINKED ISSUE #1762: Chained constructor call is not formatted correctly
**Describe the bug**
When I use .callThisConstructor to have one constructor call another one, it looks formatting was forgotten.

**To Reproduce**

```
      TypeSpec.classBuilder("HelloWorld")
        .addFunction(
          FunSpec.constructorBuilder()
            .addParameter("x".repeat(30), String::class)
            .addParameter("y".repeat(30), String::class)
            .callThisConstructor(
              CodeBlock.of("x".repeat(30)),
              CodeBlock.of("y".repeat(30)),
              CodeBlock.of("x".repeat(30)),
              CodeBlock.of("y".repeat(30)),
              buildCodeBlock {
                beginControlFlow("thisIsAMethodThatUsesALambdaAndReturnsSomethingForTheConstructor")
                addStatement("z".repeat(30) + "()")
                endControlFlow()
              }
            )
            .build()
        )
        .build()
    )
    .build()
    .writeTo(System.out)

```

This prints:

```
package com.test

import kotlin.String

public class HelloWorld {
  public constructor(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String, yyyyyyyyyyyyyyyyyyyyyyyyyyyyyy: String)
      : this(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx, yyyyyyyyyyyyyyyyyyyyyyyyyyyyyy,
      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx, yyyyyyyyyyyyyyyyyyyyyyyyyyyyyy,
      thisIsAMethodThatUsesALambdaAndReturnsSomethingForTheConstructor {
    zzzzzzzzzzzzzzzzzzzzzzzzzzzzzz()
  }
  )
}
```

If this() is counted as the 1st line, then the 2nd and 3rd lines are aligned vertically with the colon but they should be more indented than that, because the this() call opened a new nesting level like it was a statement call. And what follows inside the method call in the 4th and 5th lines should be indented even more but it's actually going backwards.

istvan-florio:
Fixes #1762.

## Summary

- Preserve the existing one-line delegated constructor formatting for simple `this(...)` / `super(...)` calls.
- Expand delegated constructor calls to one argument per line when any delegated constructor argument explicitly emits multiline code.
- Cover `this`, `super`, constructor body formatting, newline string literals, and long single-line arguments with exact output tests.

Note: the multiline predicate intentionally targets explicit emitted newlines in `CodeBlock` formatting. String literal contents emitted via `%S` / `%P` are left on the existing path so raw/string-literal newlines do not force expanded delegated-call formatting.

## Test plan

- `./gradlew :kotlinpoet:jvmTest --tests com.squareup.kotlinpoet.FunSpecTest`
- `./gradlew :kotlinpoet:apiCheck`
- `git diff --check`
- `./gradlew clean build`

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
@@ -175,7 +175,7 @@ private constructor(
 
     if (delegateConstructor != null) {
       codeWriter.emitCode(
-        delegateConstructorArguments.joinToCode(prefix = " : $delegateConstructor(", suffix = ")")
+        delegateConstructorArguments.toDelegateConstructorCode(delegateConstructor)
       )
     }
   }
@@ -693,3 +693,39 @@ private constructor(
     }
   }
 }
+
+private fun List<CodeBlock>.toDelegateConstructorCode(delegateConstructor: String): CodeBlock {
+  if (none(CodeBlock::isMultilineCode)) {
+    return joinToCode(prefix = " : $delegateConstructor(", suffix = ")")
+  }
+
+  return buildCodeBlock {
+    add(" : %L(\n⇥", delegateConstructor)
+    forEachIndexed { index, argument ->
+      if (index > 0) add("\n")
+      add("%L,", argument.trimTrailingNewLine())
+    }
+    add("\n⇤)")
+  }
+}
+
+private fun CodeBlock.isMultilineCode(): Boolean {
+  var argumentIndex = 0
+  for (formatPart in formatParts) {
+    if ('\n' in formatPart) return true
+
+    if (formatPart == "%L") {
+      when (val argument = args[argumentIndex]) {
+        is CodeBlock -> if (argument.isMultilineCode()) return true
+        is String -> if ('\n' in argument) return true
+      }
+    }
+
+    if (formatPart in ARGUMENT_PLACEHOLDERS) {
+      argumentIndex++
+    }
+  }
+  return false
+}
+
+private val ARGUMENT_PLACEHOLDERS = setOf("%L", "%N", "%S", "%P", "%T", "%M")
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
@@ -15,6 +15,7 @@
  */
 package com.squareup.kotlinpoet
 
+import com.squareup.kotlinpoet.CodeBlock.Companion.isPlaceholder
 import com.squareup.kotlinpoet.KModifier.ABSTRACT
 import com.squareup.kotlinpoet.KModifier.EXPECT
 import com.squareup.kotlinpoet.KModifier.EXTERNAL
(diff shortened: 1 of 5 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -693,3 +693,39 @@ private constructor(
...
+    add("\n⇤)")
+  }
+}
+
+private fun CodeBlock.isMultilineCode(): Boolean {
+  var argumentIndex = 0
+  for (formatPart in formatParts) {
+    if ('\n' in formatPart) return true
+
+    if (formatPart == "%L") {
```
JakeWharton, line 717: This wouldn't handle positional format arguments.

JakeWharton: Unless we normalize those on the way in. I honestly can't remember...

istvan-florio:
Thanks, good catch. I pushed a follow-up for the positional/nested `CodeBlock` case.

What changed:
- `trimTrailingNewLine` now recurses into trailing `CodeBlock` placeholder args instead of only trimming `String` args.
- Added a reordered positional regression case using `%2L, %1L` so the test covers the indexed-placeholder path directly.
- Added nested `CodeBlock` coverage for `ensureEndsWithNewLine` as well.

Verification run locally:
- `./gradlew :kotlinpoet:jvmTest :kotlinpoet:apiCheck`
- `git diff --check`

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/FunSpecTest.kt
```diff
@@ -1031,6 +1031,194 @@ class FunSpecTest {
       )
   }
 
+  @Test
+  fun thisConstructorDelegateWithMultilineArgument() {
+    val typeSpec =
+      TypeSpec.classBuilder("Test")
+        .primaryConstructor(FunSpec.constructorBuilder().build())
```
Egorand: nit: I think this primary constructor is not needed in this test, we can remove it. Same for the next test below.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -693,3 +693,39 @@ private constructor(
...
+      when (val argument = args[argumentIndex]) {
+        is CodeBlock -> if (argument.isMultilineCode()) return true
+        is String -> if ('\n' in argument) return true
+      }
+    }
+
+    if (formatPart in ARGUMENT_PLACEHOLDERS) {
```
Egorand: Please use the `isPlaceholder()` method in `CodeBlock.Companion` instead, and remove the `ARGUMENT_PLACEHOLDERS` set declared below.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -693,3 +693,39 @@ private constructor(
...
+  }
+
+  return buildCodeBlock {
+    add(" : %L(\n⇥", delegateConstructor)
+    forEachIndexed { index, argument ->
+      if (index > 0) add("\n")
+      add("%L,", argument.trimTrailingNewLine())
+    }
+    add("\n⇤)")
+  }
```
Egorand: Can be simplified to this:


```suggestion
  return if (none(CodeBlock::isMultilineCode)) {
    joinToCode(prefix = " : $delegateConstructor(", suffix = ")")
  } else {
    joinToCode(prefix = " : $delegateConstructor(\n⇥", separator = "\n", suffix = "\n⇤)") { argument ->
      CodeBlock.of("%L,", argument.trimTrailingNewLine())
    }
  }
```

Egorand (commented):
Thanks! Looks good and the test coverage is pretty solid. Called out a couple small issues that would be great to address before merging.

I wonder if we can and should apply the same treatment to other places where we accept potentially multiline code blocks, such as property initializers, default argument values, etc.

istvan-florio:
Thanks, addressed the review comments in a follow-up push.

What changed:
- Removed the redundant empty primary constructors from the two new tests and updated their expected class headers.
- Replaced the local placeholder set with `CodeBlock.Companion.isPlaceholder`.
- Simplified the multiline delegate constructor formatting to use `joinToCode(...){ ... }`.

Verification:
- `git diff --check`
- `./gradlew --no-daemon --max-workers=1 :kotlinpoet:spotlessKotlinCheck`
- `./gradlew --no-daemon --max-workers=1 -Dkotlin.compiler.execution.strategy=in-process :kotlinpoet:jvmTest --tests com.squareup.kotlinpoet.FunSpecTest`

Egorand:
Thanks, looks good! Can you please sign the CLA and tick the box in the pull request description so we can proceed with the merge?

Egorand:
@istvan-florio a friendly reminder: we still need you to sign the CLA so we can merge your contribution. Thanks!

Egorand:
I'm closing this PR since we can't merge it without a signed CLA and we haven't heard back from you.

PR #2334, opened 2026-06-12
Disallow the deprecated `inline class` syntax

### LINKED ISSUE #2330: Disallow "inline class" syntax
According to [this blog post](https://blog.jetbrains.com/kotlin/2021/02/new-language-features-preview-in-kotlin-1-4-30/#inline-value-classes) dating back to 02/2021 and Kotlin 1.4.30:

> You define a value class with one constructor parameter and annotate it with `@JvmInline`. We expect everyone to use this new syntax starting from Kotlin 1.5. The old syntax `inline class` will continue to work for some time. It will be deprecated with a warning in 1.5 that will include an option to migrate all your declarations automatically. It will later be deprecated with an error.

As of the time of writing, `inline class` only produces a warning, however, it's been over 5 years since this syntax has been officially deprecated, so the library can probably disallow it at this point.

eyupcanakman:
`TypeSpec.classBuilder("Foo").addModifiers(KModifier.INLINE)` still emits `inline class Foo`, five years after the syntax was deprecated in favor of `value` classes. `build()` now rejects the modifier on classes the same way `PropertySpec` already rejects it on properties. The value-class check messages drop the inline wording too.

Fixes #2330.

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
@@ -544,8 +544,8 @@ private constructor(
     internal val isCompanion
       get() = kind == Kind.OBJECT && COMPANION in modifiers
 
-    internal val isInlineOrValClass
-      get() = kind == Kind.CLASS && (INLINE in modifiers || VALUE in modifiers)
+    internal val isValueClass
+      get() = kind == Kind.CLASS && VALUE in modifiers
 
     internal val isSimpleClass
       get() = kind == Kind.CLASS && !isEnum && !isAnnotation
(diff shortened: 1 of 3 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/TypeSpec.kt
```diff
@@ -841,6 +841,13 @@ private constructor(
     // endregion
 
     public fun build(): TypeSpec {
+      if (kind == Kind.CLASS && INLINE in modifiers) {
+        throw IllegalArgumentException(
+          "KotlinPoet doesn't allow setting the inline modifier on " +
+            "classes. You should use the value modifier instead."
```
JakeWharton, line 847: Not sure we should bother checking this. The language feature hasn't existed for a long time, and we don't check for any other completely invalid modifiers.

But if for some reason we do keep this, the message needs worded to not be sentences. Just "inline is not a valid class modifier" is enough.

Egorand: My vote is for dropping this check.

felcrk-ctrl (commented):
,,,,

eyupcanakman:
Makes sense, closing this out.

PR #2338, opened 2026-06-20
Don't escape `get` and `set` operator function names

### LINKED ISSUE #1869: The get / set index access operator overloads are incorrectly escaped
**Describe the bug**
Generating the index-access operator overloads for get / set escapes the function name producing:
```
public operator fun `get`(index: Int): Int = values[index]
```

instead of:
```
public operator fun get(index: Int): Int = values[index]
```

**To Reproduce**
This code to generate a hypothetical collection:
```
val file = FileSpec.builder("", "MyCollectionExample")
        .addType(
            TypeSpec.classBuilder(ClassName("", "MyCollectionExample"))
                .primaryConstructor(
                    FunSpec.constructorBuilder()
                        .addParameter("values", typeNameOf<IntArray>())
                        .build()
                ).addProperty(
                    PropertySpec.builder("values", typeNameOf<IntArray>(), KModifier.PRIVATE)
                        .initializer("values")
                        .build()
                )
                .addFunction(
                    FunSpec.builder("get")
                        .addModifiers(KModifier.OPERATOR)
                        .returns(Int::class)
                        .addParameter("index", Int::class)
                        .addStatement("return values[index]")
                        .build()
                )
                .build()
        )
        .build()

    file.writeTo(System.out)
```

produces:
```
public class MyCollectionExample(
  private val values: IntArray,
) {
  public operator fun `get`(index: Int): Int = values[index]
}
```
notice that the function name is escaped

**Expected behavior**
I would expect the function name not to be escaped when overloading the index access operator.

**Additional context**
This was probably introduced by:
https://github.com/square/kotlinpoet/pull/994

Skimming through the code, it's interesting that MemberName.emit tries to account for this scenario by only escaping the name when the operator is null but I'm not familiar with the code.  I wonder if we can just remove the get / set soft-keywords from Util.KEYWORDS but I'm not sure if that would introduce any side effects.

eyupcanakman:
`get` and `set` are soft keywords, so KotlinPoet wrapped them in backticks when generating a function name. That also hit the index access operator overloads, so an `operator fun get` came out with a backtick-escaped name instead of the bare `get`. `KOperator` has no entry for these operators, so the operator check in `MemberName` never applied. KotlinPoet now keeps the bare `get`/`set` name for an operator function.

Fixes #1869.

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
@@ -19,6 +19,7 @@ import com.squareup.kotlinpoet.KModifier.ABSTRACT
 import com.squareup.kotlinpoet.KModifier.EXPECT
 import com.squareup.kotlinpoet.KModifier.EXTERNAL
 import com.squareup.kotlinpoet.KModifier.INLINE
+import com.squareup.kotlinpoet.KModifier.OPERATOR
 import com.squareup.kotlinpoet.KModifier.VARARG
 import java.lang.reflect.Type
 import javax.lang.model.element.Element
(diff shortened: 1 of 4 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FunSpec.kt
```diff
@@ -163,7 +164,13 @@ private constructor(
           codeWriter.emitCode("%T.", receiverType)
         }
       }
-      codeWriter.emitCode("%N", this)
+      if (OPERATOR in modifiers && (name == "get" || name == "set")) {
```
Egorand, line 167: I guess we probably don't want to add get and set to `KOperator` since we won't really be able to use them in `MemberName`?

Egorand (approved):
Thanks!

PR #2340, opened 2026-06-22
Add `CodeBlock.Builder.addComment()`

### LINKED ISSUE #1690: CodeBlock.builder() is missing addComment() method
Why does CodeBlock.builder() not have addComment() method like e.g. FunSpec.builder() ?

eyupcanakman:
Adds `addComment()` to `CodeBlock.Builder`, mirroring `FunSpec.Builder.addComment()` so a comment can be added when only a `CodeBlock.Builder` is in scope.

Closes #1690

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,7 @@ Change Log
  * New: Support for explicit backing fields. (#2325)
  * New: `value class` validations have been relaxed to support multi-field value classes. (#2329)
  * New: Extract `CodeBlockHolder` interface for constructs that can hold a `CodeBlock` body and their builders. (#1553)
+ * New: Add `CodeBlock.Builder.addComment()` for adding `//` comments. (#1690)
  * Fix: Keep the `//` prefix on wrapped file comment lines. (#1922)
  * Fix: `TypeVariableName.equals`/`hashCode` no longer overflow on recursively bound generics like `Enum<E : Enum<E>>`. (#1737)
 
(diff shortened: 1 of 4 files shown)


JakeWharton:
This seems like it should go on `CodeBlockHolder.Builder` so it automatically shows up anywhere you can "write" code.

Egorand:
> This seems like it should go on `CodeBlockHolder.Builder` so it automatically shows up anywhere you can "write" code.

`CodeBlock` does not implement `CodeBlockHolder` though, but I'd say let's add it to both.

Egorand (approved):
Thanks! I think this can be merged as is, with a follow up that adds a similar method to `CodeBlockHolder`.

JakeWharton:
Should it be called CodeHolder and CodeBlock (and it's Builder) also be a subtype then?

Egorand:
> Should it be called CodeHolder and CodeBlock (and it's Builder) also be a subtype then?

Worth exploring. Can't think of a use case where you'd need a shared interface between a `FunSpec` and a `CodeBlock`, but the API is already pretty much identical so why not.

JakeWharton:
In the past, for me, it was creating helpers that added reusable code patterns and then wanting to apply them in different contexts. Sometimes a function, sometimes a property initializer, etc.

cbishopjha (approved):

eyupcanakman:
I can take the follow-up. One thing to decide on `FileSpec` first. It already has `addComment()` as the deprecated file-comment alias, so a body one can't just go on the interface without clashing there. Exclude `FileSpec`, or give it a body `addComment()` too?

Egorand:
> I can take the follow-up. One thing to decide on `FileSpec` first. It already has `addComment()` as the deprecated file-comment alias, so a body one can't just go on the interface without clashing there. Exclude `FileSpec`, or give it a body `addComment()` too?

I'd say we can have `FileSpec` implement `CodeHolder` and mark `addComment()` with `override`, even though it's deprecated. My expectation is that if someone calls that method on a reference with compile time type of `CodeHolder` and runtime type of `FileSpec` they'll check the docs to find out what exactly `FileSpec.addComment()` does, and if both compile and runtime types are `FileSpec` then the deprecation will still have an effect.

PR #2347, opened 2026-07-03
Add `CodeBlockHolder.Builder.addComment()`

eyupcanakman:
Adds `addComment()` to `CodeBlockHolder.Builder` so it's available on any code-block builder. `FileSpec.Builder` marks its deprecated `addComment()` as `override`, so it resolves through the interface but still errors on direct calls.

Follow-up to #2340.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlockHolder.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlockHolder.kt
@@ -28,6 +28,8 @@ public interface CodeBlockHolder {
 
     public fun addStatement(format: String, vararg args: Any?): T
 
+    public fun addComment(format: String, vararg args: Any?): T
+
     public fun beginControlFlow(controlFlow: String, vararg args: Any?): T
 
     public fun nextControlFlow(controlFlow: String, vararg args: Any?): T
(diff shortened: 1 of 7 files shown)


kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/FileSpec.kt
```diff
@@ -315,7 +315,9 @@ private constructor(builder: Builder, private val tagMap: TagMap = builder.build
       ReplaceWith("addFileComment(format, args)"),
       DeprecationLevel.ERROR,
     )
-    public fun addComment(format: String, vararg args: Any): Builder = addFileComment(format, *args)
+    override fun addComment(format: String, vararg args: Any?): Builder = apply {
+      comment.add(format, *args)
+    }
```
Egorand: nit: I'd revert this change, IMO it's more clear that `addFileComment()` is a direct replacement for `addComment()` if the latter directly calls the former vs both having the exact same implementation. Plus, less of a chance that the implementations will diverge in the future.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlockHolder.kt
Egorand, line 1: Not in scope of this PR, but it'd be awesome to document this API!

Egorand (approved):

eyupcanakman:
Done, `addComment()` delegates to `addFileComment()` now. It needed `addFileComment()`'s args widened to `Any?` (the interface forces `addComment()` to `Any?`), which is API-invisible since both erase to `Object[]`, and it lines up with `CodeBlock.add()`. If you'd rather leave `addFileComment()`'s signature alone, I'll keep the inline body instead.

Egorand:
> Done, `addComment()` delegates to `addFileComment()` now. It needed `addFileComment()`'s args widened to `Any?` (the interface forces `addComment()` to `Any?`), which is API-invisible since both erase to `Object[]`, and it lines up with `CodeBlock.add()`. If you'd rather leave `addFileComment()`'s signature alone, I'll keep the inline body instead.

Thanks! Yep, I'm fine with the signature change.

PR #2365, opened 2026-08-17
Don't leak an expression body's indent into later declarations

### LINKED ISSUE #1421: Expression function runaway indentation
**Describe the bug**

When using `withIndent` in an expression function, the emitted code does not unindent properly. Automatic wrapping then causes compilation to fail as deeply nested code gets squeezed against the margin.

**To Reproduce**

Unfortunately I don't have a simple example, but I do have a branch that reproduces the problem: https://github.com/andrewparmet/protokt/tree/demo-broken-indentation

To reproduce, run `./gradlew protokt-core-lite:build`. The issue manifests in the generated code file `descriptor.kt`.

Errors are like: `e: /Users/.../protokt/protokt-core-lite/build/generated-sources/main/protokt/com/toasttab/protokt/descriptor.kt: (8132, 176): Expecting a top level declaration`

**Expected behavior**

My overall goal is to generate expression functions whose bodies begin on the next line, like so:

```kotlin
fun foo() =
    "some very long expression " +
        "that stretches on for many lines"
```

And the source of this bug is in `MessageGenerator.kt`:
```kotlin
buildCodeBlock {
    add(
        "return \nother is %T &&\n".bindSpaces(),
        msg.className
    )
    withIndent {
        equalsLines(properties).forEach(::add)
        add("other.unknownFields == unknownFields".bindSpaces())
    }
}
```

If I remove `withIndent()` then the code generates ok, although it's not pretty.

Thanks!

**Additional Context**

Note that the code generator has a formatting pass that is disabled; it is found in `GeneratedCodeCleanup.kt`.

eyupcanakman:
An expression body that opens or closes an indent leaks it into every declaration after it, so each following function lands two spaces deeper than the last. Deeper nesting throws `cannot unindent` instead.

`CodeBlock.trim()` strips no-arg placeholders from both ends, dropping one half of a pair while its partner stays behind. It now takes them back from either end until nothing is left unmatched.

A block that was already unmatched before trimming is left alone, since taking one end back cannot balance it. Markers that open and close out of order are left alone as well, and building that shape takes raw placeholders rather than the builder.

A body that opens an indent before the returned expression emits as a block body now, since the trimmed block no longer starts with the return. Same code, different shape. Most of the diff is tests.

Fixes #1421.

    be added to the changelog.

### DIFF
--- a/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
+++ b/kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
@@ -119,8 +119,34 @@ private constructor(internal val formatParts: List<String>, internal val args: L
   /**
    * Returns a copy of the code block without leading and trailing no-arg placeholders (`⇥`, `⇤`,
    * `«`, `»`).
+   *
+   * With [keepBalanced], placeholders are taken back from either end until nothing left in the
+   * block is unmatched, which would otherwise leak the indent or the statement into whatever is
+   * emitted next. A body ending inside `withIndent` loses the `⇤` that closed it, e.g.:
+   * ```
+   * ["return ", "%S", "⇥", ".trim()", "⇤"]
+   *   trim()                    -> ["return ", "%S", "⇥", ".trim()"]
+   *   trim(keepBalanced = true) -> ["return ", "%S", "⇥", ".trim()", "⇤"]
+   * ```
+   *
+   * Both ends move when both are unmatched, e.g.:
+   * ```
+   * ["⇥", "body1", "⇤", "⇥", "body2", "⇤"]
+   *   trim()                    -> ["body1", "⇤", "⇥", "body2"]
+   *   trim(keepBalanced = true) -> ["⇥", "body1", "⇤", "⇥", "body2", "⇤"]
+   * ```
+   *
+   * A block that was already unbalanced before trimming is left to the plain behavior, since taking
+   * one end back cannot balance it and moves the following declaration further out, e.g.:
+   * ```
+   * ["⇥", "return ", "%S", "⇤", "⇥", ".trim()"]
+   *   trim()                    -> ["return ", "%S", "⇤", "⇥", ".trim()"]
+   *   trim(keepBalanced = true) -> ["return ", "%S", "⇤", "⇥", ".trim()"]
+   * ```
+   *
+   * [trimTrailingNewLine] relies on the default, which always drops both runs.
    */
-  internal fun trim(): CodeBlock {
+  internal fun trim(keepBalanced: Boolean = false): CodeBlock {
     var start = 0
     var end = formatParts.size
     while (start < end && formatParts[start] in NO_ARG_PLACEHOLDERS) {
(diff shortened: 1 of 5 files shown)


JakeWharton:
I'm wondering if we should just turn off expression bodies. Seems like we only gain pain by having them enabled.

eyupcanakman:
Turning them off does fix it, but 101 of the 1172 tests here fail and the two #1300 examples don't move at all since those are property initializers rather than function bodies. Reworked this one to leave a block alone when its indents already don't match.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -119,8 +119,16 @@ private constructor(internal val formatParts: List<String>, internal val args: L
...
    * Returns a copy of the code block without leading and trailing no-arg placeholders (`⇥`, `⇤`,
    * `«`, `»`).
+   *
+   * With [keepBalanced], placeholders are taken back from either end until nothing left in the
+   * block is unmatched, which would otherwise leak the indent or the statement into whatever is
+   * emitted next. The leading end stops once nothing dips below where the block starts, and the
+   * trailing end stops once nothing is still open. A block that was already unbalanced before
+   * trimming is left to the plain behavior, since taking one end back cannot balance it and moves
+   * the following declaration further out. [trimTrailingNewLine] relies on the default, which
+   * always drops both runs.
```
Egorand: I think this doc could benefit from a couple examples, as it's a bit hard to visualize the described behaviour by just reading the doc.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -129,6 +137,65 @@ private constructor(internal val formatParts: List<String>, internal val args: L
     while (start < end && formatParts[end - 1] in NO_ARG_PLACEHOLDERS) {
       end--
     }
+    if (keepBalanced && (start > 0 || end < formatParts.size) && isBalanced()) {
```
Egorand: This algorithm feels a bit verbose to me. I wonder if the following could work as well and be a bit more straightforward to implement:

1. Make a pass over `formatParts` and find indices of all relevant placeholders, store them separately per placeholder (e.g. in a `Map<placeholder: Char, indices: IntArray>`.
2. Execute the existing trimming logic to find `start` and `end`.
3. Using the indices of placeholders, find out if the range between `start` and `end` is balanced.
4. If not, calculate how many indices and in which direction we need to backtrack to come up with a balanced substring.

eyupcanakman: Switched to that, a single pass now collects the placeholders into a `CharArray` and the rest is index arithmetic over it, which cuts `trim` on a 50k-statement block roughly in half.

The whole-block balance check stays though, since backtracking cannot rescue an already unmatched block.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
eyupcanakman: Added three examples, including one where both ends move and one where the block gets left alone.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -129,6 +155,79 @@ private constructor(internal val formatParts: List<String>, internal val args: L
...
+        }
+        var last = keptEnd
+        while (last < placeholderCount && (indentTotal > 0 || statementTotal > 0)) {
+          when (placeholders[last]) {
+            '⇥' -> indentTotal++
+            '⇤' -> indentTotal--
+            '«' -> statementTotal++
+            '»' -> statementTotal--
+          }
+          if (indentTotal < 0 || statementTotal < 0) break
```
Egorand, line 238: Is there a test case where this statement will evaluate to true? All tests still pass if I comment it out.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
Egorand: Examples look great, thanks!

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -129,6 +155,79 @@ private constructor(internal val formatParts: List<String>, internal val args: L
     while (start < end && formatParts[end - 1] in NO_ARG_PLACEHOLDERS) {
       end--
     }
+    if (keepBalanced && (start > 0 || end < formatParts.size)) {
```
Egorand: It took me a while to fully grasp what's going on here, and something that I think could help is making the naming a bit less ambiguous, I've got a few suggestions:

1. First, let's rename `start` and `end` above to `keepFrom` and `keepUntil`. Simpler naming was fine while this method was trivial, now it heavily clashes with `first`, `last` and `keptEnd` making the whole thing hard to grok.
2. Similarly, let's rename `keptEnd` to `keepPlaceholdersUntil`. 
3. Let's also introduce `keepPlaceholdersFrom` instead of using `keepFrom`. While it's convenient that they have the same value, IMO it's better to have a separate pointer for a separate array.
4. We can probably squash `first` and `last` into a single variable called `pointer` or just `i`, but don't feel strongly here.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
```diff
@@ -148,6 +247,21 @@ private constructor(internal val formatParts: List<String>, internal val args: L
 
   internal fun hasStatements() = formatParts.any { "«" in it }
 
+  private fun isBalanced(placeholders: CharArray, count: Int): Boolean {
```
Egorand: Let's move this helper inside `trim()`?

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/CodeBlockTest.kt
```diff
@@ -494,6 +494,13 @@ class CodeBlockTest {
     assertThat(CodeBlock.of("«»⇥⇤").trim()).isEqualTo(CodeBlock.of(""))
   }
 
+  @Test
+  fun trimKeepsAStatementAndItsIndentTogether() {
```
Egorand: nit: maybe `trimKeepsStatementsAndIndentsBalanced`?

Egorand (commented):
Looks good overall, got some feedback on naming.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
eyupcanakman: It only fires when a statement and an indent cross, e.g. `⇥return %S«.trim()⇤⇥»⇤`. Without it the trailing walk pulls in `⇤⇥»` and emitting the block throws `cannot unindent 1 from 0`, so I added `trimLeavesCrossedStatementAndIndentAlone` to pin that shape.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
eyupcanakman: Renamed all four and went with `pointer` for the merged index. The placeholder array gets its own `keepPlaceholdersFrom` now.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
eyupcanakman: Done, it's a local fun now.

kotlinpoet/src/jvmTest/kotlin/com/squareup/kotlinpoet/CodeBlockTest.kt
eyupcanakman: Took that name.

kotlinpoet/src/jvmMain/kotlin/com/squareup/kotlinpoet/CodeBlock.kt
Egorand: Awesome, thanks for adding a dedicated test for it!

PR #2380, opened 2026-09-07
Keep carriage returns and raw-string newlines in emitted string literals

lenamonj:
A carriage return could not survive `stringLiteralWithQuotes`. The margin form writes the value into a `"""` literal, the lexer reads a bare CR as a line terminator, and `trimMargin()` rejoins the lines with `\n`, so a CRLF value came out LF-only. The raw-string form used for `%P` lost it the same way, and in a constant context wrote a bare newline that picked up indentation.

Values carrying a CR now take the escaped form, and a raw string splices both line terminators in as `${'\r'}` and `${'\n'}`, as the file already does for `"` and `$`. Values with newlines only are unchanged.

Verified with the added `UtilTest.stringLiteral` cases, red on main and green with the change, and `./gradlew build`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,8 @@ Change Log
 
 ## Unreleased
 
+ * Fix: Carriage returns are no longer lost in `%S` and `%P` string literals. (#2380)
+
 ## Version 2.4.0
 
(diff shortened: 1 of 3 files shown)


Egorand (approved):
Thanks!

PR #2382, opened 2026-09-07
Keep the package slot for default-package classes in Class.asClassName()

lenamonj:
`Class<*>.asClassName()` appended the package element only when the binary name contained a dot, so a class in the default package produced a `ClassName` with no package slot. Every accessor reads that list positionally: `packageName` returned the class's own name, `simpleNames` was empty, and `toString()`, `reflectionName()`, `topLevelClassName()` and `peerClass()` threw `IndexOutOfBoundsException`.

The package element is now appended unconditionally, empty for the default package, which is the form `ClassName("", "Foo")` and `bestGuess("Foo")` already produce.

Verified with `classNameFromClassInDefaultPackage`, red on main and green with the change, and `./gradlew build`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,8 @@ Change Log
 
 ## Unreleased
 
+ * Fix: `Class.asClassName()` no longer drops the package slot of a class in the default package. (#2382)
+
 ## Version 2.4.0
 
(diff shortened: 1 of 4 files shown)


Egorand (approved):
Nice, thanks!

